PR #1, opened 2024-02-15
feat: add zombienet config

evilrobot-01:
Adds zombienet config and simple readme using pop cli.

Tasks:
- [ ] add `chain` once runtimes defined - e.g. `devnet`

NOTE: should only be merged once defaul branch updated with initial version.

### DIFF
--- a/networks/rococo.toml
+++ b/networks/rococo.toml
@@ -1,0 +1,23 @@
+[settings]
+timeout = 1000
+node_spawn_timeout = 300
+
+[relaychain]
+chain = "rococo-local"
+default_command = "./polkadot"
+default_args = ["--insecure-validator-i-know-what-i-do"]
+
+    [[relaychain.nodes]]
+    name = "alice"
+    validator = true
+
+    [[relaychain.nodes]]
+    name = "bob"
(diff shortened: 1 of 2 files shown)


peterwht:
Looks good. My only recommendations are
- add a "name" field under [parachain] as we will have a few runtimes. 
- [subjective] Change the pop collator name to something like: "pop-collator01" or "pop01"

networks/rococo.toml
```diff
@@ -0,0 +1,23 @@
+[settings]
+timeout = 1000
+node_spawn_timeout = 300
+
+[relaychain]
+chain = "rococo-local"
+default_command = "./polkadot"
```
peterwht, line 7:
```suggestion
default_command = "./bin/polkadot"
```

networks/rococo.toml
```diff
@@ -0,0 +1,23 @@
...
+default_args = ["--insecure-validator-i-know-what-i-do"]
```
peterwht, line 8: This flag is already added by zombienet and will result in an error when added twice. 
```suggestion
```

PR #6, opened 2024-02-24
chore: update cargo manifests, rename from template to pop and reset version

peterwht:


evilrobot-01 (approved):
Pushed a few additional commits, feel free to object otherwise LGTM!

Issue #8, opened 2024-02-26
Add'em all!


al3mart:
Add the following pallets to runtime:
- [x] `pallet-nfts`
- [x] `pallet-nfts-runtime-api`
- [x] `pallet-nft-fractionalization`
- [x] `pallet-assets`
- [x] `pallet-scheduler` --> Requires `pallet-pregimage`

All of them added at #9

PR #9, opened 2024-02-26
Add asset pallet suite + scheduler

al3mart:
This merge request adds the following pallets:

- `pallet-nfts`
- `pallet-nfts-runtime-api`
- `pallet-nft-fractionalization`
- `pallet-assets`
- `pallet-scheduler` --> Requires `pallet-preimage`

Closes: #8

### DIFF
--- a/runtime/src/assets_config.rs
+++ b/runtime/src/assets_config.rs
@@ -1,0 +1,127 @@
+use crate::{
+    AccountId, RuntimeEvent, Balances, Balance, UNIT, deposit, DAYS, BlockNumber,
+    RuntimeHoldReason, EXISTENTIAL_DEPOSIT, Runtime, Assets, Nfts,
+};
+use frame_system::{
+    EnsureRoot, EnsureSigned,
+};
+use frame_support::{
+    parameter_types,
+    traits::{
+        AsEnsureOriginWithArg, ConstU32,
+    },
+    BoundedVec, PalletId,
+};
+use pallet_nfts::PalletFeatures;
+use parachains_common::{
+    CollectionId, ItemId, Signature, AssetIdForTrustBackedAssets,
+    
+};
+use sp_runtime::{
+  traits::{
+        Verify,
+    },  
+};
+
+/// We allow root to execute privileged asset operations.
+pub type AssetsForceOrigin = EnsureRoot<AccountId>;
+
+parameter_types! {
+	pub const AssetDeposit: Balance = UNIT / 10; // 1 / 10 UNITS deposit to create asset
+	pub const AssetAccountDeposit: Balance = deposit(1, 16);
+	pub const ApprovalDeposit: Balance = EXISTENTIAL_DEPOSIT;
+	pub const AssetsStringLimit: u32 = 50;
+	/// Key = 32 bytes, Value = 36 bytes (32+1+1+1+1)
+	// https://github.com/paritytech/substrate/blob/069917b/frame/assets/src/lib.rs#L257L271
+	pub const MetadataDepositBase: Balance = deposit(1, 68);
+	pub const MetadataDepositPerByte: Balance = deposit(0, 1);
+}
+
+parameter_types! {
+    pub NftsPalletFeatures: PalletFeatures = PalletFeatures::all_enabled();
+    pub const NftsCollectionDeposit: Balance = UNIT / 10;
+    pub const NftsItemDeposit: Balance = UNIT / 1_000;
+    pub const NftsMetadataDepositBase: Balance = deposit(1, 129);
+    pub const NftsAttributeDepositBase: Balance = deposit(1, 0);
+    pub const NftsDepositPerByte: Balance = deposit(0, 1);
+    pub const NftsMaxDeadlineDuration: BlockNumber = 12 * 30 * DAYS;
+}
+
+impl pallet_nfts::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type CollectionId = CollectionId;
+	type ItemId = ItemId;
+	type Currency = Balances;
+	type CreateOrigin = AsEnsureOriginWithArg<EnsureSigned<AccountId>>;
+	type ForceOrigin = AssetsForceOrigin;
+	type Locker = ();
+	type CollectionDeposit = NftsCollectionDeposit;
+	type ItemDeposit = NftsItemDeposit;
+	type MetadataDepositBase = NftsMetadataDepositBase;
+	type AttributeDepositBase = NftsAttributeDepositBase;
+	type DepositPerByte = NftsDepositPerByte;
+	type StringLimit = ConstU32<256>;
+	type KeyLimit = ConstU32<64>;
+	type ValueLimit = ConstU32<256>;
+	type ApprovalsLimit = ConstU32<20>;
+	type ItemAttributesApprovalsLimit = ConstU32<30>;
+	type MaxTips = ConstU32<10>;
+	type MaxDeadlineDuration = NftsMaxDeadlineDuration;
+	type MaxAttributesPerCall = ConstU32<10>;
+	type Features = NftsPalletFeatures;
+	type OffchainSignature = Signature;
+	type OffchainPublic = <Signature as Verify>::Signer;
+	type WeightInfo = pallet_nfts::weights::SubstrateWeight<Self>;
+	#[cfg(feature = "runtime-benchmarks")]
+	type Helper = ();
+}
+
+parameter_types! {
+	pub const NftFractionalizationPalletId: PalletId = PalletId(*b"fraction");
+	pub NewAssetSymbol: BoundedVec<u8, AssetsStringLimit> = (*b"FRAC").to_vec().try_into().unwrap();
+	pub NewAssetName: BoundedVec<u8, AssetsStringLimit> = (*b"Frac").to_vec().try_into().unwrap();
+}
+
+impl pallet_nft_fractionalization::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type Deposit = AssetDeposit;
+	type Currency = Balances;
+	type NewAssetSymbol = NewAssetSymbol;
+	type NewAssetName = NewAssetName;
+	type StringLimit = AssetsStringLimit;
+	type NftCollectionId = <Self as pallet_nfts::Config>::CollectionId;
+	type NftId = <Self as pallet_nfts::Config>::ItemId;
+	type AssetBalance = <Self as pallet_balances::Config>::Balance;
+	type AssetId = <Self as pallet_assets::Config<TrustBackedAssetsInstance>>::AssetId;
+	type Assets = Assets;
+	type Nfts = Nfts;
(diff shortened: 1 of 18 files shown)


runtime/src/lib.rs
```diff
@@ -211,6 +216,12 @@ pub const UNIT: Balance = 1_000_000_000_000;
 pub const MILLIUNIT: Balance = 1_000_000_000;
 pub const MICROUNIT: Balance = 1_000_000;
 
+// Deposits
+
+pub const fn deposit(items: u32, bytes: u32) -> Balance {
```
evilrobot-01, line 221: I believe this is defined in the contracts module as well - can we remove the one there and used this one instead?

Also, not sure it needs to be `pub`, `pub(crate) `probably sufficient.

runtime/src/lib.rs
```diff
@@ -523,32 +578,50 @@ construct_runtime!(
...
         XcmpQueue: cumulus_pallet_xcmp_queue = 30,
         PolkadotXcm: pallet_xcm = 31,
         CumulusXcm: cumulus_pallet_xcm = 32,
         MessageQueue: pallet_message_queue = 33,
 
+        // Contracts
         Contracts: pallet_contracts = 40,
+
+        // Nfts
+        Nfts: pallet_nfts = 50,
```
evilrobot-01, line 597: Any reason for not just continuing with 41? Doesn't matter, just interested 🙂

runtime/src/assets_config.rs
```diff
@@ -0,0 +1,127 @@
...
+	type Assets = Assets;
+	type Nfts = Nfts;
+	type PalletId = NftFractionalizationPalletId;
+	type WeightInfo = pallet_nft_fractionalization::weights::SubstrateWeight<Self>;
+	type RuntimeHoldReason = RuntimeHoldReason;
+	#[cfg(feature = "runtime-benchmarks")]
+	type BenchmarkHelper = ();
+}
+
+pub type TrustBackedAssetsInstance = pallet_assets::Instance1;
```
evilrobot-01, line 105: Suggestion - drop the `Instance` in the name so it just reads `TrustBackedAssets` with the instance being implied?

runtime/src/lib.rs
```diff
@@ -523,32 +578,50 @@ construct_runtime!(
...
+        NftFractionalization: pallet_nft_fractionalization = 51,
+        Assets: pallet_assets::<Instance1> = 52,
```
evilrobot-01, line 599: Perhaps `TrustBackedAssets` instead of `Assets`? Just thinking that when an additional instance gets added for `ForeignAssets` then the two are differentiated more?

Can probably also use the `TrustBackedAssets` type alias rather than `Instance1` so it reads more clearly.

I guess AH wasn't able to easily rename the existing `Assets` name whereas we have the opportunity to do it right.

runtime/src/lib.rs
evilrobot-01: Aaah I see AH uses the 50+ range for assets...

runtime/src/lib.rs
al3mart: We should ideally use the deposit function part of polkadot constants. But I am happy to amend the code base until have the proper thing in place.

runtime/src/lib.rs
al3mart: Not a special opinion from my side. I was mainly leaving some slack just in case.

runtime/src/lib.rs
```diff
@@ -500,6 +511,50 @@ impl pallet_collator_selection::Config for Runtime {
     type WeightInfo = ();
 }
 
+parameter_types! {
+	pub MaximumSchedulerWeight: Weight = Perbill::from_percent(60) *
```
Daanvdplas, line 515: I see[ moonbeam ](https://github.com/moonbeam-foundation/moonbeam/blob/master/runtime/moonbeam/src/lib.rs#L202)has 75%, curious where you inspired this from

runtime/src/lib.rs
```diff
@@ -523,32 +578,50 @@ construct_runtime!(
...
+        NftFractionalization: pallet_nft_fractionalization = 51,
+        Assets: pallet_assets::<Instance1> = 52,
     }
 );
 
 #[cfg(feature = "runtime-benchmarks")]
 mod benches {
     frame_benchmarking::define_benchmarks!(
         [frame_system, SystemBench::<Runtime>]
+        [pallet_assets, Local]
```
Daanvdplas, line 607: Curious why this is then also not `TrustBacked`. I see this is also how its done for asset hub so no need to change. 

I'm just always curious why they called local assets `TrustBackedAssets` in stead of `LocalAssets`. Foreign assets on the local chain are in practise also local assets so perhaps that is why they overcomplicated the name. Just had to share this :)

Daanvdplas (approved):
Nice work Ale

runtime/src/lib.rs
evilrobot-01: EVM is heavy? :p

al3mart: I set a very conservative restriction as a safe start.

runtime/src/lib.rs
al3mart: Unfortunately we don't get to choose the type alias name: https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/support/src/instances.rs

al3mart:
I've addressed the comments and remove all included benchmarks in this same PR from pallets we don't maintain.

runtime/src/assets_config.rs
```diff
@@ -0,0 +1,116 @@
...
+    pub const AssetsStringLimit: u32 = 50;
+    /// Key = 32 bytes, Value = 36 bytes (32+1+1+1+1)
+    // https://github.com/paritytech/substrate/blob/069917b/frame/assets/src/lib.rs#L257L271
+    pub const MetadataDepositBase: Balance = deposit(1, 68);
+    pub const MetadataDepositPerByte: Balance = deposit(0, 1);
+}
+
+parameter_types! {
+    pub NftsPalletFeatures: PalletFeatures = PalletFeatures::all_enabled();
+    pub const NftsCollectionDeposit: Balance = UNIT / 10;
```
evilrobot-01, line 31: As above: https://github.com/polkadot-fellows/runtimes/blob/59172d5fb8feea2646490618bead3b2b4fb06220/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L719-L720

runtime/src/assets_config.rs
```diff
@@ -0,0 +1,116 @@
...
+impl pallet_nft_fractionalization::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type Deposit = AssetDeposit;
+    type Currency = Balances;
+    type NewAssetSymbol = NewAssetSymbol;
+    type NewAssetName = NewAssetName;
+    type StringLimit = AssetsStringLimit;
+    type NftCollectionId = <Self as pallet_nfts::Config>::CollectionId;
+    type NftId = <Self as pallet_nfts::Config>::ItemId;
+    type AssetBalance = <Self as pallet_balances::Config>::Balance;
```
evilrobot-01, line 83: Net result is the same, but this should be amended to `<Self as pallet_assets::Config<TrustBackedAssets>>::Balance`. This has long bugged me. 🤣

runtime/src/lib.rs
Daanvdplas: This can be `pub (crate)`

runtime/src/assets_config.rs
```diff
@@ -0,0 +1,116 @@
...
+    type MetadataDepositBase = MetadataDepositBase;
+    type MetadataDepositPerByte = MetadataDepositPerByte;
+    type ApprovalDeposit = ApprovalDeposit;
+    type StringLimit = AssetsStringLimit;
+    type Freezer = ();
+    type Extra = ();
+    type WeightInfo = pallet_assets::weights::SubstrateWeight<Self>;
+    type CallbackHandle = ();
+    type AssetAccountDeposit = AssetAccountDeposit;
+    type RemoveItemsLimit = frame_support::traits::ConstU32<1000>;
```
evilrobot-01, line 113: Unneccessary path prefix `frame_support::traits::`

runtime/src/lib.rs
```diff
@@ -500,6 +511,46 @@ impl pallet_collator_selection::Config for Runtime {
...
+    #[cfg(not(feature = "runtime-benchmarks"))]
+    type MaxScheduledPerBlock = ConstU32<50>;
+    type WeightInfo = pallet_scheduler::weights::SubstrateWeight<Runtime>;
+    type OriginPrivilegeCmp = EqualPrivilegeOnly;
+    type Preimages = Preimage;
+}
+
+parameter_types! {
+    pub const PreimageHoldReason: RuntimeHoldReason = RuntimeHoldReason::Preimage(pallet_preimage::HoldReason::Preimage);
+    pub const BaseDeposit: Balance = 1 * UNIT;
```
evilrobot-01, line 537: Should we use the `deposit()` function for these, as per https://github.com/polkadot-fellows/runtimes/blob/59172d5fb8feea2646490618bead3b2b4fb06220/relay/polkadot/src/lib.rs#L241-L242 ?

runtime/src/lib.rs
evilrobot-01: I mean the pallet name within the runtime:

TrustBackedAssets: pallet_assets::<Instance1> = 52

So when we have another instance the names will be `TrustBackedAssets` and `ForeignAssets`. See the comment at https://github.com/polkadot-fellows/runtimes/blob/59172d5fb8feea2646490618bead3b2b4fb06220/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L280, whereby we dont have to worry about a breaking change in the name as its a new chain.

runtime/src/lib.rs
```diff
@@ -523,13 +574,25 @@ construct_runtime!(
...
         // XCM helpers.
         XcmpQueue: cumulus_pallet_xcmp_queue = 30,
         PolkadotXcm: pallet_xcm = 31,
         CumulusXcm: cumulus_pallet_xcm = 32,
         MessageQueue: pallet_message_queue = 33,
 
+        // Contracts
         Contracts: pallet_contracts = 40,
+
+        // Nfts
```
evilrobot-01, line 592: Would prefer `// Assets` in line with the assets config module

evilrobot-01 (commented):
Awesome!! 

Mostly subtle tweaks and comments from me, but a few minor things which may warrant addressing.

runtime/src/assets_config.rs
al3mart: ah, good catch

runtime/src/lib.rs
al3mart: Ah, changed this one to `// Assets` and reverted this morning, forgot to include it back :ok_hand:

evilrobot-01 (approved):
I'd still change `Assets` in `construct_runtime!` to `TrustBackedAssets` but otherwise LGTM.

evilrobot-01:
@al3mart Need to get commit signing sorted.

PR #13, opened 2024-02-28
ci: grant write permission to actions

evilrobot-01:
Grants permission to the workflow to annotate the checks with any clippy warnings.


peterwht (approved):
LGTM. Useful reference here: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

al3mart (approved):

PR #17, opened 2024-03-04
feat(pop-api): errors

peterwht:
Adds error handling approach to surface pallet errors from runtime calls.


pop-api/src/v0/nfts.rs
```diff
@@ -211,3 +213,161 @@ pub(crate) enum NftCalls {
...
+    MaxAttributesLimitReached,
+    /// The provided namespace isn't supported in this call.
+    WrongNamespace,
+    /// Can't delete non-empty collections.
+    CollectionNotEmpty,
+    /// The witness data should be provided.
+    WitnessRequired,
+}
+
+impl From<u32> for Error {
```
peterwht, line 312: TryFrom seems better here for graceful error handling. We don't want the contract to always panic if they get stuck receiving an invalid error.

pop-api/src/v0/nfts.rs
```diff
@@ -211,3 +213,161 @@ pub(crate) enum NftCalls {
...
+    fn from(value: u32) -> Self {
+        use Error::*;
+        match value {
```
peterwht, line 315: Using `num_derive` & `num_traits` would remove the need to do these manual matches.

pop-api/src/lib.rs
```diff
@@ -27,14 +28,18 @@ pub type Result<T> = core::result::Result<T, PopApiError>;
...
+    Nfts(nfts::Error),
 }
 
 impl ink::env::chain_extension::FromStatusCode for PopApiError {
     fn from_status_code(status_code: u32) -> core::result::Result<(), Self> {
         match status_code {
             0 => Ok(()),
-            1 => Err(Self::PlaceholderError),
+            1 => Err(Self::RuntimeError),
+            50_000..=50_999 => {
```
peterwht, line 40: We should probably introduce a const NFT_INDEX = 50.

pop-api/src/v0/nfts.rs
evilrobot-01: Agreed, done

pop-api/src/lib.rs
evilrobot-01: Gave it a go, it doesnt like (`nfts::PALLET_INDEX * 1_000`) in the match arm so had to revert.

pop-api/src/v0/nfts.rs
evilrobot-01: I'm personally hesitant to take a dependency when it can be implemented perfectly easily in local Rust code. Suggest we revist when the number of errors explode.

evilrobot-01 (approved):

al3mart (approved):
Looks good.

I have removed one unused import and renamed the tests so they are more descriptive.

peterwht:
LGTM!

PR #21, opened 2024-03-06
feat(pop-api): further implementation of nfts api

evilrobot-01:
Implements the remaining runtime calls that were listed, along with a few queries.

Notes: 
- Removes the offchain signing calls for now.
- Have not implemented those which require AccountId to be added to primitives (e.g. custom attribute). It results in a dependency mismatch between the runtime and that required by ink. Need to find a solution for this.
- Suggest moving `pop-api-primitives` to `pop-primitives` in a separate PR

### DIFF
--- a/pop-api/primitives/src/lib.rs
+++ b/pop-api/primitives/src/lib.rs
@@ -1,3 +1,18 @@
 #![cfg_attr(not(feature = "std"), no_std, no_main)]
 
+pub use bounded_collections::{BoundedBTreeMap, BoundedBTreeSet, BoundedVec, ConstU32};
+//use scale::{Decode, Encode, MaxEncodedLen};
+
 pub mod storage_keys;
+
+// /// Some way of identifying an account on the chain.
+// #[derive(Encode, Decode, Debug, MaxEncodedLen)]
+// pub struct AccountId([u8; 32]);
+// Id used for identifying non-fungible collections.
+pub type CollectionId = u32;
+// Id used for identifying non-fungible items.
+pub type ItemId = u32;
+/// The maximum length of an attribute key.
+pub type KeyLimit = ConstU32<64>;
+/// The maximum approvals an item could have.
+pub type ApprovalsLimit = ConstU32<20>;
--- a/pop-api/src/v0/nfts.rs
+++ b/pop-api/src/v0/nfts.rs
@@ -1,7 +1,7 @@
 use super::RuntimeCall;
 use crate::{PopApiError::UnknownStatusCode, *};
 use ink::prelude::vec::Vec;
-use primitives::MultiAddress;
+use primitives::{ApprovalsLimit, BoundedBTreeMap, CollectionId, ItemId, KeyLimit, MultiAddress};
 use scale::Encode;
 use types::*;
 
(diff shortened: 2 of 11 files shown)


evilrobot-01:
Converted to draft as it is missing the storage queries to make it more complete.

pop-api/primitives/src/lib.rs
```diff
@@ -1,3 +1,18 @@
 #![cfg_attr(not(feature = "std"), no_std, no_main)]
 
+pub use bounded_collections::{BoundedBTreeMap, BoundedBTreeSet, BoundedVec, ConstU32};
+//use scale::{Decode, Encode, MaxEncodedLen};
+
 pub mod storage_keys;
+
+// /// Some way of identifying an account on the chain.
+// #[derive(Encode, Decode, Debug, MaxEncodedLen)]
+// pub struct AccountId([u8; 32]);
```
peterwht, line 10: Remove comment if no longer necessary

pop-api/src/v0/nfts.rs
```diff
@@ -18,14 +35,370 @@ pub fn mint(
...
+	pub fn attribute(
+		collection: CollectionId,
+		item: ItemId,
+		key: BoundedVec<u8, KeyLimit>,
+	) -> Result<Option<Vec<u8>>> {
+		Ok(state::read(RuntimeStateKeys::Nfts(NftsKeys::Attribute(collection, item, key)))?)
+	}
+
+	// /// Get the custom attribute value of `item` of `collection` corresponding to `key`.
+	// pub fn custom_attribute(
```
peterwht, line 214: Why is this method commented out?

pop-api/src/v0/nfts.rs
```diff
@@ -320,3 +673,210 @@ impl From<PopApiError> for Error {
 		}
 	}
 }
+
+// Local implementations of pallet-nfts types
+mod types {
```
peterwht, line 678: Should this module be added to primitives?

runtime/src/extensions.rs
```diff
@@ -182,6 +194,52 @@ where
...
+			env.charge_weight(T::DbWeight::get().reads(1_u64))?;
+			Ok(pallet_nfts::Pallet::<T>::owner(collection, item).encode())
+		},
+		NftsKeys::Attribute(collection, item, key) => {
+			env.charge_weight(T::DbWeight::get().reads(1_u64))?;
+			Ok(pallet_nfts::Pallet::<T>::attribute(&collection, &item, &key).encode())
+		},
+		// NftsKeys::CustomAttribute(account, collection, item, key) => {
+		// 	env.charge_weight(T::DbWeight::get().reads(1_u64))?;
+		// 	Ok(pallet_nfts::Pallet::<T>::custom_attribute(&account, &collection, &item, &key)
```
peterwht, line 228: If this is commented out because `item` needs to be of type `Option<&ItemId>` you can use `item.as_ref()` to convert from a `Option<ItemId>` to `Option<&ItemId>`

peterwht (commented):
Looks really good! Left a few nit-picky comments

runtime/src/assets_config.rs
```diff
@@ -38,7 +38,9 @@ parameter_types! {
 
 impl pallet_nfts::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
+	// TODO: source from primitives
```
peterwht, line 41: Do we want to address the TODOs before merging?

pop-api/primitives/src/lib.rs
evilrobot-01: See comment in description about AccountId issues.

pop-api/src/v0/nfts.rs
evilrobot-01: As above, see comment in description about AccountId issues.

pop-api/src/v0/nfts.rs
evilrobot-01: No, these are the ink! versions of the types in `pallet_nfts`, so that the types can be decoded without need a dependency on `pallet_nfts` in contracts.

runtime/src/extensions.rs
evilrobot-01: As above, see comment in description about `AccountId` issues.

runtime/src/assets_config.rs
evilrobot-01: Need to refactor primitives as noted in the PR description and didnt want it to affect any work that Ale might be busy with right now.

al3mart: I'm not using primitives for anything cross-chain so far

evilrobot-01: It should be a quick and small PR, so better to address separately imo.

peterwht (approved):
Approved. See issues https://github.com/r0gue-io/pop-node/issues/27 and https://github.com/r0gue-io/pop-node/issues/26

al3mart (approved):
All looks good.

I guess we can address moving those primitives in the same PR that takes ParachainSystem read state logic out of the general `read_state()` function.

We are coming back to having modules for the different implementations of the extension ? ;)

PR #22, opened 2024-03-06
test(pop-api): read_state chain extension test

evilrobot-01:
Adds test to ensure that read_state chain extension works.

Notes:
- added `MaxEncodedLen` to key types to use `env.read_as()`
- charges a db read for reading the specified key from contract memory, to be replaced once benchmarked
- charges a db read for obtaining the last relay block
- removed unnecessary `pub` modifiers

### DIFF
--- a/runtime/src/extensions.rs
+++ b/runtime/src/extensions.rs
@@ -77,7 +77,7 @@ impl TryFrom<u16> for v0::FuncId {
 	}
 }
 
-pub(crate) fn dispatch<T, E>(env: Environment<E, InitState>) -> Result<(), DispatchError>
+fn dispatch<T, E>(env: Environment<E, InitState>) -> Result<(), DispatchError>
 where
 	T: pallet_contracts::Config + frame_system::Config,
 	<T as SysConfig>::AccountId: UncheckedFrom<<T as SysConfig>::Hash> + AsRef<[u8]>,
@@ -133,7 +133,7 @@ where
 	Ok(())
 }
 
-pub(crate) fn read_state<T, E>(env: Environment<E, InitState>) -> Result<(), DispatchError>
+fn read_state<T, E>(env: Environment<E, InitState>) -> Result<(), DispatchError>
 where
 	T: pallet_contracts::Config + frame_system::Config,
 	E: Ext<T = T>,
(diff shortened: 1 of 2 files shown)


peterwht (approved):

runtime/src/extensions.rs
```diff
@@ -439,4 +444,64 @@ mod tests {
...
+				0,
+				Weight::from_parts(100_000_000_000, 3 * 1024 * 1024),
+				None,
+				params,
+				DEBUG_OUTPUT,
+				pallet_contracts::CollectEvents::UnsafeCollect,
+				pallet_contracts::Determinism::Relaxed,
+			);
+
+			if DEBUG_OUTPUT == pallet_contracts::DebugInfo::UnsafeDebug {
```
Daanvdplas, line 495: This if statement feels a little unnecessary but maybe I'm not understanding it correctly

runtime/src/extensions.rs
```diff
@@ -208,7 +213,7 @@ mod tests {
 		Ok((wasm_binary, code_hash))
 	}
 
-	pub fn function_selector(name: &str) -> Vec<u8> {
+	fn function_selector(name: &str) -> Vec<u8> {
```
Daanvdplas, line 216: There is also the [ink::selector_bytes!](https://docs.rs/ink_lang/latest/ink_lang/macro.selector_bytes.html) macro

Daanvdplas (approved):
Well done! Left two comments

runtime/src/extensions.rs
evilrobot-01: Wouldn't want to take a dependency on ink! in the runtime.

runtime/src/extensions.rs
evilrobot-01: Agreed. Was something that was copied from existing tests above. Idea seems to be that changing global constant value has no affect on test output.

All these tests need clean up, especially as they rely on the contracts being built locally in the relevant paths and are therefore set as ignored to not break CI checks.

Don't think it warrants further time right now as there are more pressing things to get done.

runtime/src/extensions.rs
Daanvdplas: Yes figured that too while I was working out :)

PR #33, opened 2024-03-09
feat(pop-api): cross chain messaging

evilrobot-01:
Used more message-based terms to try provide clarity that it is not a sync call to local runtime, but a send of message to a runtime on another chain.

Note: rebased on main


peterwht (approved):

runtime/src/extensions.rs
```diff
@@ -99,62 +95,86 @@ impl TryFrom<u16> for v0::FuncId {
 	}
 }
 
+mod utils {
+	use super::*;
+
+	pub fn dispatch_call<T, E>(env: &mut Environment<E, BufInBufOutState>, call: <T as SysConfig>::RuntimeCall) -> Result<(), DispatchError>
```
evilrobot-01: Why pub? Everything is Rust is private by default, so overexposing functions when never required by an external crate is unneccessary imo. Consider pub(crate) or pub(super) to only grant 'just enough' access rather than overexposing. The utils module is indeed private, but if someone were to mistakenly overexpose that then suddenly all internal functions are publicly accessible.

https://doc.rust-lang.org/reference/visibility-and-privacy.html

runtime/src/extensions.rs
```diff
@@ -99,62 +95,86 @@ impl TryFrom<u16> for v0::FuncId {
...
+	where
+		T: frame_system::Config<RuntimeOrigin = RuntimeOrigin, RuntimeCall = RuntimeCall>,
+		RuntimeOrigin: From<RawOrigin<<T as SysConfig>::AccountId>>,
+		<T as SysConfig>::AccountId: UncheckedFrom<<T as SysConfig>::Hash> + AsRef<[u8]>,
+		E: Ext<T = T>,
+	{
+		const LOG_PREFIX: &str = " send_xcm |";
```
evilrobot-01, line 116: value doesnt seem correct for the function: dispatch call

runtime/src/extensions.rs
```diff
@@ -99,62 +95,86 @@ impl TryFrom<u16> for v0::FuncId {
...
+				Ok(())
+			},
+			Err(err) => {
+				log::debug!(target:LOG_TARGET, "{} failed: error: {:?}", LOG_PREFIX, err.error);
+				Err(err.error)
+			},
+		}
+	}
+
+	pub fn charge_overhead_weight<T, E>(env: &mut Environment<E, BufInBufOutState>, len: u32) -> Result<ChargedAmount, DispatchError>
```
evilrobot-01: visibility as above

runtime/src/extensions.rs
peterwht: The module utils is private, but the function needs to be public within the utils module

runtime/src/extensions.rs
```diff
@@ -306,35 +313,12 @@ where
 	};
 
 	// Generate runtime call to dispatch
-	let call = crate::RuntimeCall::PolkadotXcm(pallet_xcm::Call::send {
+	let call: <T as SysConfig>::RuntimeCall = <T as SysConfig>::RuntimeCall::PolkadotXcm(pallet_xcm::Call::send {
```
evilrobot-01: Does this extra complexity bring any benefit?

runtime/src/extensions.rs
```diff
@@ -99,62 +95,86 @@ impl TryFrom<u16> for v0::FuncId {
 	}
 }
 
+mod utils {
```
evilrobot-01, line 103: Subjective, but dont see the benefit to a module here. As a maintainer adding a new extension function, I need to now know that I need to call such function within some utils submodule, adding to the complexity rather than just a function that exists within the extension module itself. Utils seems more broad, in general usage beyond this extension which isnt the case.

I can see more benefit in some `dispatch` module now that the dispatch functionality is becoming overloaded. Likewise some module specific to reading state as those use cases grow.

evilrobot-01:
Cant add a formal review as the PR was created by me, but add some very opionated suggestions.

runtime/src/extensions.rs
peterwht: pub(super) was the right one!

runtime/src/extensions.rs
peterwht: No, it was stray code from trying to make it generic. Latest commit simplifies types.

runtime/src/extensions.rs
peterwht: I removed the utils module. I request we wait on moving dispatch out to a separate module until we start the Pop API pallet. The reason being that the extension should be come simpler as we move functionality to the pallet.

Daanvdplas (approved):
Amazing work! Curious how the structure will look like for `pop_api::xcm` in the future. Very exciting!

PR #34, opened 2024-03-09
fix(pop-api): read_state uses RelayChainDataProvider

al3mart:
Use [`RelaychainDataProvider`](https://github.com/paritytech/polkadot-sdk/blob/ea458d0b95d819d31683a8a09ca7973ae10b49be/cumulus/pallets/parachain-system/src/lib.rs#L1715) to retrieve the relay chain block number instead.


evilrobot-01 (approved):
LGTM, just needs updating so checks pass

peterwht (approved):

PR #38, opened 2024-03-11
test(integration-tests): placing order for instantaneous coretime

evilrobot-01:
A very hacky implementation which works around the issues of the transact failing due to the specified para id not being registered as on-demand. A potential solution is to manually set the raw storage with the required state.

Note also that the xcm send currently needs to be dispatched as root, rather than signed, due to the fact that the Rococo config does not seem to support creating a derivative/keyless account for some corresponding account on PopNet, which is not the case on Kusama/Polkadot.


Daanvdplas (approved):

integration-tests/src/lib.rs
```diff
@@ -431,6 +450,105 @@ fn reserve_transfer_native_asset_from_para_to_system_para() {
...
+			message: bx!(VersionedXcm::V4(message)),
+		})
+		// TODO: replace root with signed, currently prohibited by HashedDescription<AccountId, DescribeFamily<DescribeBodyTerminal>> (https://github.com/paritytech/polkadot-sdk/blob/a6713c55fd5082d333518c3ca13f2a4294726fcc/polkadot/runtime/rococo/src/xcm_config.rs#L67) rather than HashedDescription<AccountId, DescribeFamily<DescribeAllTerminal>> (https://github.com/polkadot-fellows/runtimes/blob/e42821da8d85f721d0dd1670dfb23f4dd91bd3e8/relay/kusama/src/xcm_config.rs#L76)
+		//.dispatch(RawOrigin::Signed(beneficiary).into());
+		.dispatch(RawOrigin::Root.into());
+
+		assert!(res.is_ok());
+		type RuntimeEvent = <PopNetworkPara as Chain>::RuntimeEvent;
+		// Check that the message was sent
+		assert_expected_events!(
```
peterwht, line 516: Should this be added to the `PopNetworkPara::execute_with` that is handling events already?

al3mart (approved):

integration-tests/src/chains/rococo/mod.rs
```diff
@@ -21,6 +21,10 @@ decl_test_relay_chains! {
 			Sudo: rococo_runtime::Sudo,
 			Balances: rococo_runtime::Balances,
 			Hrmp: rococo_runtime::Hrmp,
+			Registrar: rococo_runtime::Registrar,
+			Configuration: rococo_runtime::Configuration,
+			Paras: rococo_runtime::Paras,
+			ParasSudoWrapper: rococo_runtime::ParasSudoWrapper,
```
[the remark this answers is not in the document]
evilrobot-01: Sorry, just saw you addressed these, thank you!

integration-tests/src/lib.rs
evilrobot-01: Sorry, just saw you addressed these, thank you!

PR #39, opened 2024-03-12
feat: spot order contract

peterwht:
Also updates ink! deps to 5.0.0-rc3

PR #40, opened 2024-03-12
feat: spot order contract

peterwht:


evilrobot-01:
Converted to draft for now as #39  seems to carry these commits. Will revisit once the other is complete and see if any difference.

peterwht:
Closing. See https://github.com/r0gue-io/pop-node/pull/39

PR #45, opened 2024-03-21
feat: network token enforcement

evilrobot-01:
Network now launches with zero total issuance, requiring a reserve transfer from the relay to get started. This will be automated with a pop up feature, but for now one can reserve transfer from relay to Pop with the following encoded calls:
- Send 100 UNIT to Alice on Pop
`0x630804000100098e0400010100d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d04040000000b00407a10f35a0000000000`
- Send 100 UNIT to Bob on Pop
`0x630804000100098e04000101008eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a4804040000000b00407a10f35a0000000000`

Commits:
- filter `balances::force*` calls to prevent minting of native token, ensuring network can only be funded from relay token
- removes endowments from chainspec 
- removes endowments from genesis within integration tests, updating tests accordingly

### DIFF
--- a/integration-tests/src/chains/pop_network/genesis.rs
+++ b/integration-tests/src/chains/pop_network/genesis.rs
@@ -1,4 +1,4 @@
-use emulated_integration_tests_common::{accounts, build_genesis_storage, collators};
+use emulated_integration_tests_common::{build_genesis_storage, collators};
 use pop_runtime::Balance;
 use sp_core::storage::Storage;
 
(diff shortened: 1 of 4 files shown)


Daanvdplas (approved):
Very nice!

Should we add testing the integrations in the CI or is that too much?

peterwht (approved):
LGTM. We will probably want to introduce some tests later on for allowing users to withdraw tokens back to the relay chain

peterwht:
https://github.com/r0gue-io/pop-node/issues/47

evilrobot-01:
> LGTM. We will probably want to introduce some tests later on for allowing users to withdraw tokens back to the relay chain

Already exists: https://github.com/r0gue-io/pop-node/blob/9102be205e2201006b1f9e64e620bfd2a2ce0cf0/integration-tests/src/lib.rs#L325

peterwht:
> > LGTM. We will probably want to introduce some tests later on for allowing users to withdraw tokens back to the relay chain
> 
> Already exists:
> 
> https://github.com/r0gue-io/pop-node/blob/9102be205e2201006b1f9e64e620bfd2a2ce0cf0/integration-tests/src/lib.rs#L325

Ah nice. Missed that one and thought we were testing transfer to SA account.

evilrobot-01:
> Should we add testing the integrations in the CI or is that too much?

Already included - see https://github.com/r0gue-io/pop-node/actions/runs/8383155297/job/22958325605:
image

al3mart (approved):

PR #46, opened 2024-03-22
feat: separate runtimes

al3mart:
This PR separates the runtime into two different ones
- Devnet (meant to be used locally or privately)
- Testnet (meant to be deployed on public testnets)

It configures the corresponding chain specs and its configurations depending the relay chain the network is aimed to be deployed on:
- `dev-rococo`
- `dev-paseo`
- `pop-rococo`
- `pop-paseo`

Also includes the specs generated for `rococo` and `paseo` from this code base.

I have created a pair of accounts to have a place holder in the chain spec for the testnet options, I'll be curious to know your opinions there. Aura keys might need to be changed.

A new feature has been included `paseo` to set the corresponding runtime values for deploying in Paseo testnet.
This feature sets the runtime for sync backing and uses monetary unit of 10 decimals. 
Not activating this feature will default to set the runtime to work with async backing and a monetary unit of 12 decimals.


node/src/chain_spec.rs
```diff
@@ -59,29 +67,67 @@ where
...
-pub fn development_config() -> ChainSpec {
-	// Give your base currency a unit name and decimal places
-	let mut properties = sc_chain_spec::Properties::new();
-	properties.insert("tokenSymbol".into(), "UNIT".into());
-	properties.insert("tokenDecimals".into(), 12.into());
+fn configure_for_relay(
+	relay: Relay,
+	properties: &mut sc_chain_spec::Properties,
+) -> (Extensions, u32) {
+	properties.insert("tokenSymbol".into(), "POP".into());
```
peterwht, line 84: I think this `properties.insert("tokenSymbol".into(), "POP".into());` should be based on what relay it is. `POP` token does not exist.

node/src/command.rs
```diff
@@ -13,16 +13,78 @@ use sp_runtime::traits::AccountIdConversion;
...
+		path => {
+			let path: PathBuf = path.into();
+			match path.runtime() {
+				Runtime::Devnet | Runtime::Default => {
+					Box::new(chain_spec::DevnetChainSpec::from_json_file(path)?)
+				},
+				Runtime::Testnet => Box::new(chain_spec::TestnetChainSpec::from_json_file(path)?),
+			}
+		},
+		_ => {
```
peterwht, line 84:
```
84 |         _ => {
   |         ^ unreachable pattern
```

peterwht (changes requested):
`cargo test` build fails, especially with the `integration-tests` folder

node/src/chain_spec.rs
al3mart: Ah you are so right. This needs to be changes all across all the properties. Thanks.

integration-tests/src/chains/pop_network/genesis.rs
```diff
@@ -1,43 +1,43 @@
 use emulated_integration_tests_common::{accounts, build_genesis_storage, collators};
-use pop_runtime::Balance;
+use pop_runtime_common::Balance;
 use sp_core::storage::Storage;
 
-pub(crate) const ED: Balance = pop_runtime::EXISTENTIAL_DEPOSIT;
+pub(crate) const ED: Balance = pop_runtime_devnet::EXISTENTIAL_DEPOSIT;
```
evilrobot-01, line 5: We should ideally alias `pop_runtime_devnet` to `runtime` to simplify the rest of the code in the file. Something like `use pop_runtime_devnet as runtime`

integration-tests/src/chains/pop_network/mod.rs
```diff
@@ -13,19 +13,19 @@ decl_test_parachains! {
 	pub struct PopNetwork {
 		genesis = genesis::genesis(),
 		on_init = {
-			pop_runtime::AuraExt::on_initialize(1);
+			pop_runtime_devnet::AuraExt::on_initialize(1);
```
evilrobot-01, line 16: Consider runtime alias.

integration-tests/src/chains/rococo/genesis.rs
```diff
@@ -3,7 +3,7 @@ use emulated_integration_tests_common::{
 	validators,
 };
 use polkadot_primitives::{AssignmentId, ValidatorId};
-use pop_runtime::Balance;
+use pop_runtime_common::Balance;
```
evilrobot-01, line 6: This seems weired, Rococo genesis shouldnt need to know about pop. Seems it isnt being used in any case, so can probably be removed.

integration-tests/src/lib.rs
```diff
@@ -458,15 +459,15 @@ fn place_coretime_spot_order_from_para_to_relay() {
 	let beneficiary: sp_runtime::AccountId32 = [1u8; 32].into();
 
 	// Setup: reserve transfer from relay to Pop, so that sovereign account accurate for return transfer
-	let amount_to_send: Balance = pop_runtime::UNIT * 1000;
+	let amount_to_send: Balance = pop_runtime_devnet::UNIT * 1000;
```
evilrobot-01, line 462: As mentioned elsewhere, maybe an alias of `pop_runtime_devnet` to `pop_runtime` to simplify?

integration-tests/Cargo.toml
```diff
@@ -38,7 +38,9 @@ cumulus-primitives-core.workspace = true
 emulated-integration-tests-common.workspace = true
 
 # Local
-pop-runtime.workspace = true
+pop-runtime-devnet.workspace = true
+pop-runtime-testnet.workspace = true
```
evilrobot-01, line 42: I dont think we need the testnet runtime included for now, probably just adds to compilation time

node/Cargo.toml
```diff
@@ -20,7 +20,9 @@ futures = { workspace = true }
 serde_json = { workspace = true }
 
 # Local
-pop-runtime = { workspace = true }
+pop-runtime-devnet = { workspace = true }
```
evilrobot-01, line 23: Be nice to use `.workspace = true style`

node/src/chain_spec.rs
```diff
@@ -1,19 +1,27 @@
...
+pub type TestnetChainSpec =
+	sc_service::GenericChainSpec<pop_runtime_testnet::RuntimeGenesisConfig, Extensions>;
 
 /// The default XCM version to set in genesis config.
 const SAFE_XCM_VERSION: u32 = xcm::prelude::XCM_VERSION;
-
-/// The local para ID of Pop Network.
-const PARA_ID: u32 = 9090;
+pub(crate) enum Relay {
+	ROC,
```
evilrobot-01, line 20: Can we use the full network names here instead please? i.e. Rococo, Paseo, RococoLocal and PaseoLocal.

runtime/testnet/Cargo.toml
```diff
@@ -0,0 +1,214 @@
+[package]
+name = "pop-runtime-testnet"
+version = "0.1.0"
+authors = { workspace = true }
+description = "A new Cumulus FRAME-based Substrate Runtime, ready for hacking together a parachain."
```
evilrobot-01, line 5: Can we update this to the repo description: "Pop Network makes it easy for smart contract developers to use the Power of Polkadot. "

runtime/devnet/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "pop-runtime"
+name = "pop-runtime-devnet"
 version = "0.1.0"
 authors = { workspace = true }
 description = "A new Cumulus FRAME-based Substrate Runtime, ready for hacking together a parachain."
```
evilrobot-01, line 5: Can we update this to the repo description? "Pop Network makes it easy for smart contract developers to use the Power of Polkadot."

runtime/testnet/src/extensions.rs
```diff
@@ -0,0 +1,740 @@
+use cumulus_pallet_parachain_system::RelaychainDataProvider;
```
evilrobot-01, line 1: I think we need to refactor the chain extension into common in a separate PR if possible.

runtime/testnet/src/lib.rs
```diff
@@ -0,0 +1,923 @@
...
+	type SessionHandler = <SessionKeys as sp_runtime::traits::OpaqueKeys>::KeyTypeIdProviders;
+	type Keys = SessionKeys;
+	type WeightInfo = ();
+}
+
+impl pallet_aura::Config for Runtime {
+	type AuthorityId = AuraId;
+	type DisabledValidators = ();
+	type MaxAuthorities = ConstU32<100_000>;
+	type AllowMultipleBlocksPerSlot = ConstBool<true>;
```
evilrobot-01, line 441: I believe this needs to be addressed, as I'm not sure this will work on Paseo? Suggest a feature flag, perhaps `async-backing` on the testnet runtime to toggle async backing on or off, depending on which network it is being deployed to...

evilrobot-01 (changes requested):
LGTM, mostly just suggestions apart from the potential issue with async-backing enabled on Paseo. We might need to switch all the items noted in https://wiki.polkadot.network/docs/maintain-guides-async-backing#phase-3---activate-async-backing off when deploying to Paseo, which could be done with an `async-backing` feature flag.

integration-tests/src/chains/pop_network/genesis.rs
al3mart: This is good idea

node/Cargo.toml
al3mart: Can refactor in a different PR

al3mart:
Will address your comments @evilrobot-01 and maybe also move the definition of `UNITS` together with a feature so we can choose the precision we want to use.

I've just noticed that even if we are setting the spec correctly we still have this fixed definition in the runtime.

evilrobot-01:
> Will address your comments @evilrobot-01 and maybe also move the definition of `UNITS` together with a feature so we can choose the precision we want to use.

Probably need a better feature name than async-backing then. As Rococo is short term, maybe we just use `rococo`, with the defaults being for Paseo.

node/chain_specs/pop-rococo.plain.json
brunopgalvao, line 1: Do we need the plain chain spec uploaded? It is different then what `node/src/chain_spec.rs` generates?

Daanvdplas: No but it is nice to have somewhere. Since the chain is live and public, it can be good to have the chainspec somewhere to easily launch a collator and join the network.

node/src/chain_spec.rs
```diff
@@ -59,29 +67,70 @@ where
...
+			para_id = 4385;
+			properties.insert("tokenSymbol".into(), "ROC".into());
+			properties.insert("tokenDecimals".into(), 12.into());
+
+			(Extensions { relay_chain: "rococo".into(), para_id }, para_id)
+		},
+		Relay::PAS => {
+			para_id = 4001;
+			properties.insert("tokenSymbol".into(), "PAS".into());
+			properties.insert("tokenDecimals".into(), 10.into());
```
Daanvdplas, line 99: Why does Paseo have 10 decimals in stead of 12?

Daanvdplas:
A suggestion is to diff the lib.rs file of both testnet & devnet, and perhaps the others just to be sure, and check for inconsistencies. I see the struct `FilteredCalls` missing in the devnet runtime/lib.rs now

node/chain_specs/pop-rococo.plain.json
evilrobot-01: Only need the raw spec to join as a collator, but plain spec can be required by zombienet when not able to be built via `build-spec`, which is the case now with some runtimes in the fellowship repo.

node/src/chain_spec.rs
evilrobot-01: Same as Polkadot

evilrobot-01:
> A suggestion is to diff the lib.rs file of both testnet & devnet, and perhaps the others just to be sure, and check for inconsistencies.

Agreed, it is important that these two align, at least in terms of structure, so that they can easily be diffed to ensure certain things are not missing. They will diverge over time, but need to be able to easily diff to ensure what should be the same, is the same.

node/src/chain_spec.rs
al3mart: Correct, it follows Polkadot runtime.

runtime/testnet/src/extensions.rs
al3mart: Agree

peterwht (commented):
Overall looks good to me. Agree with Frank's comments from above about using runtime alias to simplify naming.

integration-tests/src/chains/pop_network/genesis.rs
evilrobot-01: Addressed in a PR to this one.

integration-tests/src/chains/pop_network/mod.rs
evilrobot-01: Addressed in a PR to this one.

integration-tests/src/lib.rs
evilrobot-01: Addressed in a PR to this one.

node/Cargo.toml
evilrobot-01: Was quick to do so addressed in a PR to this one so its done.

node/src/chain_spec.rs
evilrobot-01: Clippy complained about capitalised variant names so used title case instead, addressed in a PR to this one.

node/src/chain_spec.rs
```diff
@@ -93,47 +142,62 @@ pub fn development_config() -> ChainSpec {
...
+			get_account_id_from_seed::<sr25519::Public>("Charlie"),
+			get_account_id_from_seed::<sr25519::Public>("Dave"),
+			get_account_id_from_seed::<sr25519::Public>("Eve"),
+			get_account_id_from_seed::<sr25519::Public>("Ferdie"),
+			get_account_id_from_seed::<sr25519::Public>("Alice//stash"),
+			get_account_id_from_seed::<sr25519::Public>("Bob//stash"),
+			get_account_id_from_seed::<sr25519::Public>("Charlie//stash"),
+			get_account_id_from_seed::<sr25519::Public>("Dave//stash"),
+			get_account_id_from_seed::<sr25519::Public>("Eve//stash"),
+			get_account_id_from_seed::<sr25519::Public>("Ferdie//stash"),
```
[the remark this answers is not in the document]
evilrobot-01: Addressed in a PR to this one.

evilrobot-01 (commented):
Added a PR to this one, addressing the outstanding comments. Once that is accepted and merged into this one, I am good to proceed.

evilrobot-01 (approved):

PR #49, opened 2024-03-24
feat(pop-api): add call filter

peterwht:
Adds an allowed list call filter to chain extension.

`pallet-contracts`'s `CallFilter` is not being used because it will allowb contracts to directly use `call_runtime`. Once the pop api pallet is introduced, the call filter should be implemented there.

There is a new example contract for testing purposes that tests that the call filter works.


runtime/src/lib.rs
```diff
@@ -315,6 +315,14 @@ impl Contains<RuntimeCall> for FilteredCalls {
 	}
 }
 
+/// A type to identify allowed calls to the Runtime from contracts. Used by Pop API
+pub struct AllowedApiCalls;
+impl Contains<RuntimeCall> for crate::AllowedApiCalls {
+	fn contains(c: &RuntimeCall) -> bool {
+		matches!(c, RuntimeCall::Balances { .. } | RuntimeCall::Nfts { .. })
```
evilrobot-01, line 322: I think we must be explicit with the individual calls here, a call should only be added if there is a corresponding implementation within the extension/api which requires it.

pop-api/examples/filtered-call/lib.rs
```diff
@@ -0,0 +1,110 @@
+#![cfg_attr(not(feature = "std"), no_std, no_main)]
```
evilrobot-01, line 1: This shouldn't be in `examples` as it is not showing anything useful to builder. As they are only being used for tests, suggest moving to a `tests/contracts` directory in the root, ala Rust integration tests.

pop-api/examples/filtered-call/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "pop_api_filtered_call"
+version = "0.1.0"
+authors = ["[your_name] <[your_email]>"]
+edition = "2021"
+
+[dependencies]
+ink = { version = "5.0.0-rc.3", default-features = false }
```
evilrobot-01, line 8: Can we use official version now?

pop-api/examples/filtered-call/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "pop_api_filtered_call"
+version = "0.1.0"
```
evilrobot-01, line 3: version is optional, can remove.

pop-api/examples/filtered-call/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+authors = ["[your_name] <[your_email]>"]
```
evilrobot-01, line 4: Can remove authors imo

pop-api/examples/filtered-call/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+#[derive(scale::Encode)]
+pub enum RuntimeCall {
+	#[codec(index = 0)]
+	System(SystemCalls),
+}
+
+pub type Result<T> = core::result::Result<T, ContractError>;
+#[ink::chain_extension(extension = 909)]
+pub trait PopApi {
```
evilrobot-01, line 33: This might be confusing imo, suggest renaming to something like `MyPopApi` to differentiate

pop-api/examples/filtered-call/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+/// BLOCKED the call. An Err() means that the filter did not block the call,
+/// which is incorrect behavior.
+/// The pop-api crate is not used because it does not support invalid calls ;)
+use scale::Encode;
+
+use ink::{env::Environment, prelude::vec::Vec};
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq, scale::Encode, scale::Decode)]
+#[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+pub enum ContractError {
```
evilrobot-01, line 16: Leaks types unnecessaritly, should be pub(crate) imo. Get its only a test, but useful to be mindful of not overexposing things.

pop-api/src/lib.rs
```diff
@@ -32,6 +33,7 @@ impl ink::env::chain_extension::FromStatusCode for PopApiError {
 	fn from_status_code(status_code: u32) -> core::result::Result<(), Self> {
 		match status_code {
 			0 => Ok(()),
+			5 => Err(PopApiError::CallFiltered),
```
evilrobot-01, line 36: Where does this value originate from? Assume it is related to some call filtered error via dispatch, but might need to ensure that this won't change if being relied upon from subsystem.

evilrobot-01 (changes requested):
Allowed calls should be explicit rather than module-wide and the included test contract shouldnt be located within examples.

pop-api/examples/filtered-call/lib.rs
peterwht: Unfortunately, ink! requires these types to be public. Because the ink! messages are required to be pub, the return type can not be private.

pop-api/src/lib.rs
peterwht: Originates from frame_system. Added comment and updated name.

tests/contracts/filtered-call/lib.rs
```diff
@@ -0,0 +1,116 @@
+#![cfg_attr(not(feature = "std"), no_std, no_main)]
+
+/// This is an example contract used for unit testing purposes.
+/// It is intended to ensure that the call filter for the Pop API chain
+/// extension works as expected.
+/// An Ok(()) returned from `get_filtered` means that the filter successfully
+/// BLOCKED the call. An Err() means that the filter did not block the call,
+/// which is incorrect behavior.
+/// The pop-api crate is not used because it does not support invalid calls ;)
```
Daanvdplas, line 9: I don't get this line. It does not support *filtered calls?

True, but is that not what this test is for? Devs using pop api and not being able to use this call in their contracts?

Daanvdplas (approved):

al3mart (commented):
It would be good to set the same filter for pallet-contracts in general ?

In case some contracts decide to not use `pop-api` to call the runtime.

tests/contracts/filtered-call/lib.rs
peterwht: The call defined in the contract is filtered (blocked) by the call filter. We do not want anything in the pop-api that is not a valid call (i.e. filtered)

peterwht:
> It would be good to set the same filter for pallet-contracts in general ?
> 
> In case some contracts decide to not use `pop-api` to call the runtime.

We do not want to expose call_runtime directly because then we may break contracts with runtime upgrades. If we were to utilize the CallFilter in pallet-contracts, we would have to ONLY allow calls going into the Pop API pallet. Then we can filter a second layer in to ensure the next call is allowed

al3mart (approved):

evilrobot-01 (approved):
LGTM. Would ideally be good to address the clippy warnings in `extensions.rs`, perhaps when resolving conflicts after runtime split is merged.

al3mart (approved):

PR #52, opened 2024-03-26
refactor: separate runtimes

evilrobot-01:
Various minor improvements identified during review, along with one fix of a regression in functionality caused by the split.

### DIFF
--- a/node/src/chain_spec.rs
+++ b/node/src/chain_spec.rs
@@ -142,20 +142,6 @@ pub fn development_config(relay: Relay) -> DevnetChainSpec {
 				get_collator_keys_from_seed("Bob"),
 			),
 		],
-		vec![
-			get_account_id_from_seed::<sr25519::Public>("Alice"),
-			get_account_id_from_seed::<sr25519::Public>("Bob"),
-			get_account_id_from_seed::<sr25519::Public>("Charlie"),
-			get_account_id_from_seed::<sr25519::Public>("Dave"),
-			get_account_id_from_seed::<sr25519::Public>("Eve"),
-			get_account_id_from_seed::<sr25519::Public>("Ferdie"),
-			get_account_id_from_seed::<sr25519::Public>("Alice//stash"),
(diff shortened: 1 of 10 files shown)


peterwht (approved):

Daanvdplas (approved):

node/src/chain_spec.rs
```diff
@@ -142,20 +142,6 @@ pub fn development_config(relay: Relay) -> DevnetChainSpec {
...
-			get_account_id_from_seed::<sr25519::Public>("Dave"),
-			get_account_id_from_seed::<sr25519::Public>("Eve"),
-			get_account_id_from_seed::<sr25519::Public>("Ferdie"),
-			get_account_id_from_seed::<sr25519::Public>("Alice//stash"),
-			get_account_id_from_seed::<sr25519::Public>("Bob//stash"),
-			get_account_id_from_seed::<sr25519::Public>("Charlie//stash"),
-			get_account_id_from_seed::<sr25519::Public>("Dave//stash"),
-			get_account_id_from_seed::<sr25519::Public>("Eve//stash"),
-			get_account_id_from_seed::<sr25519::Public>("Ferdie//stash"),
-		],
```
al3mart, line 158: Do we really don't want to fund these for the development config ?

evilrobot-01: I feel we are only going to find issues by working with it in the proper way from the beginning. I'd rather discover issues early and address them with time than discover them late, potentially once deployed. For example, what is the minimum requirement for collators to actually produce blocks, can we produce blocks without funding to have relay tokens transferred in etc. How do we bootstrap with some limited amount.

There is an additional PR from Peter to automate the onboarding, simulating how it will work in practice. I hope that this will be included in the binary and be triggered by a step in `pop up` so that we don't have to do anything - our tooling takes care of all of it.

A little additional work, but it proves that we are doing it the right way from the beginning, with the correct checks in place.

Issue #47, opened 2024-03-22
feat: subxt script to reserve transfer DOT to Pop


peterwht:
See https://github.com/r0gue-io/pop-node/pull/45

Daanvdplas:
Closed by #53

PR #53, opened 2024-03-27
feat: subxt script to fund dev accounts via XCM

peterwht:
https://github.com/r0gue-io/pop-node/issues/47


Cargo.toml
```diff
@@ -13,10 +13,11 @@ members = [
     "node",
     "runtime",
     "integration-tests",
+    "scripts/fund-dev-accounts",
```
Daanvdplas, line 16: Perhaps exclude this one as well? It now always compiles the script when compiling the node binary which includes subxt.

scripts/fund-dev-accounts/main.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+use relay::*;
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+	let relay_api = OnlineClient::<PolkadotConfig>::from_url("ws://127.0.0.1:8833").await?;
+
+	let dev_accounts = vec![dev::alice(), dev::bob(), dev::charlie()];
+	let fund_pop_accounts =
```
Daanvdplas, line 124: `fund_pop_accounts_calls`

Daanvdplas (commented):
Nicely done! One thing to add is three queries to Pop Network to ensure that alice, bob and charlie have received the funds.

evilrobot-01:
Just thinking about this a little, would it not make sense to add this as a command to the `pop-node` binary? e.g. `pop-node init`.

It then becomes a first-class citizen, usable whenever a network launched, rather than compiling another side-car executable and manually launching. It could remain as its own crate, e.g. pop-node-init and be wrapped in a default feature so that it can be compiled out for releases if necessary.

I'm just thinking about the UX via `pop-cli`, where currently the node is automatically sourced and built. A node would be launched, but would not have any funds and users would not be able to deploy any contracts.

Also - have the `.scale` files been optimised to only include metadata for the pallets which are actually used? They seem large and I am sure that there is an option to trim them done to only the functionality which is used.

scripts/fund-dev-accounts/main.rs
```diff
@@ -0,0 +1,153 @@
...
+		},
+		xcm::{v3::WeightLimit, VersionedAssets, VersionedLocation},
+	};
+
+	// generate XCM message to reserve transfer funds to a designated account on
+	// Pop Parachain
+	pub(crate) fn gen_account_fund_message_call(account: Keypair) -> RuntimeCall {
+		let pop_location = VersionedLocation::V4(Location {
+			parents: 0,
+			interior: X1([Junction::Parachain(9090)]),
```
al3mart, line 35: We might want to use the paraId set in the chainspec for pop network on rococo local here https://github.com/r0gue-io/pop-node/pull/46

`4385`

Changing the one in the zombienet file would be nice as well.

scripts/fund-dev-accounts/main.rs
```diff
@@ -0,0 +1,153 @@
...
+			},
+			VersionedMultiAssets as VersionedAssets, VersionedMultiLocation as VersionedLocation,
+		},
+	};
+
+	// generate XCM message to reserve transfer funds to a designated account on
+	// Pop Parachain
+	pub(crate) fn gen_account_fund_message_call(account: Keypair) -> RuntimeCall {
+		let pop_location =
+			VersionedLocation::V3(Location { parents: 0, interior: X1(Junction::Parachain(9090)) });
```
al3mart, line 91: As above, `9090` should be changed in line with what is set in the spec.

(unfortunately)

peterwht:
> Also - have the `.scale` files been optimised to only include metadata for the pallets which are actually used? They seem large and I am sure that there is an option to trim them done to only the functionality which is used.

Unfortunately, it seems subxt has a bug with trimming metadata files (or I'm doing something wrong). Please see issue: https://github.com/paritytech/subxt/issues/1509

peterwht:
> Just thinking about this a little, would it not make sense to add this as a command to the `pop-node` binary? e.g. `pop-node init`.

Love the idea. This will be done in a separate PR as time allows. See tracker https://github.com/r0gue-io/pop-node/issues/57

scripts/fund-dev-accounts/main.rs
```diff
@@ -119,13 +124,18 @@ mod relay {
 }
 
 use relay::*;
+pub(crate) const AMOUNT_TO_FUND: u128 = UNIT * 1_000_000;
+
+#[subxt::subxt(runtime_metadata_path = "../metadata/pop-net.scale")]
```
evilrobot-01, line 129: This seems an oversight?

evilrobot-01 (commented):
Seems the pop interface is missing?

Daanvdplas:
Everything works! Got an error `messageQueue.ProcessingFailed` but I think that was due `xcmPallet.VersionNotifyRequested`. The accounts were funded and the local network is set up as it should. One nitpick is the output:

Screenshot 2024-04-01 at 12 45 27

It would be nice to make this look plain and simple so that the developer can simply read what happened and doesn't get confused. E.g.:
```
XCM Reserved Transfer sent from relay.
Checking Pop Balances:
Alice: amount
Bob: amount
Charlie: amount
```
Another thing is documentation. We should explain simple and short what is happening when someone spins up a local network with Pop Network. 

Both docs & output can be done in issue #57 as well

evilrobot-01 (approved):
LGTM. Looking forward to seeing it as a `pop-node` command!

evilrobot-01:
> Everything works! Got an error `messageQueue.ProcessingFailed` but I think that was due `xcmPallet.VersionNotifyRequested`. The accounts were funded and the local network is set up as it should. One nitpick is the output:

Interesting that the output includes Rococo, which should not be the case?

Daanvdplas (approved):

PR #74, opened 2024-04-12
docs: add pop api documentation

brunopgalvao:
ditto

Please be hard on the feedback so we can make this perfect.

### DIFF
--- a/pop-api/README.md
+++ b/pop-api/README.md
@@ -1,1 +1,2 @@
-Reserved crate for pop-api.
+# Pop API
+
--- a/pop-api/README.md
+++ b/pop-api/README.md
@@ -1,2 +1,73 @@
 # Pop API
 
+## Examples
+ink! smart contract examples using the Pop API
+- [balance-transfer](./examples/balance-transfer/)
+- [NFTs](./examples/nfts/)
+- [place-spot-order](./examples/place-spot-order/)
+- [read-runtime-state](./examples/read-runtime-state/)
+
+## What is the Pop API?
+
+One of the core value propositions of the Pop Network is to enable smart contracts to easily access the power of Polkadot. As such, the Pop API was built to enable smart contracts to easily utilize the functionality provided by the Pop Network parachain runtime.
+
+Substrate already exposes a Runtime API which is typically associated with the “outer node” calling into the runtime via RPC. Pop Network extends this concept in a direction that makes it usable by smart contracts, allowing untrusted contracts to make use of carefully exposed trusted functionality available within the runtime - this is the Pop API.
+
+The Pop API is designed to be:
+- **Stable**: contracts are generally immutable and therefore need a stable interface, especially as the runtime evolves over time.
+- **Future Proof**: technologies improve, so the Pop API implementation is adaptable to new approaches without affecting the interface. The Pop API implementation also selects the optimal approach to realize a use case from a range of possible technical implementations.
+- **Versioned**: not every future technology can be retro-fitted, so the API is versioned.
+- **Simple**: providing an abstraction layer over runtime-level features. Hand-rolled with love, Pop API is easy to use for smart contract development.
+
+The Pop API consists of three main parts:
+- the Pop API ink! library
+- the Pop API runtime code (chain extension)
+- shared primitive types between the Pop API ink! library and the Pop API runtime code
+
+Let's go over the flow.
+
+## The Pop API ink! Library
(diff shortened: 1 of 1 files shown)


pop-api/README.md
```diff
@@ -1 +1,73 @@
...
+- [balance-transfer](./examples/balance-transfer/)
+- [NFTs](./examples/nfts/)
+- [place-spot-order](./examples/place-spot-order/)
+- [read-runtime-state](./examples/read-runtime-state/)
+
+## What is the Pop API?
+
+One of the core value propositions of the Pop Network is to enable smart contracts to easily access the power of Polkadot. As such, the Pop API was built to enable smart contracts to easily utilize the functionality provided by the Pop Network parachain runtime.
+
+Substrate already exposes a Runtime API which is typically associated with the “outer node” calling into the runtime via RPC. Pop Network extends this concept in a direction that makes it usable by smart contracts, allowing untrusted contracts to make use of carefully exposed trusted functionality available within the runtime - this is the Pop API.
```
Daanvdplas, line 14: I have my reservations here. Not sure whether they are comparable, and the relevance of comparing it with the "runtime api of the outer node."

Daanvdplas (commented):
Nicely done!

evilrobot-01:
Sadly I think we should wait until we downgrade `pop-api` to ink! v4 before we can merge this.

brunopgalvao:
> Sadly I think we should wait until we downgrade `pop-api` to ink! v4 before we can merge this.

Bumping the PR. Now that we will not be downgrading to `inkv4` I believe we can merge this.

pop-api/README.md
Daanvdplas: Still think that this whole bit is unnecessary. It is complicated and not really something a smart contract developer needs to understand.

pop-api/README.md
```diff
@@ -1 +1,73 @@
...
+## The Glue
+
+Certain types are shared between the ink! library portion of the Pop API and the runtime portion of the Pop API. These types can be found in [`pop_primitives`](../primitives/src/), outside the [`pop-api`](./src/) folder.
+
+## The Entry into the Pop API
+
+When we use the Pop API in our smart contract like so:
+```rust
+use pop_api::nfts
+pop_api::nfts::mint(collection_id, item_id, receiver)?;
```
Daanvdplas, line 46: `pop_api::nfts` is not necessary if you look at the examples.

pop-api/README.md
```diff
@@ -1 +1,73 @@
...
+```
+
+This will call the Pop API `mint` function in the [./src/v0/nfts.rs](./src/v0/nfts.rs) file, which is a wrapper to `dispatch` a `Runtime::Call` to the NFTs pallet's mint function. This is how most of the Pop API is built. This abstraction allows for creating a nice API over runtime level features such as calling pallets, reading state, and cross-chain interactions. All Pop API functionality can be found in [./src/v0/](./src/v0/) which is the current version of the Pop API.
```
Daanvdplas, line 49: "nice" => "user friendly"

pop-api/README.md
```diff
@@ -1 +1,73 @@
...
+
+## The Pop API ink! Library
+Everything in [`pop-api`](./src/) **is the** Pop API ink! library.
+
+So when the ink! smart contract wants to use the Pop API library, it can simply have a line like:
+```rust
+use pop_api::nfts;
+```
+
+## The Glue
```
Daanvdplas, line 37: Remove section. Not very important to understand for the smart contract developer and not theoretically true since it is much more than just the types in primitives

pop-api/README.md
brunopgalvao: Updated.

pop-api/README.md
brunopgalvao: I disagree. This is part is essential. It is the bridge btwn ink! pop-api and runtime pop-api.

Daanvdplas (approved):

pop-api/README.md
Daanvdplas: It is not special to pop api of pop network though. This is already possible since chain extensions were developed

pop-api/README.md
Daanvdplas: These types can also be found in the pop api so you would have to add that then.

AlexD10S (approved):
Nice! 
I suggest you add this text (Maybe without the links to the code) in our Documentation page too: https://learn.onpop.io/pop/welcome/pop-api

al3mart (approved):

PR #75, opened 2024-04-13
build(release): add macos support

evilrobot-01:
Reworks the release workflow to include macOS binaries (Apple Silicon and x64), along with aarch64 on Linux.

Examples at https://github.com/r0gue-io/pop-node/releases/tag/test, where binary for each release is compressed and labelled using `pop-node-(target)(-cpu)` convention, making the files much smaller to download and more importantly, all use `pop-node` once extracted.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,9 +8,12 @@ on:
       ref:
         description: ref to build binary from
         required: false
+  push:
+    branches:
+      - 'frank/build-macos'
 
 jobs:
-  build-node:
+  build-node-for-linux:
     runs-on: ubuntu-22.04
     permissions:
       contents: write
@@ -24,6 +27,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
+          fetch-depth: 0
           ref: ${{ github.event.inputs.ref }}
 
       - name: Install packages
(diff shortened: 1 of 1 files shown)


evilrobot-01:
Naming of the binaries isnt great, would be better to compress them so the archive name is unique in the release but once downloaded and extracted we just have pop-node.

evilrobot-01:
Pushed a change to compress the resulting mac binaries as a trial: https://github.com/r0gue-io/pop-node/actions/runs/8679195330

al3mart (approved):

al3mart:
Note that while I could review the new workflow, I couldn't run the resulting binaries.
It would be great if another person could verify the result is expected

al3mart (commented):
We might want to add `--features paseo` for now :sweat_smile:

.github/workflows/release.yml
```diff
@@ -8,48 +8,83 @@ on:
...
+        run: |
+          brew install protobuf
+          protoc --version
+
+      - name: Add target
+        run: rustup target add ${{ matrix.platform.target }}
+
       - name: Build node
-        run: cargo build --profile=production -p pop-node
+        run: cargo build --profile=production -p pop-node --target ${{ matrix.platform.target }}
```
al3mart, line 63:
```suggestion
        run: cargo build --profile=production -p pop-node --features paseo --target ${{ matrix.platform.target }}
```

evilrobot-01: Done!

peterwht (approved):

PR #81, opened 2024-04-22
style: trust backed assets naming

Daanvdplas:
This is how it is done in assethub, changes the naming from Assets to TrustBackedAssets

### DIFF
--- a/runtime/devnet/src/assets_config.rs
+++ b/runtime/devnet/src/assets_config.rs
@@ -1,6 +1,6 @@
 use crate::{
-	deposit, AccountId, Assets, Balance, Balances, BlockNumber, Nfts, Runtime, RuntimeEvent,
-	RuntimeHoldReason, DAYS, EXISTENTIAL_DEPOSIT, UNIT,
+	deposit, AccountId, Balance, Balances, BlockNumber, Nfts, Runtime, RuntimeEvent,
+	RuntimeHoldReason, TrustBackedAssets, DAYS, EXISTENTIAL_DEPOSIT, UNIT,
 };
 use frame_support::{
 	parameter_types,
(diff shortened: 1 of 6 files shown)


brunopgalvao:
Looks like you need to update the naming in this integration test:
https://github.com/r0gue-io/pop-node/blob/2229baeb0a0cfdb6c37a4bb941e51454c73f22a3/integration-tests/src/chains/pop_network/mod.rs#L27

Daanvdplas:
This macro only accepts `Assets`: https://github.com/r0gue-io/pop-node/blob/2229baeb0a0cfdb6c37a4bb941e51454c73f22a3/integration-tests/src/chains/pop_network/mod.rs#L36 ...

brunopgalvao:
Keeping it as `Assets` is simple and it is also the pallet name. I see no need for naming it `TrustBackedAssets`. But I am also not against it if someone feels strongly about it.

evilrobot-01:
> I think it makes sense for AssetHub to have `TrustBackedAssets` because AssetHub is considered a "reserve".

I dont think that is the necessarily the case, but rather it seems to have stemmed from asset instancing a long time back and needing to differentiate more clearly in the difference between instances - https://github.com/paritytech/cumulus/pull/1908. I understand it to be more as a 'buyer beware' in that anyone can create assets and correspondingly freeze or destroy those assets. Using the asset implies some trust that the owner will not do as much, at least depending on how the asset is create/configured.

The phrasing at https://github.com/polkadot-fellows/runtimes/blob/5cf99f383c6692feeb26647a3ec544d80b66eaae/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L300-L302 alludes more to stablecoins where indeed there is trust that the asset is backed offchain with some corresponding fiat.

I think there could be some benefit in going against the status quo of just using assets to convey these additional considerations, but do admit that TrustBackedAssets isnt the friendliest name and its use from an end user pespective might be based for UX. Its possibly a tradeoff between signalling versus simple naming.

Daanvdplas:
We have decided to go for `Assets` naming in stead of `TrustBackedAssets`

PR #82, opened 2024-04-24
test: create an NFT collection

brunopgalvao:
ditto

### DIFF
--- a/runtime/devnet/src/extensions.rs
+++ b/runtime/devnet/src/extensions.rs
@@ -530,6 +530,9 @@ mod tests {
 				log::debug!("result: {:?}", result);
 			}
 
+			// check that the nft collection was created
+			assert_eq!(Nfts::collection_owner(0), Some(addr.into()));
+
 			// // check for revert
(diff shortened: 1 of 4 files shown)


runtime/devnet/src/extensions.rs
```diff
@@ -474,6 +474,67 @@ mod tests {
...
+				log::debug!(
+					"Contract debug buffer - {:?}",
+					String::from_utf8(result.debug_message.clone())
+				);
+				log::debug!("result: {:?}", result);
+			}
+
+			// // check for revert
+			assert!(!result.result.unwrap().did_revert(), "Contract reverted!");
+		});
```
al3mart, line 535: Ideally we would want to check that the collection has been created

brunopgalvao: Yes ser 🫡

al3mart (approved):

Daanvdplas (approved):

al3mart (approved):

PR #83, opened 2024-04-25
chore: update codegen

brunopgalvao:
If you get the following when trying to run the `fund-dev-accounts` script:
```
cargo run
   Compiling fund-dev-accounts v0.0.0 (/Users/bruno/src/pop-node/scripts/fund-dev-accounts)
    Finished dev [unoptimized + debuginfo] target(s) in 7.21s
     Running `/Users/bruno/src/pop-node/target/debug/fund-dev-accounts`
Error: Metadata(IncompatibleCodegen)
```

You can update the codegen for each respective chain by specifying the url:
```
subxt codegen --url ws://127.0.0.1:58043 | rustfmt > rococo_interface.rs
subxt codegen --url ws://127.0.0.1:58051 | rustfmt > pop_interface.rs
```

> Source: https://docs.rs/subxt/latest/subxt/book/setup/codegen/index.html#the-cli-tool


Daanvdplas (approved):

Daanvdplas (changes requested):
Still get the same error when running zombienet with relay v1.9.0

brunopgalvao:
> Still get the same error when running zombienet with relay v1.9.0

Most likely you will need to re-run these commands and re-create the metadata:
```
subxt codegen --url ws://127.0.0.1:58043 | rustfmt > rococo_interface.rs
subxt codegen --url ws://127.0.0.1:58051 | rustfmt > pop_interface.rs
```
Then run again.

Anytime the runtime changes, which for us will be quite often because we are building Pop Network, then we will have to recreate the metadata with the above command.

Daanvdplas:
> > Still get the same error when running zombienet with relay v1.9.0

Also fails with relay v1.8.0 for me

brunopgalvao:
> Also fails with relay v1.8.0 for me

Are you getting the error: `Error: Metadata(IncompatibleCodegen)`? If so you will need to re-run those commands properly.

Let's hop on a call when you are back and we can look into it.

Cargo.lock
```diff
@@ -4846,14 +4846,14 @@ dependencies = [
 
 [[package]]
 name = "jsonrpsee"
-version = "0.21.0"
+version = "0.22.4"
```
brunopgalvao, line 4849: `jsonrpsee` is getting bumped here. I would not want the dependencies in this script to conflict with the dependencies of the pop-node.

One thought is to create a `Cargo.toml` within this project directory.

Daanvdplas:
Closed by #87

Daanvdplas (approved):

evilrobot-01 (commented):
I don't know that we need to bump subxt here. I just updated subxt-cli locally and used it to generate the updated interface, which is probably all it needs.

Suggest reverting the subxt bump in cargo toml and lock and then retesting to see if all still works.

Daanvdplas:
Will test now

Daanvdplas:
Getting this error when running the script, 

Screenshot 2024-05-09 at 09 57 39

brunopgalvao:
I get the same error as Daan if I revert the subxt and subxt-signer to 0.34.0 on this branch and `cargo build`.

brunopgalvao:
Two ways that I see this being solved:

1. Make this script into its own crate with its own `Cargo.toml` file
2. Create a namespace for `subxt` and `subxt-signer` for this script

brunopgalvao:
Closing this. Re-open when we have a proper solution in-place for this.

PR #89, opened 2024-06-21
refactor: remove rococo

evilrobot-01:
Removes Rococo-specific configuration in favour of Paseo.

Also uses paseo zombienet config to fund dev/sovereign accounts. 

Potential todos:
- [x] remove `networks/rococo.toml`
- remove `scripts/fund-dev-accounts` > https://github.com/r0gue-io/pop-node/issues/137
- update integration tests to use paseo > https://github.com/r0gue-io/pop-node/issues/138


node/src/service.rs
```diff
@@ -482,7 +482,7 @@ where
 		relay_chain_slot_duration,
 		proposer,
 		collator_service,
-		authoring_duration: Duration::from_millis(1500),
+		authoring_duration: Duration::from_millis(2000),
```
peterwht, line 485: I'm not sold on this change still. https://wiki.polkadot.network/docs/maintain-guides-async-backing#phase-2---update-parachain-nodes Says 1,500, with the possible for two seconds, but it's more risky.

Note, [polkadot-parachain](https://github.com/paritytech/polkadot-sdk/blob/c4b3c1c6c6e492c4196e06fbba824a58e8119a3b/cumulus/polkadot-parachain/src/service.rs#L741) and [hyberbridge](https://github.com/polytope-labs/hyperbridge/blob/d4fffd7e6672ca3bd1a927c9fc677e14c5066186/parachain/node/src/service.rs#L476) both use 1,500

runtime/common/src/lib.rs
```diff
@@ -16,10 +16,7 @@ pub type Nonce = u32;
 /// up by `pallet_aura` to implement `fn slot_duration()`.
 ///
 /// Change this to adjust the block time.
-#[cfg(not(feature = "paseo"))]
 pub const MILLISECS_PER_BLOCK: u64 = 6000;
-#[cfg(feature = "paseo")]
-pub const MILLISECS_PER_BLOCK: u64 = 12000;
```
peterwht, line 22: Should we be removing 12 second block times already? I would like to see a structured strategy for the next Pop Net release for enabling async backing that we could thoroughly test.

peterwht: Note: I would love to see async backing enabled on on our testnet :)

runtime/common/src/lib.rs
```diff
@@ -70,10 +61,7 @@ pub const EXISTENTIAL_DEPOSIT: Balance = MILLIUNIT;
 // Async backing
 /// Maximum number of blocks simultaneously accepted by the Runtime, not yet included
 /// into the relay chain.
-#[cfg(not(feature = "paseo"))]
 pub const UNINCLUDED_SEGMENT_CAPACITY: u32 = 3;
```
peterwht, line 64: Feels a little weird introducing async backing to Paseo in a PR for removing Rococo.

node/src/service.rs
evilrobot-01: > Note: Set authoring_duration to whatever you want, taking your own hardware into account. But if the backer who should be slower than you due to reading from disk, times out at two seconds your candidates will be rejected.

https://wiki.polkadot.network/docs/maintain-guides-async-backing#phase-1---update-parachain-runtime
https://github.com/paritytech/polkadot-sdk/blob/c4b3c1c6c6e492c4196e06fbba824a58e8119a3b/templates/parachain/node/src/service.rs#L215
https://github.com/paritytech/polkadot-sdk/pull/4295 approved by Kian/Basti etc

runtime/common/src/lib.rs
evilrobot-01: reverted enablement

runtime/common/src/lib.rs
evilrobot-01: reverted

node/src/service.rs
evilrobot-01: reverted

Daanvdplas (approved):

peterwht (approved):

Cryptopod13 (approved):

PR #112, opened 2024-07-19
refactor: remove old api code

Daanvdplas:
Only removes old code in the runtime so that it can no longer be used on testnet until API complete. Removing the code in other crates will be done on the `api` branch as soon as it is rebased on this PR because it is very different and feels like I would have to do the job twice.

### DIFF
--- a/runtime/devnet/src/extensions.rs
+++ b/runtime/devnet/src/extensions.rs
@@ -1,30 +1,17 @@
-use cumulus_pallet_parachain_system::RelaychainDataProvider;
 use frame_support::traits::{Contains, OriginTrait};
 use frame_support::{
 	dispatch::{GetDispatchInfo, RawOrigin},
 	pallet_prelude::*,
-	traits::nonfungibles_v2::Inspect,
 };
 use pallet_contracts::chain_extension::{
 	BufInBufOutState, ChainExtension, ChargedAmount, Environment, Ext, InitState, RetVal,
 };
-use pop_primitives::{
-	cross_chain::CrossChainMessage,
-	storage_keys::{NftsKeys, ParachainSystemKeys, RuntimeStateKeys},
-	CollectionId, ItemId,
-};
+use pop_primitives::storage_keys::RuntimeStateKeys;
 use sp_core::crypto::UncheckedFrom;
-use sp_runtime::{
-	traits::{BlockNumberProvider, Dispatchable},
-	DispatchError,
-};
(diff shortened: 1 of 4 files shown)


runtime/devnet/src/extensions.rs
```diff
@@ -49,7 +33,6 @@ where
 	fn call<E: Ext>(&mut self, env: Environment<E, InitState>) -> Result<RetVal, DispatchError>
```
peterwht, line 33: clippy warning

runtime/devnet/src/extensions.rs
```diff
@@ -213,11 +184,8 @@ where
 	let key: RuntimeStateKeys = env.read_as()?;
 
 	let result = match key {
-		RuntimeStateKeys::Nfts(key) => read_nfts_state::<T, E>(key, &mut env),
-		RuntimeStateKeys::ParachainSystem(key) => {
-			read_parachain_system_state::<T, E>(key, &mut env)
-		},
-	}?
+		_ => Vec::<u8>::default(),
```
peterwht, line 187: clippy warning

runtime/testnet/src/extensions.rs
```diff
@@ -44,7 +33,6 @@ where
 	fn call<E: Ext>(&mut self, env: Environment<E, InitState>) -> Result<RetVal, DispatchError>
```
peterwht, line 33: clippy warning

runtime/testnet/src/extensions.rs
```diff
@@ -202,11 +184,8 @@ where
 	let key: RuntimeStateKeys = env.read_as()?;
 
 	let result = match key {
-		RuntimeStateKeys::Nfts(key) => read_nfts_state::<T, E>(key, &mut env),
-		RuntimeStateKeys::ParachainSystem(key) => {
-			read_parachain_system_state::<T, E>(key, &mut env)
-		},
-	}?
+		_ => Vec::<u8>::default(),
+	}
```
peterwht, line 188: clippy warning

peterwht (commented):
looks good. Just a couple clippy warnings that should be addressed

Daanvdplas:
As discussed elsewhere, we are ignoring those clippy warnings for now because they will be introduced back in with the pop api PR.

evilrobot-01 (approved):

peterwht (approved):

Issue #104, opened 2024-07-05
feat (pop api): `transfer_from` and `decrease_allowance`


Daanvdplas:
Implemented in:
- fungibles pallet, including tests and if required benchmark function
- integration tests

Issue #101, opened 2024-07-05
refactor: exclude encoding of reading state results in chain extension


Daanvdplas:
Because we funnelled the read state functions through `read_state` on the contract size the result had to be raw bytes, therefore encoded on the runtime and decoded in the contract. The read state functions are now called specifically and therefore the result can be its type (e.g. `Balance`). By refactoring it prevents the pop api having to decode the read state results which can decrease the contract size.

With the current contract it decreases it from 7.1K to 6.8K (don't compare with the psp22 contract as the logic implemented is not equal).

In stead of:
```
fn read_fungibles_state<T, E>(
	key: Keys<T>,
	env: &mut Environment<E, BufInBufOutState>,
) -> Result<Vec<u8>, DispatchError>
where
	T: pallet_contracts::Config
		+ pallet_assets::Config<TrustBackedAssetsInstance, AssetId = AssetId>
		+ fungibles::Config,
	E: Ext<T = T>,
	T: frame_system::Config<AccountId = sp_runtime::AccountId32>,
{
	env.charge_weight(T::DbWeight::get().reads(1_u64))?;
	match key {
		TotalSupply(id) => Ok(fungibles::Pallet::<T>::total_supply(id).encode()),
	}
	...
}
```
We do:
```
fn read_fungibles_state<T, E>(
	key: Keys<T>,
	env: &mut Environment<E, BufInBufOutState>,
) -> Result<(), DispatchError>
where
	T: pallet_contracts::Config
		+ pallet_assets::Config<TrustBackedAssetsInstance, AssetId = AssetId>
		+ fungibles::Config,
	E: Ext<T = T>,
	T: frame_system::Config<AccountId = sp_runtime::AccountId32>,
{
	env.charge_weight(T::DbWeight::get().reads(1_u64))?;
	match key {
		TotalSupply(id) => env.write(&fungibles::Pallet::<T>::total_supply(id), false, None),
	}
	...
	Ok(())
}
```

Daanvdplas:
Closed by #122

Issue #109, opened 2024-07-18
CI: pop api + integration tests


Daanvdplas:
We want CI to run the tests + integration tests in the pop api crate. For inspiration for the integration tests: https://github.com/use-ink/ink/tree/master/integration-tests. The ink repository has something similar.

Issue #93, opened 2024-07-05
refactor: `Error` from primitives to contract library


Daanvdplas:
The Error type is only used in runtime via tests, and the fact that a primitives crate is referring to pop api on line 16 makes it seem off.

I think the tests that use this error should probably live in pop-api/integration-tests, especially as it already takes a dev-dependency on the runtime. Then this error should only need to be defined in the pop-api crate, where it probably belongs as it is a contract-side only representation of the errors which exist in the runtime.

If a type isnt being used by both a contract AND the runtime, it probably doesnt need to be in this crate.

Daanvdplas:
Closing in favour of #91

PR #113, opened 2024-07-19
feat: api fungibles pallet

Daanvdplas:
Adds the api pallet implementing the psp22 interface. It resulted into an overall refactor of the pop api, mainly simplifying the implementation. Some small other relevant changes:
- Removal of old api code (sub0)
- Refactor of `assets` to `fungibles` feature in pop-api.
- Created a folder `contracts` in `pop-api/integration-tests` which is contains the contracts for the integration tests.
- Removed `transfer_from` tests since it was a copy from `transfer. This is still to be implemented.

Remaining todos:
- [ ] #104 
- [x] benchmark function for `approve`
- [x] #101 - fixed in #122


pallets/fungibles/src/lib.rs
```diff
@@ -0,0 +1,64 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
evilrobot-01, line 1: Can we start with a single crate named `pallets/api` and work from there? I am concerned about the additional boilerplate that might be required with loads of pallets with a few functions. Finding the balance will come with time and we can decide before progressing beyond devnet.

Also, https://github.com/paritytech/polkadot-sdk/tree/master/polkadot/runtime/parachains/src gives an example how pallets can be defined within a single crate, per module. I think we should ultimately follow this type of approach.

pallets/fungibles/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+	impl<T: Config> Pallet<T> {
+		#[pallet::call_index(0)]
+		#[pallet::weight(<T as pallet_assets::Config<TrustBackedAssetsInstance>>::WeightInfo::transfer_keep_alive())]
+		pub fn transfer(
+			origin: OriginFor<T>,
+			id: AssetIdParameterOf<T>,
+			target: AccountIdLookupOf<T>,
+			amount: BalanceOf<T>,
+		) -> DispatchResult {
+			pallet_assets::Pallet::<T, TrustBackedAssetsInstance>::transfer_keep_alive(
```
evilrobot-01, line 53: Can we show the multi-call wrapping instead so this has more value as a prototype? Same adding the corresponding weight.

runtime/devnet/src/lib.rs
```diff
@@ -663,6 +663,7 @@ construct_runtime!(
 		Nfts: pallet_nfts = 50,
 		NftFractionalization: pallet_nft_fractionalization = 51,
 		Assets: pallet_assets::<Instance1> = 52,
+		Fungibles: pallet_fungibles = 53,
```
evilrobot-01, line 666: Suggest an API section, leaving a decent gap for future SDK pallets.

pallets/fungibles/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+				origin, id, target, amount,
+			)
+		}
+	}
+
+	impl<T: Config> Pallet<T> {
+		pub fn total_supply(id: AssetIdOf<T>) -> BalanceOf<T> {
```
evilrobot-01, line 60: Can we show usage of this when ready.

runtime/devnet/src/extensions/v0/assets.rs
```diff
@@ -49,27 +49,33 @@ pub(crate) fn construct_assets_key(
 pub(crate) fn construct_assets_call(
 	call_index: u8,
 	params: Vec<u8>,
-) -> Result<pallet_assets::Call<Runtime, TrustBackedAssetsInstance>, DispatchError> {
+	// ) -> Result<pallet_assets::Call<Runtime, TrustBackedAssetsInstance>, DispatchError> {
+) -> Result<pallet_fungibles::Call<Runtime>, DispatchError> {
 	match call_index {
 		9 => {
```
evilrobot-01, line 55: Pallet id doesnt match but yeah

evilrobot-01 (commented):
Left some suggestions on initial structuring to make it easier to reason and start out closer to how we may end up.

pallets/fungibles/src/lib.rs
Daanvdplas: Could you elaborate please?

runtime/devnet/src/extensions/v0/assets.rs
Daanvdplas: Yes this was simply to see if the tests were passing

pallets/fungibles/src/lib.rs
Daanvdplas: See latest commit

runtime/devnet/src/lib.rs
Daanvdplas: Took 150 with 255 as max. Leaving space for 100 pallets and 100 api pallets

pallets/fungibles/src/lib.rs
Daanvdplas: See latest commit. 

What do we prefer, like it is now or have a `fungibles` folder with `mod.rs` (like in the pop api crate)?

pallets/api/Cargo.toml
```diff
@@ -0,0 +1,50 @@
+[package]
+name = "pallet-pop-api"
```
evilrobot-01, line 2:
```suggestion
name = "pallet-api"
```

Pop prefix adds no value, its obvious based on the repo.

pallets/api/Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+authors.workspace = true
+description = "Pop api pallet, enabling smart contracts to use the power of Polkadot"
```
evilrobot-01, line 4:
```suggestion
description = "API pallet, enabling smart(er) contracts with the power of Polkadot."
```

pallets/api/src/fungibles/benchmarking.rs
```diff
@@ -0,0 +1,22 @@
+//! Benchmarking setup for pallet-parachain-template
```
evilrobot-01, line 1: No need if we are just wrapping to start, suggest removing until it is actually required.

pallets/api/src/mock.rs
```diff
@@ -0,0 +1,60 @@
+use frame_support::{parameter_types, traits::Everything};
```
evilrobot-01, line 1: Like that you kept this in the crate root.

pallets/api/src/fungibles.rs
```diff
@@ -0,0 +1,84 @@
...
+			origin: OriginFor<T>,
+			id: AssetIdParameterOf<T>,
+			target: AccountIdLookupOf<T>,
+			amount: BalanceOf<T>,
+		) -> DispatchResult {
+			Assets::<T>::transfer_keep_alive(origin, id, target, amount)
+		}
+
+		#[pallet::call_index(1)]
+		#[pallet::weight(AssetsWeightInfo::<T>::cancel_approval())]
```
evilrobot-01, line 60:
```suggestion
		#[pallet::weight(AssetsWeightInfo::<T>::cancel_approval() + AssetsWeightInfo::<T>::approve_transfer())]
```

pallets/api/src/fungibles.rs
```diff
@@ -0,0 +1,84 @@
...
+		<T as frame_system::Config>::AccountId,
+	>>::AssetId;
+	type AssetIdParameterOf<T> =
+		<T as pallet_assets::Config<TrustBackedAssetsInstance>>::AssetIdParameter;
+	type Assets<T> = pallet_assets::Pallet<T, TrustBackedAssetsInstance>;
+	type AssetsWeightInfo<T> = <T as pallet_assets::Config<TrustBackedAssetsInstance>>::WeightInfo;
+	type BalanceOf<T> = <pallet_assets::Pallet<T, TrustBackedAssetsInstance> as Inspect<
+		<T as frame_system::Config>::AccountId,
+	>>::Balance;
+	// Should be defined in primitives.
```
evilrobot-01, line 34: Should be configurable via the pallet config trait rather

runtime/devnet/src/config/assets.rs
```diff
@@ -12,6 +8,11 @@ use pallet_nfts::PalletFeatures;
 use parachains_common::{AssetIdForTrustBackedAssets, CollectionId, ItemId, Signature};
 use sp_runtime::traits::Verify;
 
+use crate::{
```
evilrobot-01, line 11: Why are we leaving spaces between imports? It infers some cryptic separation that some maintainer needs to know about and adds no value. Far easier to group them into a single block and let cargo fmt handle consistency.

pallets/fungibles/src/lib.rs
evilrobot-01: Assuming we will have a separate tests.rs submodule, probably fungibles/mod.rs. My subjective preference is to only break out into a directory with multiple module files when it warrants it.

pallets/fungibles/src/lib.rs
evilrobot-01: Usage from the chain extension. At the moment its just an unused function, would be good to show where it plugs in for the prototype.

runtime/devnet/src/config/assets.rs
Daanvdplas: External and local crates, rust best practice. This is how it is done in the pop cli code (maybe not the code you wrote). I'm fine with both but I would like to make a decision here so that it doesn't effect reviewing and writing code in the future of our repos.

pallets/api/src/fungibles.rs
Daanvdplas: Thank you

pallets/api/src/fungibles.rs
Daanvdplas: Great suggestion!

runtime/devnet/src/config/assets.rs
evilrobot-01: Source?

Id prefer to not have to worry about this, but https://rust-lang.github.io/rustfmt/?version=master&search=import#group_imports isn't stable. Formatting should be consistent but automated imo.

runtime/devnet/rustc-ice-2024-07-20T20_16_19-33192.txt
```diff
@@ -0,0 +1,67 @@
+thread 'rustc' panicked at compiler/rustc_middle/src/util/bug.rs:35:44:
```
evilrobot-01, line 1: Any reason this is included?

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+			id: AssetIdOf<T>,
+			spender: AccountIdOf<T>,
+			value: BalanceOf<T>,
+		) -> DispatchResultWithPostInfo {
+			let spender = T::Lookup::unlookup(spender);
+			let id: AssetIdParameterOf<T> = id.into();
+			Assets::<T>::cancel_approval(origin.clone(), id.clone(), spender.clone())
+				.map_err(|e| e.with_weight(AssetsWeightInfo::<T>::cancel_approval()))?;
+			Assets::<T>::approve_transfer(origin, id, spender, value).map_err(|e| {
+				e.with_weight(
```
evilrobot-01, line 84: Seems unnecessary when the predispatch weight charged is exactly this

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+					AssetsWeightInfo::<T>::cancel_approval()
+						+ AssetsWeightInfo::<T>::approve_transfer(),
+				)
+			})?;
+			Ok(().into())
+		}
+	}
+
+	impl<T: Config> Pallet<T> {
+		pub fn total_supply(id: AssetIdOf<T>) -> BalanceOf<T> {
```
evilrobot-01, line 94: As you have now defined the enum in the pallet, why not just have a single `read(key)` function

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,122 @@
...
+	type AssetsWeightInfo<T> = <T as pallet_assets::Config<AssetsInstanceOf<T>>>::WeightInfo;
+	type BalanceOf<T> = <pallet_assets::Pallet<T, AssetsInstanceOf<T>> as Inspect<
+		<T as frame_system::Config>::AccountId,
+	>>::Balance;
+
+	/// The required input for state queries in pallet assets.
+	#[derive(Encode, Decode, Debug, MaxEncodedLen)]
+	#[repr(u8)]
+	#[allow(clippy::unnecessary_cast)]
+	pub enum Keys<T: Config> {
```
evilrobot-01, line 40: Not sure Keys is the correct name, as those values arent explicitly keys. Also enum naming convention is typically singular, as one typically only specifies one value at a type. Different if bitflags mind.

runtime/devnet/rustc-ice-2024-07-20T20_16_19-33192.txt
Daanvdplas: Thanks

pallets/api/src/fungibles/mod.rs
Daanvdplas: Thanks

primitives/src/lib.rs
```diff
@@ -6,30 +6,27 @@ use scale::{Decode, Encode, MaxEncodedLen};
...
-	/// The maximum length of an attribute key.
-	pub type KeyLimit = ConstU32<64>;
-	/// The maximum approvals an item could have.
-	pub type ApprovalsLimit = ConstU32<20>;
+#[cfg(feature = "fungibles")]
+pub mod constants {
+	/// Module index.
+	pub const FUNGIBLES: u8 = 150;
+	pub mod fungibles {
+		pub const TOTAL_SUPPLY: u8 = 0;
```
evilrobot-01, line 23: Is this not overly complex? Does using an index attribute on the enum in the runtime and hardcoding the u8 value in the pop-api function not achieve the same thing without the additional 'coupling'.

Seems additional maintenance and tons of consts just to guarantee that the same index is used, when this would probably be covered by some test to ensure that functionality works as expected.

I figure its far simpler to define the index on the enum in the runtime and then just specify the same value in the relevant function within pop_api crate.

runtime/devnet/src/extensions/v0/mod.rs
```diff
@@ -1,2 +1 @@
-pub(crate) mod assets;
 pub(crate) mod error;
```
evilrobot-01, line 1: v0/mod.rs is effectively empty, with error.rs as a submodule. It would be far simpler to just have v0.rs with the functionality (a few functions and consts) from error.rs included directly within v0.rs.

pallets/api/src/fungibles/mod.rs
Daanvdplas: Right now there is still the option to implement #101 (updated)

primitives/src/lib.rs
Daanvdplas: It depends what we want. If we think this overcomplicates things too much we can definitely do what we had before. I thought the 'coupling' is something that we wanted and therefore implemented this to show.

runtime/devnet/src/extensions/mod.rs
```diff
@@ -373,118 +290,27 @@ impl TryFrom<u8> for FuncId {
...
-		NftsKeys::CollectionAttribute(collection, key) => {
-			env.charge_weight(T::DbWeight::get().reads(1_u64))?;
-			Ok(pallet_nfts::Pallet::<T>::collection_attribute(&collection, &key).encode())
-		},
-	}
-}
-
-fn read_assets_state<T, E>(
-	key: AssetsKeys,
+fn read_fungibles_state<T, E>(
```
evilrobot-01, line 293: Replacing all the individual read functions with pallet::read(key) could eliminate the need for this function.

runtime/devnet/src/extensions/mod.rs
```diff
@@ -247,48 +196,18 @@ where
...
-	let origin: RuntimeOrigin = RawOrigin::Root.into();
-	// Generate runtime call to dispatch.
-	let call = RuntimeCall::PolkadotXcm(pallet_xcm::Call::send {
-		dest: Box::new(dest),
-		message: Box::new(VersionedXcm::V4(message)),
-	});
-	dispatch_call::<T, E>(env, call, origin, LOG_PREFIX)
+// Example wrapper to enable versioning of state read keys
+#[derive(Encode, Decode, Debug, MaxEncodedLen)]
+enum VersionedRuntimeStateKeys<T: fungibles::Config> {
```
evilrobot-01, line 201: Maybe align these enum names with the extension function names - e.g. VersionedStateRate and VersionedDispatch

pallets/api/src/fungibles/mod.rs
Daanvdplas: Changed it to `FungibleKey`

runtime/devnet/src/extensions/v0/mod.rs
Daanvdplas: Refactored

runtime/devnet/src/extensions/mod.rs
Daanvdplas: Refactored to VersionedStateRead and VersionedDispatch

pallets/api/src/fungibles/mod.rs
evilrobot-01: Technically speaking one of these implementations could look up multiple values, combining them into or calculating a result. Something more abstract like 'State' or 'Query' or 'StateQuery' might be a better starting name, mostly as people just tend to copy some existing pattern.

This may suggest an improvement to the corresponding extension name too.

runtime/devnet/src/extensions/mod.rs
Daanvdplas: I will show #101 prototype with a PR to this branch and see if we want that. If so, this will not be possible because we need to return the result directly with its type in stead of encoding it.

primitives/src/lib.rs
Daanvdplas: Removed the constants from `primitives`.  Curious to hear whether we should use them in pop api. Previously this was desired as it improves readability but does add a lot of constants.

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+
+		/// Returns the token name for a given asset ID.
+		///
+		/// # Arguments
+		/// * `id` - The ID of the asset.
+		///
+		/// # Returns
+		/// The name of the token as a byte vector, or an error if the operation fails.
+		pub fn token_name(id: AssetIdOf<T>) -> Vec<u8> {
+			<Assets<T> as MetadataInspect<AccountIdOf<T>>>::name(id)
```
peterwht, line 212: Type alias for: `<Assets<T> as MetadataInspect<AccountIdOf<T>>>`?

pop-api/integration-tests/src/local_fungibles.rs
```diff
@@ -263,15 +260,10 @@ fn token_decimals_asset(asset_id: AssetId) -> u8 {
...
-#[ignore]
 fn total_supply_works() {
 	new_test_ext().execute_with(|| {
 		let _ = env_logger::try_init();
-		let addr = instantiate(
-			"../../pop-api/examples/fungibles/target/ink/fungibles.wasm",
-			INIT_VALUE,
-			vec![],
-		);
+		let addr = instantiate("contracts/fungibles/target/ink/fungibles.wasm", INIT_VALUE, vec![]);
```
peterwht, line 266: IMO, these .wasm files should be moved to an `artifacts` or `fixtures` directory. Otherwise these tests may fail if someone hasn't built all of the contracts.

peterwht (commented):
Looking good so far. Exciting to see how this is shaping up. Left some comments.

runtime/devnet/src/extensions/mod.rs
```diff
@@ -247,54 +187,24 @@ where
...
-	};
-	// TODO: revisit to replace with signed contract origin
-	let origin: RuntimeOrigin = RawOrigin::Root.into();
-	// Generate runtime call to dispatch.
-	let call = RuntimeCall::PolkadotXcm(pallet_xcm::Call::send {
-		dest: Box::new(dest),
-		message: Box::new(VersionedXcm::V4(message)),
-	});
-	dispatch_call::<T, E>(env, call, origin, LOG_PREFIX)
+// Example wrapper to enable versioning of `RuntimeStateKeys`.
```
evilrobot-01, line 190: Remove `example` in comment, convert to doc comments and add doc comments for variants if this is being retained.

runtime/devnet/src/extensions/mod.rs
```diff
@@ -247,54 +187,24 @@ where
...
+#[derive(Decode, Debug)]
+enum VersionedStateRead<T: fungibles::Config> {
+	#[codec(index = 0)]
+	V0(RuntimeStateKeys<T>),
+}
+
+// Wrapper to enable versioning of `RuntimeCall`.
```
evilrobot-01, line 197: Convert to doc comments and add doc comments for variants if this is being retained.

runtime/devnet/src/extensions/mod.rs
```diff
@@ -373,118 +281,27 @@ impl TryFrom<u8> for FuncId {
...
 ) -> Result<Vec<u8>, DispatchError>
 where
 	T: pallet_contracts::Config
-		+ pallet_assets::Config<TrustBackedAssetsInstance, AssetId = AssetId>,
+		+ pallet_assets::Config<TrustBackedAssetsInstance, AssetId = AssetId>
+		+ fungibles::Config,
 	E: Ext<T = T>,
 	T: frame_system::Config<AccountId = sp_runtime::AccountId32>,
 {
+	env.charge_weight(T::DbWeight::get().reads(1_u64))?;
```
evilrobot-01, line 295: As mentioned elsewhere, a reminder that any state read may in theory perform multiple reads to get the data to return, say by combining. Minor risk that it gets forgotten by being a catchall at the top.

runtime/devnet/src/lib.rs
```diff
@@ -252,75 +254,15 @@ impl Contains<RuntimeCall> for FilteredCalls {
 
 /// A type to identify allowed calls to the Runtime from contracts. Used by Pop API
 pub struct AllowedApiCalls;
```
evilrobot-01, line 256: Could be moved to suggested `/configs/api.rs` module. Or we just have an `api.rs` module for this, the state read enum and the pallet configs so it is treated as one api from the runtime maintenance perspective.

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+		/// # Arguments
+		/// * `id` - The ID of the asset.
+		/// * `to` - The recipient account.
+		/// * `value` - The number of tokens to transfer.
+		///
+		/// # Returns
+		/// Returns `Ok(())` if successful, or an error if the transfer fails.
+		#[pallet::call_index(0)]
+		#[pallet::weight(AssetsWeightInfo::<T>::transfer_keep_alive())]
+		pub fn transfer(
```
evilrobot-01, line 79: Important question: what happens when v1 needed. Is the plan to create a whole new pallet, taking up an index in the runtime, or to just add new functions here?

If the latter, should these functions be prefixed with v0_ to indicate intentions? Whilst renaming wont affect encoding based on call index, it may break integrations using the name obtained via metadata.

pallets/api/src/fungibles/benchmarking.rs
```diff
@@ -0,0 +1,60 @@
+//! Benchmarking setup for pallet-cards
```
evilrobot-01, line 1: Not reviewed

pallets/api/src/fungibles/tests.rs
```diff
@@ -0,0 +1,144 @@
+use crate::mock::*;
```
evilrobot-01, line 1: Not reviewed

evilrobot-01 (commented):
Shaping up, still a decent amount of comments sadly. 

Also annoying to have to re-review changes which have already been merged to main, making this take much longer than expected.

pop-api/integration-tests/src/local_fungibles.rs
Daanvdplas: Included in #109

pallets/api/src/mock.rs
```diff
@@ -0,0 +1,122 @@
...
+	}
+);
+
+parameter_types! {
+	pub const BlockHashCount: u64 = 250;
+	pub const SS58Prefix: u8 = 42;
+}
+
+#[derive_impl(frame_system::config_preludes::TestDefaultConfig as frame_system::DefaultConfig)]
+impl frame_system::Config for Test {
```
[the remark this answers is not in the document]
Daanvdplas: The reason for using those old configurations was because of the old polkadot version we are currently add. I wasn't sure those were already available so I took reference of another project we are building when it was on polkadot 1.7. I suggest we refactor those when doing the upgrades.

pallets/api/src/mock.rs
```diff
@@ -0,0 +1,122 @@
...
+	type AccountData = pallet_balances::AccountData<u128>;
+	type OnNewAccount = ();
+	type OnKilledAccount = ();
+	type SystemWeightInfo = ();
+	type SS58Prefix = SS58Prefix;
+	type OnSetCode = ();
+	type MaxConsumers = ConstU32<16>;
+}
+
+impl pallet_balances::Config for Test {
```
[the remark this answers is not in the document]
Daanvdplas: See above comment

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+		///
+		/// # Arguments
+		/// * `id` - The ID of the asset.
+		/// * `spender` - The account that is allowed to spend the tokens.
+		/// * `value` - The number of tokens to approve.
+		///
+		/// # Returns
+		/// Returns `Ok(())` if successful, or an error if the approval fails.
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::DbWeight::get().reads(1) + AssetsWeightInfo::<T>::cancel_approval() + AssetsWeightInfo::<T>::approve_transfer())]
```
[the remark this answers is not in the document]
Daanvdplas: Correct

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+					|e| {
+						e.with_weight(
+							T::DbWeight::get().reads(1) + AssetsWeightInfo::<T>::cancel_approval(),
+						)
+					},
+				)?;
+				Assets::<T>::approve_transfer(origin, id, spender, value)?;
+			}
+
+			Ok(().into())
```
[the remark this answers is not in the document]
Daanvdplas: Question is whether those should be returned by the contract or the pallet?

Will create an issue for it when we decided on this.

pallets/api/src/fungibles/mod.rs
Daanvdplas: This is hard to predict but this theoretically doesn't have to change, ever. As long as the standards are being used the api in the contract should never change. If parameter types change slightly, these should be handled in the pallet via a runtime upgrade but this won't require a different version because the type in the contract stays the same. 

When a parameter gets added to pallet assets' `transfer` there is a list of things that will happen:
- For old contracts nothing will change, the pallet fungibles' `transfer` will add some logic via a runtime upgrade. 
> should these functions be prefixed with v0_ to indicate intentions? Whilst renaming wont affect encoding based on call index, it may break integrations using the name obtained via metadata.

+ Renaming makes sure integrations outside of contracts won't break.
- Renaming makes the code uglier and perhaps we will never need it. If we need it in the end we can change the naming and integrations outside of contracts will break.

- New contracts can use a `v1::fungibles::transfer` where they provide the extra parameter. This function will have a different dispatchable index and will call the `v1_transfer()`

runtime/devnet/src/extensions/mod.rs
Daanvdplas: Will be added to #122

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+		<T as frame_system::Config>::AccountId,
+	>>::AssetId;
+	type AssetIdParameterOf<T> =
+		<T as pallet_assets::Config<AssetsInstanceOf<T>>>::AssetIdParameter;
+	pub(crate) type Assets<T> = pallet_assets::Pallet<T, AssetsInstanceOf<T>>;
+	pub(crate) type AssetsInstanceOf<T> = <T as Config>::AssetsInstance;
+	type AssetsWeightInfo<T> = <T as pallet_assets::Config<AssetsInstanceOf<T>>>::WeightInfo;
+	pub(crate) type BalanceOf<T> = <pallet_assets::Pallet<T, AssetsInstanceOf<T>> as Inspect<
+		<T as frame_system::Config>::AccountId,
+	>>::Balance;
```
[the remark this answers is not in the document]
Daanvdplas: See `refactor: pallet`

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+			let spender = T::Lookup::unlookup(spender);
+			let id: AssetIdParameterOf<T> = id.into();
+			// If the new value is equal to the current allowance, do nothing.
+			if value == current_allowance {
+				return Ok(().into());
+			}
+			// If the new value is greater than the current allowance, approve the difference.
+			if value > current_allowance {
+				value.saturating_reduce(current_allowance);
+				Assets::<T>::approve_transfer(origin, id, spender, value).map_err(|e| {
```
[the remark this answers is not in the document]
Daanvdplas: see `refactor: pallet`

TMO this is not required and only makes the code uglier.

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+			spender: AccountIdOf<T>,
+			mut value: BalanceOf<T>,
+		) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin.clone())
+				// To have the caller pay some fees.
+				.map_err(|e| e.with_weight(T::DbWeight::get().reads(1)))?;
+			let current_allowance = Assets::<T>::allowance(id.clone(), &who, &spender);
+			let spender = T::Lookup::unlookup(spender);
+			let id: AssetIdParameterOf<T> = id.into();
+			// If the new value is equal to the current allowance, do nothing.
```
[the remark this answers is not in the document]
Daanvdplas: The comment serves as the whitespace here. TMO only for really big functions those whitespaces are needed. In another project we have never used whitespaces so before changing I would like to hear a clear guideline.

pallets/api/src/fungibles/mod.rs
Daanvdplas: Can you make a type alias of a trait?

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+		<T as frame_system::Config>::AccountId,
+	>>::Balance;
+
+	/// The required input for state queries through the fungibles api.
+	#[derive(Encode, Decode, Debug, MaxEncodedLen)]
+	#[repr(u8)]
+	#[allow(clippy::unnecessary_cast)]
+	pub enum FungiblesKey<T: Config> {
+		#[codec(index = 0)]
+		TotalSupply(AssetIdOf<T>),
```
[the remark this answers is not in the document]
Daanvdplas: see `refactor: pallet`

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+		#[codec(index = 3)]
+		TokenName(AssetIdOf<T>),
+		#[codec(index = 4)]
+		TokenSymbol(AssetIdOf<T>),
+		#[codec(index = 5)]
+		TokenDecimals(AssetIdOf<T>),
+	}
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config + pallet_assets::Config<Self::AssetsInstance> {
```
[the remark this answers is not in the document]
Daanvdplas: see refactor: pallet

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+	use sp_runtime::{traits::StaticLookup, Saturating};
+	use sp_std::vec::Vec;
+
+	pub(crate) type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
+	pub(crate) type AssetIdOf<T> = <pallet_assets::Pallet<T, AssetsInstanceOf<T>> as Inspect<
+		<T as frame_system::Config>::AccountId,
+	>>::AssetId;
+	type AssetIdParameterOf<T> =
+		<T as pallet_assets::Config<AssetsInstanceOf<T>>>::AssetIdParameter;
+	pub(crate) type Assets<T> = pallet_assets::Pallet<T, AssetsInstanceOf<T>>;
```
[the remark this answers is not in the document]
Daanvdplas: see refactor: pallet

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+	pub(crate) type AssetsInstanceOf<T> = <T as Config>::AssetsInstance;
+	type AssetsWeightInfo<T> = <T as pallet_assets::Config<AssetsInstanceOf<T>>>::WeightInfo;
```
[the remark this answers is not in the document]
Daanvdplas: see refactor: pallet

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+			let current_allowance = Assets::<T>::allowance(id.clone(), &who, &spender);
+			let spender = T::Lookup::unlookup(spender);
+			let id: AssetIdParameterOf<T> = id.into();
+			// If the new value is equal to the current allowance, do nothing.
+			if value == current_allowance {
+				return Ok(().into());
+			}
+			// If the new value is greater than the current allowance, approve the difference.
+			if value > current_allowance {
+				value.saturating_reduce(current_allowance);
```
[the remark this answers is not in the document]
Daanvdplas: see refactor: pallet

pop-api/integration-tests/src/local_fungibles.rs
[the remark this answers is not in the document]
Daanvdplas: Cannot call non-const on constants. This issue will change is anyway most probably: https://github.com/r0gue-io/pop-node/issues/109#issue-2415542297

pop-api/integration-tests/src/local_fungibles.rs
```diff
@@ -362,11 +348,11 @@ fn transfer_works() {
 			Module { index: 52, error: 0 },
 		);
 		// Successful transfer.
-		let bob_balance_before_mint = Assets::balance(asset, &BOB);
+		let bob_balance_before_transfer = Assets::balance(asset, &BOB);
```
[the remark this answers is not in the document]
Daanvdplas: see refactor: integration-tests

pop-api/src/lib.rs
```diff
@@ -22,7 +13,7 @@ pub type Result<T> = core::result::Result<T, StatusCode>;
 mod constants {
 	// Errors:
 	pub(crate) const DECODING_FAILED: u32 = 255;
-	pub(crate) const MODULE_ERROR: u8 = 3;
+	pub(crate) const _MODULE_ERROR: u8 = 3;
```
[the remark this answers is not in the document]
Daanvdplas: see refactor: ink api library

pop-api/src/v0/mod.rs
```diff
@@ -1,13 +1,6 @@
 use crate::{primitives::error::Error, StatusCode};
 
-#[cfg(feature = "assets")]
 pub mod assets;
```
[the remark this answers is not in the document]
Daanvdplas: see refactor: ink api library

primitives/src/lib.rs
```diff
@@ -1,37 +1,13 @@
...
-	pub type ItemId = u32;
-	/// The maximum length of an attribute key.
-	pub type KeyLimit = ConstU32<64>;
-	/// The maximum approvals an item could have.
-	pub type ApprovalsLimit = ConstU32<20>;
-}
-
 pub mod v0 {
 	use super::*;
 	pub mod error {
```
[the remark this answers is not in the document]
Daanvdplas: Created a separate issue for this #93

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+	impl<T: Config> Pallet<T> {
+		/// Transfers `value` amount of tokens from the caller's account to account `to`, with additional
+		/// `data` in unspecified format.
+		///
+		/// # Arguments
+		/// * `id` - The ID of the asset.
+		/// * `to` - The recipient account.
+		/// * `value` - The number of tokens to transfer.
+		///
+		/// # Returns
```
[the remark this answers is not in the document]
Daanvdplas: see refactor: pallet

pop-api/integration-tests/src/local_fungibles.rs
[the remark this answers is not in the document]
evilrobot-01: Don't follow.

I just mean `const CONTRACT : &str = "contracts/fungibles/target/ink/fungibles.wasm"`

runtime/devnet/Cargo.toml
```diff
@@ -119,6 +120,7 @@ std = [
     "pallet-balances/std",
     "pallet-collator-selection/std",
     "pallet-contracts/std",
+    "pallet-api/std",
```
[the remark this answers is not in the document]
Daanvdplas: see refactor: devnet runtime

runtime/devnet/src/extensions/mod.rs
Daanvdplas: see refactor: devnet runtime

runtime/devnet/src/extensions/mod.rs
Daanvdplas: see refactor: devnet runtime

pop-api/integration-tests/src/local_fungibles.rs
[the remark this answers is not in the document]
Daanvdplas: see refactor: integration-tests

runtime/devnet/src/config/assets.rs
```diff
@@ -119,3 +120,7 @@ impl pallet_assets::Config<TrustBackedAssetsInstance> for Runtime {
 	#[cfg(feature = "runtime-benchmarks")]
 	type BenchmarkHelper = ();
 }
+
+impl fungibles::Config for Runtime {
```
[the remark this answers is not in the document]
Daanvdplas: see refactor: devnet runtime

runtime/devnet/src/lib.rs
Daanvdplas: see refactor: devnet runtime

pallets/api/src/fungibles/mod.rs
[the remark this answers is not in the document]
Daanvdplas: I'm not following you here.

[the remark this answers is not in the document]
evilrobot-01: As its pre-dispatch, surely we need to calculate worst-case to ensure that the caller has sufficient funds before attempting the work?

primitives/src/lib.rs
evilrobot-01: No clear best approach sadly. In some circumstances the constants seem better, but its just a pointer to a special number used in a single place, then debatable on how much value it brings.

Perhaps we defer making this decision until everything comes together?

pallets/api/src/fungibles/mod.rs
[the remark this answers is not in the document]
evilrobot-01: Contract _may_ not have the relevant state to be able to emit such an event, requiring additional state reads just to be able to construct the event.

pallets/api/src/fungibles/mod.rs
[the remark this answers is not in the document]
Daanvdplas: see chore: benchmark fungibles approve

primitives/src/lib.rs
Daanvdplas: Sure thing. Created #124

pallets/api/src/fungibles/mod.rs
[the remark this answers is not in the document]
Daanvdplas: Very good point. Also has to be looked at, at a later stage I'm afraid - #125

pallets/api/src/fungibles/mod.rs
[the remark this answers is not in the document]
peterwht: Yes, the worst case for `cancel_approval` and `approve_transefer` is necessary.

However, only in one branch (the `else`) does `cancel_approval` get called. So, in the other branch the user is charged weight for `cancel_approval` despite it not being called.

The return `Ok(().into())` does not properly account for the actual weight used and overcharges.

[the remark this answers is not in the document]
Daanvdplas: You are right: https://github.com/r0gue-io/pop-node/pull/126

Need to finish tomorrow.

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,237 @@
...
+		/// Returns `Ok(())` if successful, or an error if the transfer fails.
+		#[pallet::call_index(0)]
+		#[pallet::weight(AssetsWeightInfo::<T>::transfer_keep_alive())]
+		pub fn transfer(
+			origin: OriginFor<T>,
+			id: AssetIdOf<T>,
+			target: AccountIdOf<T>,
+			amount: BalanceOf<T>,
+		) -> DispatchResult {
+			let target = T::Lookup::unlookup(target);
```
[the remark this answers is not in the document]
Daanvdplas: Will add to testnet if happy

pallets/api/src/fungibles/benchmarking.rs
```diff
@@ -17,9 +17,17 @@ use sp_runtime::traits::Zero;
...
+fn assert_has_event<T: Config>(
+	generic_event: <T as pallet_assets::Config<AssetsInstanceOf<T>>>::RuntimeEvent,
+) {
+	frame_system::Pallet::<T>::assert_has_event(generic_event.into());
+}
+
 #[benchmarks(
 	where
 	<pallet_assets::Pallet<T, AssetsInstanceOf<T>> as Inspect<<T as frame_system::Config>::AccountId>>::AssetId: Zero,
+	T: Config,
```
Daanvdplas, line 30: Will remove

Daanvdplas:
Remaining todos:
- refactor primitives crate name and ***
- fix benchmark approve #126

pallets/api/src/fungibles/mod.rs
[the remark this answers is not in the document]
Daanvdplas: See fix: approve benchmark + weights

pop-api/Cargo.toml
```diff
@@ -10,7 +10,7 @@ enumflags2 = { version = "0.7.7" }
 ink = { version = "5.0.0", default-features = false }
 sp-io = { version = "31.0.0", default-features = false, features = ["disable_panic_handler", "disable_oom", "disable_allocator"] }
 
-pop-primitives = { path = "../primitives", default-features = false }
+primitives = { path = "../primitives", default-features = false }
```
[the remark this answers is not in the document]
Daanvdplas: https://crates.io/crates/pop-primitives

pallets/api/src/fungibles/benchmarking.rs
Daanvdplas: Please review

pallets/api/src/fungibles/tests.rs
Daanvdplas: Please review :)

pallets/api/src/fungibles/benchmarking.rs
evilrobot-01: Please remove the mention ASAP

Daanvdplas: done

chungquantin (approved):

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+			let weight = |approve: u32, cancel: u32| -> Weight {
+				<T as Config>::WeightInfo::approve(cancel, approve)
+			};
+			let who = ensure_signed(origin.clone()).map_err(|e| e.with_weight(weight(0, 0)))?;
+			let current_allowance = AssetsOf::<T>::allowance(id.clone(), &who, &spender);
+			let spender = T::Lookup::unlookup(spender);
+			let id: AssetIdParameterOf<T> = id.into();
+			// If the new value is equal to the current allowance, do nothing.
+			if value == current_allowance {
+				return Ok(Some(weight(0, 0)).into());
```
evilrobot-01, line 123: No test coverage.

runtime/devnet/src/config/api.rs
```diff
@@ -0,0 +1,32 @@
+use crate::{config::assets::TrustBackedAssetsInstance, fungibles, Runtime, RuntimeCall};
+use codec::{Decode, Encode, MaxEncodedLen};
+use frame_support::traits::Contains;
+
+#[derive(Encode, Decode, Debug, MaxEncodedLen)]
+#[repr(u8)]
+pub enum RuntimeStateKeys<T: fungibles::Config> {
```
evilrobot-01, line 7:
```suggestion
/// A query of runtime state.
pub enum RuntimeRead<T: fungibles::Config> {
```

Align with terminology.

runtime/devnet/src/config/api.rs
```diff
@@ -0,0 +1,32 @@
...
+	#[codec(index = 150)]
+	Fungibles(fungibles::Read<T>),
```
evilrobot-01, line 9: Needs doc comments

runtime/devnet/src/extensions/mod.rs
```diff
@@ -247,48 +189,20 @@ where
...
-	// Generate runtime call to dispatch.
-	let call = RuntimeCall::PolkadotXcm(pallet_xcm::Call::send {
-		dest: Box::new(dest),
-		message: Box::new(VersionedXcm::V4(message)),
-	});
-	dispatch_call::<T, E>(env, call, origin, LOG_PREFIX)
+/// Wrapper to enable versioning of `RuntimeStateKeys`.
+#[derive(Decode, Debug)]
+enum VersionedStateRead<T: fungibles::Config> {
+	/// Version zero of reading state from api.
```
evilrobot-01, line 195:
```suggestion
	/// Version zero of reading state from runtime.
```

runtime/devnet/src/extensions/mod.rs
```diff
@@ -247,48 +189,20 @@ where
...
-	};
-	// TODO: revisit to replace with signed contract origin
-	let origin: RuntimeOrigin = RawOrigin::Root.into();
-	// Generate runtime call to dispatch.
-	let call = RuntimeCall::PolkadotXcm(pallet_xcm::Call::send {
-		dest: Box::new(dest),
-		message: Box::new(VersionedXcm::V4(message)),
-	});
-	dispatch_call::<T, E>(env, call, origin, LOG_PREFIX)
+/// Wrapper to enable versioning of `RuntimeStateKeys`.
```
evilrobot-01, line 192:
```suggestion
/// Wrapper to enable versioning of runtime state reads.
```

runtime/devnet/src/extensions/mod.rs
```diff
@@ -247,48 +189,20 @@ where
...
+#[derive(Decode, Debug)]
+enum VersionedStateRead<T: fungibles::Config> {
+	/// Version zero of reading state from api.
+	#[codec(index = 0)]
+	V0(RuntimeStateKeys<T>),
+}
+
+/// Wrapper to enable versioning of `RuntimeCall`.
```
evilrobot-01, line 200:
```suggestion
/// Wrapper to enable versioning of runtime calls.
```

runtime/devnet/src/lib.rs
```diff
@@ -1030,3 +960,20 @@ cumulus_pallet_parachain_system::register_validate_block! {
...
 	BlockExecutor = cumulus_pallet_aura_ext::BlockExecutor::<Runtime, Executive>,
 }
+
+#[cfg(test)]
+mod tests {
+	use crate::Runtime;
+	use std::any::TypeId;
+
+	#[test]
+	fn test_lookup_config() {
```
evilrobot-01, line 970: Very nice! Can you just add a comment to say that this ensures that the account if lookup doesnt not perform any state reads? Just so we know why its here and so that it doesnt get removed in the future without reason.

evilrobot-01 (changes requested):
Significant improvement in quality, well done! I think you can probably tell by my comments that I feel this is effectively done now, apart from a few finishing touches.

I think this is a good demonstration of how future implementations should be done and the expected levels of quality, cleanliness and simplicity.

pallets/api/src/fungibles/mod.rs
peterwht, line 123: This is a smaller function, so it's not a big deal. But generally, it's better to have just one return location when possible (as opposed to returning in 3 different locations). I recommend having a return weight variable at the top and then returning once at the end.

Reason being people reading the code may miss the returns, causing them to misunderstand the code.

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+			// because `approve_transfer` works additively (see pallet-assets).
+			if value > current_allowance {
+				AssetsOf::<T>::approve_transfer(
+					origin,
+					id,
+					spender,
+					value.saturating_sub(current_allowance),
+				)
+				.map_err(|e| e.with_weight(weight(1, 0)))?;
+				Ok(Some(weight(1, 0)).into())
```
peterwht, line 135: See above. Set variable instead.

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+			} else {
+				// If the new value is less than the current allowance, cancel the approval and set the new value
+				AssetsOf::<T>::cancel_approval(origin.clone(), id.clone(), spender.clone())
+					.map_err(|e| e.with_weight(weight(0, 1)))?;
+				if value.is_zero() {
+					return Ok(Some(weight(0, 1)).into());
```
peterwht, line 141: see above. Set variable instead.

runtime/devnet/src/extensions/mod.rs
```diff
@@ -373,118 +285,27 @@ impl TryFrom<u8> for FuncId {
...
-			Ok(<pallet_assets::Pallet<T, TrustBackedAssetsInstance> as MetadataInspect<
-				AccountId,
-			>>::decimals(id)
-			.encode())
+			Ok(fungibles::Pallet::<T>::allowance(id, &owner, &spender).encode())
 		},
+		TokenName(id) => Ok(fungibles::Pallet::<T>::token_name(id).encode()),
+		TokenSymbol(id) => Ok(fungibles::Pallet::<T>::token_symbol(id).encode()),
+		TokenDecimals(id) => Ok(fungibles::Pallet::<T>::token_decimals(id).encode()),
 		// AssetsKeys::AssetExists(id) => {
```
peterwht, line 309: Maybe add a TODO on this commented code?

runtime/devnet/src/extensions/mod.rs
```diff
@@ -673,501 +495,3 @@ mod tests {
 		});
 	}
 }
-// use enumflags2::BitFlags;
```
peterwht, line 676: enumflags2 is not being used anywhere but remains in the Cargo.toml

peterwht (approved):
Seconding Frank's comment, really good job on this!

node/src/command.rs
```diff
@@ -64,18 +64,9 @@ impl RuntimeResolver for PathBuf {
...
-		"dev-rococo" => Box::new(chain_spec::development_config(Relay::RococoLocal)),
-		#[cfg(feature = "paseo")]
-		"dev-paseo" => Box::new(chain_spec::development_config(Relay::PaseoLocal)),
-		#[cfg(not(feature = "paseo"))]
-		"pop-rococo" => Box::new(chain_spec::testnet_config(Relay::Rococo)),
-		#[cfg(feature = "paseo")]
-		"pop-paseo" => Box::new(chain_spec::testnet_config(Relay::Paseo)),
-		#[cfg(feature = "paseo")]
+		"dev" | "dev-paseo" => Box::new(chain_spec::development_config(Relay::PaseoLocal)),
+		"test" | "pop-paseo" => Box::new(chain_spec::testnet_config(Relay::Paseo)),
```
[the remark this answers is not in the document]
Daanvdplas: This was to be able to run the benchmarks didn't spend more time to it when it worked and because it is the solution in another PR

pallets/api/src/fungibles/mod.rs
Daanvdplas: See last commit

runtime/devnet/src/extensions/mod.rs
Daanvdplas: Removed

runtime/devnet/src/extensions/mod.rs
Daanvdplas: Thank you

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+		#[codec(index = 10)]
+		TokenDecimals(AssetIdOf<T>),
+	}
+
+	/// Configure the pallet by specifying the parameters and types on which it depends.
+	#[pallet::config]
+	pub trait Config: frame_system::Config + pallet_assets::Config<Self::AssetsInstance> {
+		/// The instance of pallet assets it is tightly coupled to.
+		type AssetsInstance;
+		/// Weight information for extrinsics in this pallet.
```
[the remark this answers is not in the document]
Daanvdplas: That is the second time, thanks for reminding me

pallets/api/src/fungibles/mod.rs
Daanvdplas: Thank you

Daanvdplas: I asked this before but not noted, what do you use for checking test coverage? Or was this just sharp as a razor?

runtime/devnet/src/extensions/mod.rs
Daanvdplas: Reworded it a bit more, check if you agree with pls

runtime/devnet/src/lib.rs
Daanvdplas: Yes good point.

pallets/api/src/fungibles/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+	#[derive(Encode, Decode, Debug, MaxEncodedLen)]
+	#[repr(u8)]
+	#[allow(clippy::unnecessary_cast)]
+	pub enum Read<T: Config> {
+		/// Total token supply for a given asset ID.
+		#[codec(index = 0)]
+		TotalSupply(AssetIdOf<T>),
+		/// Account balance for a given asset ID.
+		#[codec(index = 1)]
+		BalanceOf(AssetIdOf<T>, AccountIdOf<T>),
```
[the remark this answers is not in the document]
Daanvdplas: see last commit

pallets/api/src/fungibles/mod.rs
evilrobot-01: RR, run tests with coverage. I will get codecov added to this repo as well so its checked via CI too.

evilrobot-01 (approved):
Congrats! 🚀 I know you put a lot of work into this, so thank you for the time and dedication. I hope you are pleased with the result.

pallets/api/src/fungibles/mod.rs
evilrobot-01: Subjective, I think I prefer early returns as its means less complex code (additional variable carried through). No objections here tho.

Daanvdplas:
> I think this is a good demonstration of how future implementations should be done and the expected levels of quality, cleanliness and simplicity.

I 100% agree, it is amazing how much it has improved the last couple of weeks.

> Congrats! 🚀 I know you put a lot of work into this, so thank you for the time and dedication. I hope you are pleased with the result.

I'm very pleased with how it turned and how much I learned. Thanks for all the reviews and help, it got to this point by us together. Looking forward to implementing the rest of improvements and features and get the first pop api use case complete, ready for hacking!

pallets/api/src/fungibles/mod.rs
Daanvdplas: That would be really nice. In general, the CI needs some love. I am prioritising the remaining refactors because I want this over the finish line. If you could find the time it would be incredibly appreciated.

Issue #115, opened 2024-07-22
CI: auto formatter script


chungquantin:
Many parts of this repository is not formatted in the correct way, to reduce the time for development, we should add the auto formatter script to keep the codebase style consistent between contributors. 

The script should run the command `cargo +nightly fmt`

Daanvdplas:
https://github.com/paritytech/polkadot-sdk/blob/d1979d4be9c44d06c62aeea44d30f5f201d67b5e/.github/workflows/checks-quick.yml#L40

chungquantin:
Closed. We agreed to have only format check and let the PR owner to fix it instead of auto formatting.

PR #116, opened 2024-07-22
ci: add script for linting & formatting code

chungquantin:
Link to issue: https://github.com/r0gue-io/pop-node/issues/115

- [x] Apply lint suggestion
- [x] Format check all packages 
- [x] Apply the format (checkout branch)


Daanvdplas:
FYI: we have some formatting checks in place [here](https://github.com/r0gue-io/pop-node/blob/bd26cf05a3016ffeecabfbdce9b3c2356df2e355/.github/workflows/build.yml#L21). Here you propose CI to make formatting commits if it doesn't fulfil the requirements. In addition, the `nightly` formatting is used. The latter is something we were already talking about internally and could perhaps be a good change. Not sure about the auto formatting. What do you think @evilrobot-01?

peterwht:
Thanks for doing this Tin! 

- I am okay with auto-applying cargo fmt (maybe not nightly version though)
- I am slightly concerned with auto-applying clippy fixes as they can sometimes introduce unintended side-effects (rarely). I think it would be better to prevent merging until clippy has been resolved.

.github/workflows/lint-format.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          command: clippy
+          args: --fix --workspace
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Format check packages
+        run: cargo +nightly fmt --all -- --check
+      - name: Commit changes
+        uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: Apply formatting changes
```
evilrobot-01, line 34: Can we use conventional commit syntax please - e.g. `style: apply formatting changes`

https://www.conventionalcommits.org/en/v1.0.0/#summary

.github/workflows/lint-format.yml
```diff
@@ -0,0 +1,35 @@
...
+jobs:
+  format:
+    name: Format apply
+    timeout-minutes: 20
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
evilrobot-01, line 22: Think we should defer the toolchain to that defined in `rust-toolchain.toml`. It happens to be the same, but better to have this defined in one place. In other words, can we just remove the specification here.

evilrobot-01:
Instead of adding an additional workflow, can we rather rename `build.yml` to `ci.yml` and then add these changes into that workflow?

evilrobot-01:
Also imagine that this might create some 'churn' for those that dont auto-format locally, whereby branch is pushed, CI formats and adds a commit and then local branch is out of sync. This can hopefully be eliminated by a proper local setup, which probably enforces formatting locally more so than currently. Either you sort it out to avoid any hassle, or GH sorts it for you and you have to cope with any hassle.

Daanvdplas:
> Also imagine that this might create some 'churn' for those that dont auto-format locally, whereby branch is pushed, CI formats and adds a commit and then local branch is out of sync. This can hopefully be eliminated by a proper local setup, which probably enforces formatting locally more so than currently. Either you sort it out to avoid any hassle, or GH sorts it for you and you have to cope with any hassle.

For this reason, is it then not more simple if we do a CI check on nightly and have the developer fix it themselves?

chungquantin:
Hence, I think would be better to have: 
- Check `clippy` and prevent from merging until it is resolved
- For formatting, we still want to apply it automatically to the PR code on pushed to keep the codebase style consistent

chungquantin:
As there are a huge conflict in this PR, I will close it and create the another.

PR #71, opened 2024-04-09
feat: pallet assets pop api integration

Daanvdplas:
### Pop API + partial local fungibles implementation.

**Size of contract `pop-api/examples/fungibles/` compared to the psp22 extension in [ink-examples](https://github.com/use-ink/ink-examples/blob/main/psp22-extension/lib.rs):**
- Fungibles contract:
Original wasm size: 31.3K, Optimized: 7.6K
- Template contract:
Original wasm size: 30.3K, Optimized: 7.0K

The Pop API design is 95% in its final state. #94 will be the last addition which refactors the chain extension functions directly to the functions that are currently to be found in `pop_api/src/v0/assets/fungibles.rs` (see the issue description for explanation). Then some additional refactors need to be done or looked at (see todo's below). When this is all finished, the remaining local fungibles features can be added (again see below). Finally,  we want to include the new chain extension logic in an iterative manner. This is by feature gating the functionality between devnet and testnet runtime. At the moment this doesn't work because (I think) the node takes both runtimes and cannot make the separation clearly between those runtimes. Last, a contract similar to the example contract should be deployed on testnet so that we can test it while upgrading Pop Network to polkadot sdk latest version.

**Remaining TODOs (in the given order):**
- [ ] refactor (pop api): match chain extension functions to standard. #94
- [ ] refactor: check dependencies for coding guidelines #105
- [ ] refactor: pop api and primitives `Cargo.toml` to workspace dependencies #90 
- [ ] refactor: pop-api/examples to pop-api/integration-tests/contracts/ #95
- [ ] refactor: DispatchError to u32 conversion logic #91 
- [ ] refactor: define types used both in pop api as in chain extension in primitives #93
- [ ] refactor: exclude encoding of reading state results in chain extension #101
- [ ] refactor: simplify trait bounds in chain extension #102
- [ ] feat (pop api): approve, increase_allowance and decrease_allowance #104
- [ ] feat (pop api): add native token to local fungibles use case Pop api #97
- [ ] feat (pop api): implement asset management interfaces Pop API local fungibles #92
- [ ] feat (pop api): add burn and mint to local fungibles use case #98
- [ ] feat: add FungiblesError conversion logic for TokenError #99
- [ ] refactor: update (replace) old pop api code #103
- [ ] chore: feature gate primitives for different runtimes #100
- [ ] chore: deploy contract with local fungibles pop api on testnet for integration tests. #96

Issue #105, opened 2024-07-05
chore: coding guidelines and apply to code.


Daanvdplas:
- [x] Change CI to `cargo +nightly fmt`. This brings advantages such as formatting comments and dependencies.
- [x] Create coding guidelines in this repo.

EDIT: 
- Replaced the todo's related to pop api to workspace dependencies to a separate issue #206 . 
- Replaced the todo regarding scale inconsistency in #209 

If its already included in your work please continue and we can manage both issues in your PR.

evilrobot-01:
>     * [ ]  pop-api and primitives to workspace dependencies

Need to consider the effects of this when publishing to crates.io

Daanvdplas:
>Need to consider the effects of this when publishing to crates.io

That shouldn't be a problem. Have been working with `ink-sandbox` (crate) and it uses workspace dependencies ([ref](https://github.com/use-ink/ink/blob/master/crates/e2e/sandbox/Cargo.toml))

Daanvdplas:
Close #184

PR #121, opened 2024-07-23
refactor: build chain extension method

chungquantin:
Link to discussion: https://github.com/r0gue-io/pop-node/pull/71#discussion_r1683692879

Contract size: Original wasm size: 31.2K, Optimized: 7.5K

```
pub(crate) fn build_extension_method(
	version: u8,
	function: u8,
	module: u8,
	dispatchable: u8,
) -> ChainExtensionMethod<(), (), (), false> {
	ChainExtensionMethod::build(u32::from_le_bytes([version, function, module, dispatchable]))
}
```


evilrobot-01:
An alternative implementation which keeps things simpler imo, taking advantage of scopes: https://github.com/r0gue-io/pop-node/commit/ee4d4994bbef1bb6bc4a587b22e23d517d868677

- reuses existing `lib.rs `rather than introducing a new module for a single function. `build_extension_method` will probably be used in almost every submodule, so easier if its globally available rather than within a separate `utils` module.
- adds a `build_extension_method` at the `v0` module root, so we dont have to repeat V0 everywhere within the module
- adds simple `build_dispatch` and `build_read_state` functions as well, to avoid having to repeat the `DISPATCH` and `READ_STATE` constants throughout all the implemented functions. We could drop the `build_`* prefix on these functions too.
- no `pub(crate)` modifiers required

Daanvdplas:
Could you also add the contract size in the description of the PR with the changes made (with `--release`)

pop-api/src/lib.rs
```diff
@@ -33,6 +33,16 @@ mod constants {
 	pub(crate) const BALANCES: u8 = 10;
 }
 
+/// Helper method to build `ChainExtensionMethod``
```
evilrobot-01, line 36: Can we add comments for the arguments please? We can then simply copy them to the version specific functions.

pop-api/src/lib.rs
```diff
@@ -33,6 +33,16 @@ mod constants {
...
+pub fn build_extension_method(
```
evilrobot-01, line 37: Don't expect this to be used outside the crate so pub unnecessary for now. We want devs to use the API functions, hiding this complexity from them.

pop-api/src/v0/assets/mod.rs
```diff
@@ -91,13 +91,11 @@ const TRANSFER_APPROVED: u8 = 25;
 /// Move some assets from the sender account to another, keeping the sender account alive.
 #[inline]
 pub fn transfer_keep_alive(id: AssetId, target: AccountId, amount: Balance) -> Result<()> {
-	ChainExtensionMethod::build(u32::from_le_bytes([
-		V0,
-		DISPATCH,
+	build_dispatch(
 		ASSETS,
 		// E.D. is always respected with transferring tokens via the API.
```
evilrobot-01, line 96: Be good to place this comment above the build dispatch call, only so it fits on a single line.

pop-api/src/v0/assets/mod.rs
```diff
@@ -222,7 +220,7 @@ pub fn token_symbol(id: AssetId) -> Result<Vec<u8>> {
 
 #[inline]
 pub fn token_decimals(id: AssetId) -> Result<u8> {
-	ChainExtensionMethod::build(u32::from_le_bytes([V0, READ_STATE, ASSETS, TOKEN_DECIMALS]))
+	build_read_state(ASSETS, TOKEN_DECIMALS)
```
evilrobot-01, line 223: We could have another set of build_dispatch and build_read_state functions defined at the v0::assets module scope to remove the need for the repeated ASSETS constant use.

Essentially providing the same functions at each level, narrowing the scope from general to somewhat specific.

pop-api/src/v0/mod.rs
```diff
@@ -16,3 +22,22 @@ impl From<StatusCode> for Error {
 		value.0.into()
 	}
 }
+
+/// Helper method to build `ChainExtensionMethod``
+pub fn build_extension_method_v0(
```
evilrobot-01, line 27: As above, can remove pub and add arguments doc comments. Same with the other two functions below.

evilrobot-01 (changes requested):
Looking good, think there are a few final possible improvements and then should be good to merge.

pop-api/src/lib.rs
```diff
@@ -34,7 +34,12 @@ mod constants {
 }
 
 /// Helper method to build `ChainExtensionMethod``
-pub fn build_extension_method(
+///
+/// - `version`: The version of the chain extension
+/// - `function`: The ID of the function
+/// - `module`: The index of the runtime module
+/// - `dispatchable`: The index of the module dispatchable functions
+fn build_extension_method(
```
Daanvdplas, line 42: Please document like:

```
/// <Description>.
///
/// Parameters:
/// - '<parameter>': <description>.
```

pop-api/src/v0/mod.rs
```diff
@@ -33,11 +37,17 @@ pub fn build_extension_method_v0(
...
+/// - `dispatchable`: The index of the module dispatchable functions
+fn build_dispatch(module: u8, dispatchable: u8) -> ChainExtensionMethod<(), (), (), false> {
 	build_extension_method_v0(DISPATCH, module, dispatchable)
 }
 
 /// Helper method to build a dispatch call `ChainExtensionMethod`
-pub fn build_read_state(module: u8, dispatchable: u8) -> ChainExtensionMethod<(), (), (), false> {
+///
+/// - `module`: The index of the runtime module
+/// - `dispatchable`: The index of the module dispatchable functions
```
Daanvdplas, line 50: This is strictly not a dispatchable but a `state_query`

pop-api/src/lib.rs
Daanvdplas: This is inconsistent in the repo and will be fixed (#105), my apologies.

chungquantin: @Daanvdplas So should we ignore it and make a refractor in the other PR?

Daanvdplas: No you can update here already

evilrobot-01 (changes requested):
Looking nice and clean! One final improvement suggested.

pop-api/src/v0/mod.rs
```diff
@@ -16,3 +22,35 @@ impl From<StatusCode> for Error {
...
 	}
 }
+
+/// Helper method to build `ChainExtensionMethod` for version `v0`
+///
+/// Parameters:
+/// - 'function': The ID of the function
+/// - 'module': The index of the runtime module
+/// - 'dispatchable': The index of the module dispatchable functions
+fn build_extension_method_v0(
```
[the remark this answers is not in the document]
chungquantin: This is resolved. Thanks for the feedback

evilrobot-01 (approved):
LGTM, thanks for your patience! This will be a very important piece to everything that needs to be built out, so great to have such a tidy implementation.

Daanvdplas:
I think something went wrong when merging 'daan/api' because the helper functions are now not being used in the fungibles apis.

chungquantin:
@Daanvdplas Rebased. Should be ok now. Can you check it again?

Daanvdplas:
Good work @chungquantin!

PR #122, opened 2024-07-23
refactor: remove read state encoding

### LINKED ISSUE #101: refactor: exclude encoding of reading state results in chain extension

Daanvdplas:
Closes #101

- Removes the decoding of the read state result
- refactors the read state implementation by matching on the decoded key in the pallet in stead of the chain extension
- adds a filter for state queries (useful for the different runtimes)

Contract size from 7.4K to 7.1K

Created an issue to look into state query error handling #123


chungquantin (approved):

pallets/api/src/fungibles/mod.rs
```diff
@@ -179,7 +203,7 @@ pub mod pallet {
 		///
 		/// # Returns
 		/// The balance of the specified account, or an error if the operation fails.
-		pub fn balance_of(id: AssetIdOf<T>, owner: &AccountIdOf<T>) -> BalanceOf<T> {
+		pub(crate) fn balance_of(id: AssetIdOf<T>, owner: &AccountIdOf<T>) -> BalanceOf<T> {
```
evilrobot-01, line 206: Do we even needs these functions anymore? It's boilerplate which can be replaced by moving the assets function call into the match of read_state

pallets/api/src/fungibles/mod.rs
```diff
@@ -159,14 +159,38 @@ pub mod pallet {
...
+		/// * `key` - An instance of `FungiblesKey<T>`, which specifies the type of state query and
+		/// 		  the associated parameters.
+		///
+		/// # Returns
+		/// A vector of bytes representing the encoded result of the state query.
+		pub fn read_state(key: FungiblesKey<T>) -> Vec<u8> {
+			use FungiblesKey::*;
+
+			match key {
+				TotalSupply(id) => Self::total_supply(id).encode(),
```
evilrobot-01, line 177: Probably fine for now, but querying scale encoded state, to decode it, to reencode is inefficient. A raw storage read might be better, but the downside is not benefiting from the assets pallet API protecting against inner changes. Unit tests might resolve this gap.

pallets/api/src/fungibles/mod.rs
```diff
@@ -159,14 +159,38 @@ pub mod pallet {
...
+		/// This function matches the key to determine the type of state query and returns the
+		/// encoded result.
+		///
+		/// # Arguments
+		/// * `key` - An instance of `FungiblesKey<T>`, which specifies the type of state query and
+		/// 		  the associated parameters.
+		///
+		/// # Returns
+		/// A vector of bytes representing the encoded result of the state query.
+		pub fn read_state(key: FungiblesKey<T>) -> Vec<u8> {
```
evilrobot-01, line 173: Still maintain we should avoid using Key and use something more abstract. Also, as this is defined in the fungibles pallet, I don't see that we need the Fungibles prefix. The pallet norm is Error and Event, not FungiblesError and FungiblesEvent. Suggest something like Query or Read.

Whilst perhaps not applicable to this pallet, I'm sure there will come a time where there is a query which doesn't take an argument or performs multiple reads to calculate a result.

pallets/api/src/fungibles/mod.rs
Daanvdplas: I think when more functions get added this might be cleaner. Otherwise it will be a big match statement where the implementation of those state queries also could get more complex in stead of a one liner.

evilrobot-01: Youve got 10 lines of code, including comments, for something that could be a one liner in the match. That's a pattern that requires a decent amount of extra work/reviews that are simply unnecessary right now. We need to get stuff finished and deployed, we get there by writing as little code as possible. This is all internal to the crate, it can easily all be refactored into separate functions like this if ever required later.

pallets/api/src/fungibles/mod.rs
Daanvdplas: I have tried this out with `total_supply` but it brings other difficulties with it. For this case I am querying `AssetDetails` from `Asset` storage item and need to obtain the value from field [supply](https://github.com/paritytech/polkadot-sdk/blob/18db502172bdf438f086cd5964c646b318b8ad37/substrate/frame/assets/src/types.rs#L62). 

The flow is creating a raw key and obtaining an encoded `AssetDetails` value. This then has to be stripped manually in order to obtain the specific bytes to decode into the `supply` value. Before I implement this I want a second opinion about whether this is worth it because it feels very error prone.

I also can't decode the `AssetDetails` value to easily obtain the supply value because `supply` is not public. In addition, this comes back to the same - decoding and encoding the value.

pallets/api/src/fungibles/mod.rs
Daanvdplas: https://github.com/r0gue-io/pop-node/issues/127

pallets/api/src/fungibles/mod.rs
Daanvdplas: Yes this makes sense, refactored.

pallets/api/src/fungibles/mod.rs
```diff
@@ -180,61 +180,33 @@ pub mod pallet {
 	}
 
 	impl<T: Config> Pallet<T> {
-		/// Returns the total token supply for a given asset ID.
+		/// Reads the state of the fungible asset based on the provided key.
```
evilrobot-01, line 183:
```suggestion
		/// Reads fungible asset state based on the provided value.
```

pallets/api/src/fungibles/mod.rs
```diff
@@ -180,61 +180,33 @@ pub mod pallet {
...
-		/// Returns the token decimals for a given asset ID.
-		///
-		/// # Parameters
-		/// * `id` - The ID of the asset.
-		pub fn token_decimals(id: AssetIdOf<T>) -> u8 {
-			<AssetsOf<T> as MetadataInspect<AccountIdOf<T>>>::decimals(id)
+		/// # Arguments
+		/// * `key` - An instance of `FungiblesKey<T>`, which specifies the type of state query and
+		/// 		  the associated parameters.
+		pub fn read_state(key: Read<T>) -> Vec<u8> {
```
evilrobot-01, line 191:
```suggestion
		pub fn read_state(value: Read<T>) -> Vec<u8> {
```

pallets/api/src/fungibles/mod.rs
```diff
@@ -180,61 +180,33 @@ pub mod pallet {
...
 		///
-		/// # Parameters
-		/// * `id` - The ID of the asset.
-		pub fn total_supply(id: AssetIdOf<T>) -> BalanceOf<T> {
-			AssetsOf::<T>::total_supply(id)
-		}
-
-		/// Returns the account balance for the specified `owner` for a given asset ID. Returns `0` if
-		/// the account is non-existent.
+		/// This function matches the key to determine the type of state query and returns the
```
evilrobot-01, line 185:
```suggestion
		/// This function matches the value to determine the type of state query and returns the
```

pallets/api/src/fungibles/mod.rs
```diff
@@ -180,61 +180,33 @@ pub mod pallet {
...
-		}
-
-		/// Returns the token decimals for a given asset ID.
-		///
-		/// # Parameters
-		/// * `id` - The ID of the asset.
-		pub fn token_decimals(id: AssetIdOf<T>) -> u8 {
-			<AssetsOf<T> as MetadataInspect<AccountIdOf<T>>>::decimals(id)
+		/// # Arguments
+		/// * `key` - An instance of `FungiblesKey<T>`, which specifies the type of state query and
```
evilrobot-01, line 189: This is the problem with referencing types in comments. You change the type name and then the comments are now out of date.

```suggestion
		/// * `value` - An instance of `Read<T>`, which specifies the type of state query and
```

runtime/devnet/src/extensions/mod.rs
```diff
@@ -169,19 +166,24 @@ where
 {
 	const LOG_PREFIX: &str = " read_state |";
 
-	// Prefix params with version, pallet, index to simplify decoding.
+	// Prefix params with version, pallet, index to simplify decoding, and decode parameters for
+	// reading state.
 	params.insert(0, version);
 	params.insert(1, pallet_index);
 	params.insert(2, call_index);
 	let key = <VersionedStateRead<T>>::decode(&mut &params[..])
```
evilrobot-01, line 174:
```suggestion
	let read = <VersionedStateRead<T>>::decode(&mut &params[..])
```

runtime/devnet/src/extensions/mod.rs
```diff
@@ -169,19 +166,24 @@ where
...
 		.map_err(|_| DispatchError::Other("DecodingFailed"))?;
 
+	// Charge weight for doing one storage read.
+	env.charge_weight(T::DbWeight::get().reads(1_u64))?;
 	let result = match key {
-		VersionedStateRead::V0(key) => match key {
-			RuntimeRead::Fungibles(key) => read_fungibles_state::<T, E>(key, env),
+		VersionedStateRead::V0(key) => {
```
evilrobot-01, line 180:
```suggestion
		VersionedStateRead::V0(read) => {
```

runtime/devnet/src/extensions/mod.rs
```diff
@@ -169,19 +166,24 @@ where
...
+			ensure!(AllowedApiCalls::contains(&key), DispatchError::Other("UnknownCall"));
```
evilrobot-01, line 181: Might be better to define a function on DispatchError::unknown_call() rather than these static definitions everywhere.

evilrobot-01 (changes requested):
A few minor comments which need addressing.

pallets/api/src/fungibles/mod.rs
evilrobot-01: Yeah, overkill for a nested field. We can revisit one day perhaps. Thanks for looking at it in more detail.

evilrobot-01:
> * refactor: charging weight for state queries

If it will always be one or more reads, we could attach an attribute to the read enum variants which defines the number of reads. That way the extension can just use that static number to precharge the relevant weight, but more importantly it is defined in the pallet. Alternatively it could be an actual weight value, but that might be something for the future.

runtime/devnet/src/extensions/mod.rs
Daanvdplas: Great suggestion

Daanvdplas: Used constants in stead of functions.

evilrobot-01 (commented):
Looks good, has merge conflicts tho.

evilrobot-01 (approved):

Issue #119, opened 2024-07-23
refactor: contract integration test should streamline the error for better testing


chungquantin:
cc: @Daanvdplas 

```rs
fn decoded<T: Decode>(result: ExecReturnValue) -> T {
	match <T>::decode(&mut &result.data[2..]) {
		Ok(value) => value,
		Err(_) => panic!("\nTest failed by trying to decode `{:?}` into `T`\n", result),
	}
}
```

At the moment, any method failed will end up at line `13:39` which is inside the `decoded` method. This is very hard to debug. My recommendation is to refractor the `decoded` method to return `Result<T>` so it can streamline the error back to the main function. Something like:

```rs
fn decoded<T: Decode>(result: ExecReturnValue) -> Result<T> {
	<T>::decode(&mut &result.data[2..]).map_err("\nTest failed by trying to decode `{:?}` into `T`\n", result)
}
```

Daanvdplas:
Yes this was something that could be improved, great suggestion!

Daanvdplas:
Fixed by #130

PR #130, opened 2024-07-26
refractor: streamline error from the decoded method by returning `Result`

chungquantin:
Issue: https://github.com/r0gue-io/pop-node/issues/119


pop-api/integration-tests/src/local_fungibles.rs
```diff
@@ -327,26 +325,26 @@ fn transfer_works() {
 		// Asset does not exist.
 		assert_eq!(
 			decoded::<Error>(transfer(addr.clone(), 1, BOB, amount,)),
-			Module { index: 52, error: 3 },
+			Ok(Module { index: 52, error: 3 }),
```
Daanvdplas, line 328: Feels weird that those errors are wrapped in an `Ok()`

chungquantin: `Ok()` here means decoded successfully so I think it makes sense.

Issue #120, opened 2024-07-23
fix (test): improve error handling when calling contract with missing parameter


Daanvdplas:
In the integration tests if we are calling a contract with a missing parameter (or incorrectly encoded probably as well) it will revert but there it no clear error returned. Would be good if this can be improved for future integration tests.

chungquantin:
To provide more context, with out the data field (even empty vec<u8>), the contract call will be reverted:
```diff
fn transfer_from(
	addr: AccountId32,
	asset_id: AssetId,
	delegate: AccountId32,
	from: AccountId32,
	to: AccountId32,
	value: Balance,
+	data: Vec<u8>,
) -> ExecReturnValue {
	let function = function_selector("transfer_from");
	let params =
                 [function, asset_id.encode(), from.encode(), to.encode(), value.encode()]
+		[function, asset_id.encode(), from.encode(), to.encode(), value.encode(), data.encode()]
			.concat();
	let result = bare_call_by(addr, delegate, params, 0).expect("should work");
	result
}
```

Daanvdplas:
Closed by #133

PR #133, opened 2024-07-26
refactor: api integration tests

### LINKED ISSUE #123: refactor: state query error handling
Right now if you call a state query function (e.g. `fn total_supply`) and it is not allowed in the runtime (i.e. `AllowedApiCall` for `RuntimeStateKeys` #122) there is no error handling in api ink! library. 

Edit: error handling was already implemented, the integration tests had to be refactored.

Daanvdplas:
Closes #123 & #120

- Remove helper functions to the bottom of the file
- improved return value when calling the contracts so we can assert on a success and error at the same time.
- Have a consistent assert schema: `value, expected`
- if a contract is reverted it is clear from which function it is coming from.

Open question that might improve things even more:
https://substrate.stackexchange.com/questions/11765/what-should-data-be-decoded-to-in-execreturnvalue


evilrobot-01:
The link you included on SE is broken.

Daanvdplas:
Hmm that is weird. Here it is:

What does [data](https://paritytech.github.io/ink/ink_env/fn.debug_message.html) represent? In other words, to what type could I decode it?

Daanvdplas:
Closed in favour of #148

pop-api/integration-tests/src/local_fungibles.rs
```diff
@@ -625,3 +352,269 @@ fn token_metadata_works() {
...
+		Err(result) => panic!("Contract reverted: {:?}", result),
+	}
+}
+
+// Call balance_of contract message.
+fn balance_of(addr: AccountId32, asset_id: AssetId, owner: AccountId32) -> Result<Balance, Error> {
+	let function = function_selector("balance_of");
+	let params = [function, asset_id.encode(), owner.encode()].concat();
+	let result = bare_call(addr, params, 0).expect("should work");
+	match decoded::<Result<Balance, Error>>(result) {
```
chungquantin, line 376: These lines can be cleaner if we use `map_err`
```rs
decoded::<Result<Balance, Error>>(result).map_err(|_| panic!("Contract reverted: {:?}", result))
```

chungquantin: Another thing is this match logic is reused for all the other methods, do you think we should have another shared method called `contract_exec_decoded` or something else that can be reused for all the other methods?

chungquantin (changes requested):

pop-api/integration-tests/src/local_fungibles.rs
Daanvdplas: Regarding your first comment; the code will then return a `Result<Result<_, _>>` which we don't want. 

Second comment; the problem with that is when the panic occurs it is not clear from which test it is coming and which check within that test. As a result, we have more duplicated code but know immediately which check is failing which improves development.

Daanvdplas: To your first comment, I will refactor it to:
```
decoded::<Result<(), Error>>(result).unwrap_or_else(|_| panic!("Contract reverted: {:?}", result))
```

Daanvdplas:
Closing to make the PR to another branch

pallets/api/src/fungibles/mod.rs
```diff
@@ -214,23 +214,21 @@ pub mod pallet {
...
 			}
-
-			current_allowance.saturating_reduce(value);
 			// Cancel the aproval and set the new value if `current_allowance` is more than zero.
 			AssetsOf::<T>::cancel_approval(origin.clone(), id.clone(), spender.clone())
 				.map_err(|e| e.with_weight(Self::weight_approve(0, 1)))?;
-
-			if current_allowance.is_zero() {
+			let new_allowance = current_allowance.saturating_sub(value);
+			if new_allowance.is_zero() {
```
Daanvdplas, line 228: Changed this to be more clear that it is the new allowance value that will be set.

chungquantin: Thanks @Daanvdplas

chungquantin (approved):

pallets/api/src/fungibles/mod.rs
```diff
@@ -214,23 +214,21 @@ pub mod pallet {
...
+			let current_allowance = AssetsOf::<T>::allowance(id.clone(), &who, &spender);
 			let spender = T::Lookup::unlookup(spender);
 			let id: AssetIdParameterOf<T> = id.into();
 
 			if value.is_zero() {
 				return Ok(Some(Self::weight_approve(0, 0)).into());
 			}
-
-			current_allowance.saturating_reduce(value);
 			// Cancel the aproval and set the new value if `current_allowance` is more than zero.
```
peterwht, line 224: This comments seems like it is out of date.

peterwht (commented):
Looks good, small comments to resolve.

GitHub is not letting me put a comment on it. But the if-chain on line 156 has a clippy warning.

Please resolve it.

```
match value.cmp(current_allowance) =>
    Equal => {...},
    Less => {...},
    Greater => {...},
```

pallets/api/src/fungibles/mod.rs
Daanvdplas: Thank you

PR #139, opened 2024-07-27
refactor: migrate integration tests to paseo

### LINKED ISSUE #138: chore: update integration tests to paseo
Polkadot 1.2 uses SDK 1.7

evilrobot-01:
Migrates integration tests to use the Paseo runtime.

**IMPORTANT NOTE:** switching this dependency means we are effectively stuck on SDK 1.7 until Paseo updates. The runtimes at https://github.com/polkadot-fellows/runtimes have been [updated to 1.13](https://github.com/polkadot-fellows/runtimes/commit/c5080e43321476ec82d69b710ebfaafb634f7475) recently in `main`, but no release has been generated yet. Should we wish to update the SDK version we use, we can simply replace the paseo dependencies with the corresponding runtimes from https://github.com/polkadot-fellows/runtimes. I am doing this in another project for now.

Closes #138


integration-tests/src/chains/paseo/genesis.rs
```diff
@@ -2,15 +2,15 @@ use emulated_integration_tests_common::{
 	accounts, build_genesis_storage, get_account_id_from_seed, get_from_seed, get_host_config,
 	validators,
 };
+use paseo_runtime_constants::currency::UNITS as ROC;
```
Daanvdplas, line 5:
```suggestion
use paseo_runtime_constants::currency::UNITS as PAS;
```

runtime/devnet/src/config/xcm.rs
```diff
@@ -28,7 +28,8 @@ use xcm_executor::XcmExecutor;
 parameter_types! {
 	pub const RelayLocation: Location = Location::parent();
 	pub AssetHub: Location = Location::new(1, [Parachain(1000)]);
-	pub const RelayNetwork: Option<NetworkId> = None;
```
Daanvdplas, line 31: Do we know why this was `None`?

Daanvdplas (approved):

runtime/devnet/src/config/xcm.rs
evilrobot-01: Template: https://github.com/paritytech/polkadot-sdk/blob/10b8039a53ddaa10ca37fe70c64e549c649a22dc/templates/parachain/runtime/src/configs/xcm_config.rs#L27

integration-tests/src/chains/paseo/genesis.rs
evilrobot-01: Thanks, resolved.

runtime/testnet/src/config/xcm.rs
```diff
@@ -28,7 +28,8 @@ use xcm_executor::XcmExecutor;
 parameter_types! {
 	pub const RelayLocation: Location = Location::parent();
 	pub AssetHub: Location = Location::new(1, [Parachain(1000)]);
-	pub const RelayNetwork: Option<NetworkId> = None;
+	// 0x77afd6190f1554ad45fd0d31aee62aacc33c6db0ea801129acb813f913e0764f from https://dotapps-io.ipns.dweb.link/?rpc=wss%3A%2F%2Frpc.ibp.network%2Fpaseo#/explorer/query/0
+	pub const RelayNetwork: Option<NetworkId> = Some(ByGenesis([119,175,214,25,15,21,84,173,69,253,13,49,174,230,42,172,195,60,109,176,234,128,17,41,172,184,19,249,19,224,118,79]));
```
evilrobot-01, line 32: I think this will be problematic when testing locally, as the genesis block hash will differ when launched via zombienet.

Daanvdplas: Is this a blocker?

evilrobot-01: Not sure. Assume it would only be an issue when using xcm on a local testnet. Emulated tests are not affected.

@al3mart also mentioned that there is a risk that a regenesis of Paseo breaks this, even tho it appears to be the most accurate way. Having said that, the other available enums presumably work on name, so maybe sticking with `Polkadot` is better.

al3mart: Paseo should have not a regenesis, but if if this should be the case then it would need changing.
Given that Paseo is using [`NetworkId::Polkadot`](https://github.com/paseo-network/runtimes/blob/2fa66a19eef96bd2c079385057fab2564d09b751/relay/paseo/src/xcm_config.rs#L56) we should be good doing the change.

al3mart (approved):

al3mart (commented):
As mentioned, the `RelayNetwork` could be convenient to address now if we believe that we will hit that versioning constrain sometime soon. Note that Paseo follows Polkadot's enacted runtime versions.

That said, it should be most likely for Paseo to change its xcm configuration to use something as `NetworkId::Paseo` than a regenesis of the network.

evilrobot-01:
I was more concerned whether using this value might one day cause sending xcm messages on a local testnet to fail, if the paseo-local relay genesis block is set by zombienet for example, based on some network configuration file.

runtime/testnet/src/config/xcm.rs
evilrobot-01: Updated to `Polkadot`

Daanvdplas (approved):

PR #140, opened 2024-07-27
chore: remove fund-dev-accounts script

### LINKED ISSUE #137: chore: remove fund-dev-accounts script
No longer required after #89

evilrobot-01:
Removes the script for funding dev accounts as this is now included within the genesis state of the `paseo.toml` network config file.

Closes #137


Daanvdplas (approved):

al3mart (approved):

evilrobot-01:
@peterwht it would be good if you could give this a thumbs up, as you built it.

Testing should just be a case of launching a network and checking the balances set via zn overrides.

PR #144, opened 2024-07-28
build(testnet): bump version

evilrobot-01:
Bumps runtime version ahead of next release, purely to resolve a node syncing issue introduced with the last version bump of dependencies (i.e. 1.7 SDK update via `psvm -v 1.7.0`). 

The executor used is still the below, so the native wasm now built is panicing when verifying state roots, presumably due to the changes introduced in sp_api. 

`type ParachainExecutor<E> = NativeElseWasmExecutor<E>;`

This version bump would have had to have happend in any case, but doing so ensures that the onchain wasm is used and the node syncs.

Further investigation and testing should be carried out.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9784,7 +9784,7 @@ dependencies = [
 
 [[package]]
 name = "pop-runtime-testnet"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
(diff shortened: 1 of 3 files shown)


Daanvdplas (approved):
Thank you for figuring this out!

al3mart (approved):

PR #145, opened 2024-07-29
ci: apply format

chungquantin:
- [x] Applying the format code after linting


chungquantin:
After a discussion, we think it would be better to just check formatting (implemented) without apply changes, hence, this PR is no longer needed.

PR #146, opened 2024-07-29
ci: deterministic build

Daanvdplas:
See https://github.com/r0gue-io/pop-node/pull/146#issuecomment-2264182742


evilrobot-01:
Should be good now, at least from a manual run perspective. Example manual run, selecting the option to build runtimes only, at https://github.com/r0gue-io/pop-node/actions/runs/10207107642.

Will try a quick test release.

al3mart (approved):

peterwht (approved):

Issue #97, opened 2024-07-05
feat (pop api): add native token to local fungibles use case


Daanvdplas:
This will cause a problem with the errors `InsufficientBalance` and `NoBalance`. Probably we will need to add an error `NoFees` and `NoBalance` which are the errors when there are no funds for fees and deposit, respectively.

First needs #71 merged.

chungquantin:
## About the open questions: 

The discussion related to the design of the returned error can be referenced from this: https://github.com/paritytech/polkadot-sdk/issues/2240

From my perspective, adding `NoFees` would require extra implementation on the pop-api side to check the `total_weight` the method takes and the `total_balance` of the account (requires additional runtime dispatchs for on-chain data retrieval). Even though this does create a better user experience, I don't think trying to optimize it at this stage is recommended. 

If there can be better implementation to cover the case without adding extra dispatches, then adding `NoFees` would bring the user experience to the next level.

Daanvdplas:
Thanks for looking into this. Lets keep it as simple as possible for now.

peterwht:
Maybe a little late for this comment. But why do we need the API to support native balance transfers from `pallet-balances`?

ink! supports this by default: https://github.com/use-ink/ink-examples/blob/main/contract-transfer/lib.rs#L35

chungquantin:
@peterwht We want the users only need to care about using pop-api for interacting with the runtime and local fungible is more than just `transfer()`. But for now, seems like implementing the native balance in the same fungible pallet with asset is not very clean as there are too many mixed states. Would love to hear some thoughts from @Daanvdplas

PR #147, opened 2024-07-30
feat: add native token support for fungibles api

chungquantin:
- Link to issue: https://github.com/r0gue-io/pop-node/issues/97
- Reference: https://docs.rs/pallet-asset-conversion/19.0.0/pallet_asset_conversion/
### Remaining TODOs
- [x] Migrate from single `Asset` instance to multiple asset kinds using `NativeOrWithId`
- [x] Migrate existing methods:
  - [x] `transfer()`
  - [x] `transfer_from()` (throw `UnsupportedMethod`)
  - [x] `approve()` (throw `UnsupportedMethod`)
  - [x] `increase_allowance()` (throw `UnsupportedMethod`)
  - [x] `decrease_allowance()` (throw `UnsupportedMethod`)
- [x] Write the custom benchmarking function for the fungible api method that support the native token
- [x] Refactor POP API to support native token
- [x] Migrate read state methods
- [x] Refactor contract integration tests and pallet unit testing to cover the native token
  - [x] `transfer()`
  - [x] `transfer_from()` (throw `UnsupportedMethod`)
  - [x] `approve()` (throw `UnsupportedMethod`)
  - [x] `increase_allowance()` (throw `UnsupportedMethod`)
  - [x] `decrease_allowance()` (throw `UnsupportedMethod`)

## Concerns

Right now, the pallet is a tight coupled because it needs both pallet-assets and pallet-balances to add the native token to the fungible pallet functionalities.

However, we need create a shared type for the pallet_balances::Balance and pallet_assets::Balance because the BalanceOf<T> is used for parameter like amount in transfer() method. My implementation right now is to use of UnionOf and NativeOrWithId to provide the loose coupling to support type conversion but it would make the Fungibles pallet code mixes between loose coupling and tight coupling.


chungquantin:
Hi @Daanvdplas @evilrobot-01, before I finish other TODOs in this PR, could you help me to give it a quick review on the implementation using `NativeOrWithId` trait? This is a standard used in Polkadot SDK for `pallet-asset-conversion` as well. If there is a better and cleaner way to work on this, please let me know!

Daanvdplas (commented):
Well done for doing proper research and looking for standards in the Polkadot SDK! I like the use of an enum in stead of a hard coded number but it does require some extra conversions. 

Please go ahead with the implementation since it is hard to predict whether this causes problems in the chain extension or pallet. If we end up changing it, it won't be to much of a hassle.

Daanvdplas:
Not sure if existing methods besides `transfer` need to be adjusted. No functionality is provided in the balances pallet for approving. `balance_of` should be implemented.

As for the integration test. Could you perhaps write a test `native_token_fungibles_works()`. Here we test the `transfer` and `balance_of` functionality separately.

`total_supply()` gets complicated because we would need to query the relay chain. `allowance` is not possible due to first point I made above. The token metadata state reads I'm not sure, perhaps this information is stored on our chain.

chungquantin:
> Not sure if existing methods besides `transfer` need to be adjusted. No functionality is provided in the balances pallet for approving. `balance_of` should be implemented.
> 
> As for the integration test. Could you perhaps write a test `native_token_fungibles_works()`. Here we test the `transfer` and `balance_of` functionality separately.
> 
> `total_supply()` gets complicated because we would need to query the relay chain. `allowance` is not possible due to first point I made above. The token metadata state reads I'm not sure, perhaps this information is stored on our chain.

@Daanvdplas What I am doing in my local code is to throw `UnsupportedMethod` (defined in the fungibles pallet) back to the contract for every methods that does not support the native balance. To keep asset type consistent, I think every methods should receive `T::AssetKind` as a parameter does not matter it support native token or not. 
```
		#[pallet::call_index(5)]
		#[pallet::weight(<T as Config>::WeightInfo::approve(1, 1))]
		pub fn approve(
			origin: OriginFor<T>,
			asset: T::AssetKind,
			spender: AccountIdOf<T>,
			value: BalanceOf<T>,
		) -> DispatchResultWithPostInfo {
			match T::AssetCriteria::convert(asset) {
				Either::Left(_) => todo!(),
				Either::Right(id) => Self::do_approve_asset(origin, id, spender, value),
			}
		}
```

evilrobot-01:
Not a review, just a few random comments:

- Do we need Kind in the AsetKind name? Is that not inferred by the fact that it is a type with some bounds?
- Why retain the tight coupling to the assets pallet on the Config trait instead of just using the fungibles traits from frame_system. 
- The terminology used throughout should be 'fungible'-based rather than assets.
- Will this easily support additional pallet instances, e.g. a foreign assets instance with a different id type perhaps?

chungquantin:
> * Do we need Kind in the AssetKind name? Is that not inferred by the fact that it is a type with some bounds?

`AssetKind` does described the parameter accepts different asset class, not only `Asset`, in this case, `Balance` and `Asset` through the enum `NativeOrWithId`. Do you have any recommendation? 

> * Will this easily support additional pallet instances, e.g. a foreign assets instance with a different id type perhaps?

Yes I don't think this design is limited to that. Reference from `pallet-asset-conversion` used for foreign use cases: 
```rs
pub type NativeAndAssets = fungible::UnionOf<
	Balances,
	LocalAndForeignAssets,
	TargetFromLeft<TokenLocationV3, xcm::v3::Location>,
	xcm::v3::Location,
	AccountId,
>;
```

evilrobot-01:
The overuse of the term asset in the pallets public interface seems somewhat like a leaky abstraction. A first step might be to replace the term Asset with Fungible, which then uses the same terminology as the pallet name.

T::AssetKind to T::Fungible as an example.

An additional step might be to add a trait which can be implemented in the runtime via an adapter struct, to minimize the tight coupling to the balances and assets pallets. I get that this pallet is wrapping them, but as more types of fungibles are added it may become congested with dependencies. This can be addressed in future tho, only if required.

Daanvdplas:
> Why retain the tight coupling to the assets pallet on the Config trait instead of just using the fungibles traits from frame_system.

Functions such as `total_supply` are implemented on the pallet and therefore tight coupling is required.

chungquantin:
> > Why retain the tight coupling to the assets pallet on the Config trait instead of just using the fungibles traits from frame_system.
> 
> Functions such as `total_supply` are implemented on the pallet and therefore tight coupling is required.

To expand more on this, there are functionalities which we have to customize our own method to return the error for unsupported functionalities, for example, like @Daanvdplas said, `total_supply()` or any functionalities related to delegate spending permission. 

```rs
TokenSymbol(asset) => match T::AssetCriteria::convert(asset) {
	Either::Left(_) => todo!(),
	Either::Right(id) => {
		<AssetsOf<T> as AssetsMetadataInspect<AccountIdOf<T>>>::symbol(id).encode()
	},
},
```

evilrobot-01:
> > Why retain the tight coupling to the assets pallet on the Config trait instead of just using the fungibles traits from frame_system.
> 
> Functions such as `total_supply` are implemented on the pallet and therefore tight coupling is required.

If there is no trait for total supply then one could just create a trait TotalSupply and then implement it for an adapter struct in the runtime, which defers to the relevant pallet function. A known example of this is the benchmark helper. Tight coupling is not required. It's just the simplest. Introducing the native assets just makes me question the tight coupling moving forward.

As I said, can be refactored in the future if need be.

Daanvdplas (commented):
I was not able to do a proper review because there is a mix of tightly coupling and lously coupling. There are also todo's in the PR for some of the read state functions which make it hard to get a view on the effect of this feature on the pallet as a whole. 

Nevertheless, your work has shown the difficulties of this implementation, especially without tightly coupling pallet assets and pallet balances. I think it will be easier if you first add the native token functionality to the pallet as a whole, without lously coupling, like Frank suggested.

Next steps would be to continue with lously coupling the pallet. This can be done in a separate PR.

chungquantin:
@Daanvdplas I am on my way to refactor to avoid the mix of lously couling and tightly coupling. However, `AssetKind` and `AssetCriteria` should not be removed because it does not bring the issue of lously coupling, mainly coming from `type Assets`. The another issue that is a blocker to me at the moment is how to create a shared type of the `pallet_balances::Balance` and `pallet_assets::Balance` because the `BalanceOf<T>` right now only supports the `pallet_asset::Balance`. I came over one post on Substrate stack exchange: https://substrate.stackexchange.com/questions/3604/convert-from-asset-balance-to-currency-balance
```rs
        // Two-way conversion between asset and currency balances
        type AssetToCurrencyBalance: Convert<Self::AssetBalance, BalanceOf<Self>>;
        type CurrencyToAssetBalance: Convert<BalanceOf<Self>, Self::AssetBalance>;
```
Which I think is a feasible way to handle the type conversion but it would create more type parameters to the config.

peterwht:
Please see: https://github.com/r0gue-io/pop-node/issues/97#issuecomment-2266071053

Daanvdplas:
Had a look at it again and want to put this here for later reference. There are 3 ways moving forward:

1. Do nothing and have the devs use the ink native api:
+ positive: _
- negative: devs need to learn two apis to transfer tokens from the chain, potentially fungible docs will have to explain ink related functionality.
2. Creating a separate API, excluding the `AssetId` parameter:
`fn transfer_native(to, amount)`
+ positive: only provides functionality that devs can use, no need for devs to learn other apis than pop provides.
- negative: no value add from the api's ink provides, basically a wrapper, might become deprecated in the future, will only be one maybe two functions.
3. Using the same API with `AssetId == 0`
`fn transfer(id, to, amount)`
+ postive: in the ideal situation this is what we want, a single api, approve functionality for the native token.
- negative: a lot of work, fork of pallet contracts to be able to use assets, store relay token in pallet assets and implement fees to be paid with assets. In addition, we have to throw a lot of errors because half of the functionality provided by fungibles use case can not be used for the dot token.

peterwht:
Would like to close this PR (not delete the branch) as it is very outdated.

PR #150, opened 2024-07-31
feat: burn and mint

### LINKED ISSUE #98: feat (pop api): add burn and mint to local fungibles use case
Based on the `Mintable` and `Burnable` standards.

standards:
- [mint](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#ERC20-_mint-address-uint256-)
- [burn](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20#ERC20Burnable-burn-uint256-)

As you can see they also emit the `Transfer` event

Daanvdplas:
Adds burn and mint to the fungibles use case.

Closes #98

### DIFF
--- a/pallets/api/src/fungibles/tests.rs
+++ b/pallets/api/src/fungibles/tests.rs
@@ -217,7 +217,7 @@ fn token_metadata_works() {
 #[test]
 fn asset_exists_works() {
 	new_test_ext().execute_with(|| {
-		create_asset(ALICE, ASSET, 100);
+		create_asset(ALICE, ASSET);
 		assert_eq!(Assets::asset_exists(ASSET).encode(), Fungibles::read_state(AssetExists(ASSET)));
 	});
 }
--- a/pallets/api/src/fungibles/tests.rs
+++ b/pallets/api/src/fungibles/tests.rs
@@ -108,7 +108,7 @@ fn create_works() {
 #[test]
 fn start_destroy_works() {
 	new_test_ext().execute_with(|| {
-		create_asset(ALICE, ASSET, 100);
+		create_asset(ALICE, ASSET);
 		assert_ok!(Fungibles::start_destroy(signed(ALICE), ASSET));
 	});
 }
@@ -119,7 +119,7 @@ fn set_metadata_works() {
 		let name = vec![42];
 		let symbol = vec![42];
 		let decimals = 42;
-		create_asset(ALICE, ASSET, 100);
+		create_asset(ALICE, ASSET);
 		assert_ok!(Fungibles::set_metadata(
 			signed(ALICE),
 			ASSET,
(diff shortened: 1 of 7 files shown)


pallets/api/src/fungibles/tests.rs
```diff
@@ -147,6 +147,30 @@ fn clear_metadata_works() {
 	});
 }
 
+#[test]
+fn mint_works() {
```
chungquantin, line 151: Should we cover more cases like `Asset` does not exist?

pallets/api/src/fungibles/tests.rs
```diff
@@ -147,6 +147,30 @@ fn clear_metadata_works() {
...
+		create_asset(ALICE, ASSET);
+		let balance_before_mint = Assets::balance(ASSET, &BOB);
+		assert_ok!(Fungibles::mint(signed(ALICE), ASSET, BOB, amount));
+		let balance_after_mint = Assets::balance(ASSET, &BOB);
+		assert_eq!(balance_after_mint, balance_before_mint + amount);
+	});
+}
+
+#[test]
+fn burn_works() {
```
chungquantin, line 163: Should cover more cases like:
- if the value is saturated (like `decrease_allowance`) 
- or if burning more than the total balance of the account.
- check the `total_supply` of the asset wether it is updated.

chungquantin (changes requested):

chungquantin (changes requested):
You missed one line
```diff
#[test]
fn asset_exists_works() {
	new_test_ext().execute_with(|| {
-		create_asset(ALICE, ASSET, 100);
+ create_asset(ALICE, ASSET);
		assert_eq!(Assets::asset_exists(ASSET).encode(), Fungibles::read_state(AssetExists(ASSET)));
	});
}
```

pallets/api/src/fungibles/tests.rs
Daanvdplas: Like I've mentioned before, because it is a 1:1 mapping to the pallet assets dispatchable we should have to test the functionality, only if it works or not. Only when adding logic the test should be expanded (e.g. `approve`).

pallets/api/src/fungibles/tests.rs
Daanvdplas: See above

chungquantin (approved):

pop-api/integration-tests/src/fungibles/utils.rs
```diff
@@ -184,6 +184,32 @@ pub(super) fn clear_metadata(addr: AccountId32, asset_id: AssetId) -> Result<(),
...
+	asset_id: AssetId,
+	account: AccountId32,
+	amount: Balance,
+) -> Result<(), Error> {
+	let function = function_selector("burn");
+	let params = [function, asset_id.encode(), account.encode(), amount.encode()].concat();
+	let result = bare_call(addr, params, 0).expect("should work");
+	decoded::<Result<(), Error>>(result.clone())
+		.unwrap_or_else(|_| panic!("Contract reverted: {:?}", result))
+}
```
peterwht, line 211: Not in this PR, but all of this internal functionality could be refactored to something like:

```rust
fn mint ( ... ) {
    do_bare_call("burn", [...params])?
}
fn burn ( ... ) {
    do_bare_call("burn", [...params])?
}
// ... etc
```

Can we create an issue to handle this refactoring?

peterwht (approved):
Nicely done. Small comment to create an issue to refactor some code in integration tests.

Issue #108, opened 2024-07-17
chore: psp22 example contract


Daanvdplas:
Implement a fungibles contract example implementing the psp22 traits and error provided by the pop api.
- [x] traits implemented and following expected behaviour ([ref](https://docs.openzeppelin.com/contracts/4.x/api/token/erc20))
- [x] `From<StatusCode> for PSP22Error` implementation + tests. For each api there should be one `pallet assets` specific error tested that is converted into `PSP22Error::Custome("<pallet assets error>")`. This is to demonstrate the power of the pop api error.

Using the drink test framework. See internal spec in shortcut.

chungquantin:
Done: https://github.com/r0gue-io/pop-node/pull/300

PR #151, opened 2024-07-31
feat: asset management

### LINKED ISSUE #92: feat (pop api): implement asset management interfaces local fungibles
Implement the functionality regarding Asset Management:
- create
- start_destroy
- destroy_accounts
- destroy_approvals
- finish_destroy
- set_metadata
- clear_metadata

An open question here is whether creating assets should be put in the constructor of a contract using the local fungibles api or not. In other words, no contract will create an asset other than in the contract. I'm not sure about this at the moment and needs to be looked at. As a result of this, example contracts should be made that help developers understand how to manage assets in their contracts using the api (#108).

Include integration-tests.

Daanvdplas:
Implements the functionality for asset management:
- [x] create
- [x] start_destroy
- [x] set_metadata
- [x] clear_metadata
- [x] integration test for creating an asset in the constructor

Additional changes:
- added `DEPRECATED` to old examples and automatic formatting was applied.
- refactored the integration tests helper functions in a separate file `utils.rs`.

The dispatchables required to destroy an asset but not added to the api are `destroy_accounts`, `destroy_approvals` and `finish_destroy`. They have not been included in the api because they do not have to be signed by the owner of the asset and cause complexities in the implementation*. Asset owners (in this case via a contract) are required to complete the steps of destroying an asset via pallet assets in stead of the contract (as it is normally also the case). Only starting the destroying process of an asset has to be initiated by the contract if it is the owner of the asset. If developers want this implemented in the api, we can always do this. Yet now due to the complexity and because destroying assets doesn't happen often we leave it as is. This has to be documented, an issue is created #152. 

*It requires a config type `RemoveItemsLimit` (same as pallet assets) which has to be configured to a smaller amount than in pallet assets. `RemoveItemsLimit` is the max. # of accounts and approvals that can be removed in a single tx. Due to the call being made by a contract this has to be configured to a smaller amount in the api::fungibles pallet. 

Contract examples showing how to to use the `create()`, whether in the constructor or not (or not at all, instantiating a contract which wraps around an existing asset), have been included in #108.

Closes #92

### DIFF
--- a/pop-api/integration-tests/contracts/create_token_in_constructor/lib.rs
+++ b/pop-api/integration-tests/contracts/create_token_in_constructor/lib.rs
@@ -22,8 +22,9 @@ mod create_token_in_constructor {
 		#[ink(constructor, payable)]
 		pub fn new(id: AssetId, min_balance: Balance) -> Result<Self> {
 			let contract = Self { id };
-			let account = contract.env().account_id();
-			api::create(id, account, min_balance)?;
+			// AccountId of the contract which will be set to the owner of the fungible token.
+			let owner = contract.env().account_id();
+			api::create(id, owner, min_balance)?;
 			Ok(contract)
 		}
 
--- a/pop-api/integration-tests/contracts/create_token_in_constructor/lib.rs
+++ b/pop-api/integration-tests/contracts/create_token_in_constructor/lib.rs
@@ -1,0 +1,45 @@
+#![cfg_attr(not(feature = "std"), no_std, no_main)]
+
+use ink::prelude::vec::Vec;
+use pop_api::{
+	assets::fungibles::{self as api},
+	primitives::AssetId,
+	StatusCode,
+};
+
+pub type Result<T> = core::result::Result<T, StatusCode>;
+
+#[ink::contract]
+mod create_token_in_constructor {
+	use super::*;
+
+	#[ink(storage)]
+	pub struct Fungible {
+		id: AssetId,
+	}
+
+	impl Fungible {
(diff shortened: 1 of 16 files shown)


chungquantin (approved):

pop-api/integration-tests/contracts/create_token_in_constructor/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+#[ink::contract]
+mod create_token_in_constructor {
+	use super::*;
+
+	#[ink(storage)]
+	pub struct Fungible {
+		id: AssetId,
+	}
+
+	impl Fungible {
```
peterwht, line 21: Shouldn't this contract also test?
```
 start_destroy
 set_metadata
 clear_metadata
```

peterwht (commented):
Nicely done. Cool to see how clean it was to add in the new functionality.

Two comments:
1. The integration test contract is only testing `create` and asset exists, is this on purpose?
2. The new formatting would have been nice to do in a different PR. Made this one noisier than it needed to be.

pop-api/integration-tests/contracts/create_token_in_constructor/lib.rs
Daanvdplas: The sole reason for this contract and its integration test is for the use case where a contract creates an asset on instantiation, and testing that. Not sure whether a separate integration test is required for this because the `create` functionality has a integration test already, however is this in a contract message.

Setting the metadata and the other apis you mention have to be called via contract messages and are tested in the other contract.

PR #153, opened 2024-07-31
feat: api events

Daanvdplas:
Added events to contract library and pallet, including tests for the pallet.

Note: the weight for increase allowance has changed due to an extra storage read.

The events in the contract library will have to be demonstrated in an example contract. This will have to be done in a different PR.


pallets/api/src/fungibles/mod.rs
```diff
@@ -249,8 +323,15 @@ pub mod pallet {
 			admin: AccountIdOf<T>,
 			min_balance: BalanceOf<T>,
 		) -> DispatchResult {
-			let admin = T::Lookup::unlookup(admin);
-			AssetsOf::<T>::create(origin, id.into(), admin, min_balance)
+			AssetsOf::<T>::create(
+				origin.clone(),
+				id.into(),
```
chungquantin, line 328:
```suggestion
				id.clone().into(),
```

pallets/api/src/fungibles/mod.rs
```diff
@@ -281,7 +362,9 @@ pub mod pallet {
 			symbol: Vec<u8>,
 			decimals: u8,
 		) -> DispatchResult {
-			AssetsOf::<T>::set_metadata(origin, id.into(), name, symbol, decimals)
+			AssetsOf::<T>::set_metadata(origin, id.into(), name.clone(), symbol.clone(), decimals)?;
```
chungquantin, line 365:
```suggestion
			AssetsOf::<T>::set_metadata(origin, id.clone().into(), name.clone(), symbol.clone(), decimals)?;
```

chungquantin (changes requested):

pallets/api/src/fungibles/mod.rs
```diff
@@ -93,6 +95,57 @@ pub mod pallet {
...
+			id: AssetIdOf<T>,
+			/// Account providing allowance.
+			owner: AccountIdOf<T>,
+			/// Allowance beneficiary.
+			spender: AccountIdOf<T>,
+			/// New allowance amount.
+			value: BalanceOf<T>,
+		},
+		/// Event emitted when transfer of tokens occurs.
+		Transfer {
```
peterwht, line 113: `Approval` and `Transfer` are the only events per the PSP22.

pallets/api/src/fungibles/mod.rs
```diff
@@ -93,6 +95,57 @@ pub mod pallet {
...
+			id: AssetIdOf<T>,
+			/// Transfer sender. `None` in case of minting new tokens.
+			from: Option<AccountIdOf<T>>,
+			/// Transfer recipient. `None` in case of burning tokens.
+			to: Option<AccountIdOf<T>>,
+			/// Amount of tokens transferred (or minted/burned).
+			value: BalanceOf<T>,
+		},
+		/// Event emitted when a token is created.
+		Create {
```
peterwht, line 124: This does not seem useful to me. It is simply emitting the same event that pallet-assets is already emitting

pallets/api/src/fungibles/mod.rs
```diff
@@ -93,6 +95,57 @@ pub mod pallet {
...
+			/// Token ID.
+			id: AssetIdOf<T>,
+			/// Owner of the token created.
+			owner: AccountIdOf<T>,
+			/// Admin of the token created.
+			admin: AccountIdOf<T>,
+		},
+		/// Event emitted when a token is in the process of being destroyed.
+		StartDestroy { id: AssetIdOf<T> },
```
peterwht, line 133: likewise

pallets/api/src/fungibles/mod.rs
```diff
@@ -93,6 +95,57 @@ pub mod pallet {
...
+		/// Event emitted when new metadata is set for a token.
+		SetMetadata {
```
peterwht, line 135: likewise

pallets/api/src/fungibles/mod.rs
```diff
@@ -93,6 +95,57 @@ pub mod pallet {
...
+			id: AssetIdOf<T>,
+			/// Token name.
+			name: Vec<u8>,
+			/// Token symbol.
+			symbol: Vec<u8>,
+			/// Token decimals.
+			decimals: u8,
+		},
+		/// Event emitted when metadata is cleared for a token.
+		ClearMetadata { id: AssetIdOf<T> },
```
peterwht, line 146: likewise

peterwht (changes requested):
Unfortunately, I think this requires some additional thought.

Currently, the events are duplicates to the ones emitted from `pallet-assets`. From a contract-developer's point of view, why is that useful to them? Either way, they will have to perform the same indexing of events, the only difference is if they choose an event from `pallet-assets` or from Pop's api pallet.

I think realistically, these events should abide by the PSP22 standard, and only those should be returned.

We should think on what is actually valuable to the contract developers, and to us.

One idea:
- A contract developer may want to keep track of how their contract is used. So, we probably want to know when an API endpoint is called, which contract called it, and which use-case. The particular parameters may not be relevant as the underlying pallet can be indexed as well.

Rather, simply emitting events like: "pallet-api.psp22.create {contract_addr}` gives us and the contract developer knowledge of often their contract calls into the runtime.

Daanvdplas:
> One idea:
A contract developer may want to keep track of how their contract is used. So, we probably want to know when an API endpoint is called, which contract called it, and which use-case. The particular parameters may not be relevant as the underlying pallet can be indexed as well.

Please create an issue for it as I'm not completely following you.

Daanvdplas:
I agree that the events in the api pallet are perhaps unnecessary. However, in the contract library it might still be useful. We provide the events that developers "can" use when they are calling the apis for asset management. This immediately also provides some form of standard.

peterwht:
> I agree that the events in the api pallet are perhaps unnecessary. However, in the contract library it might still be useful. We provide the events that developers "can" use when they are calling the apis for asset management. This immediately also provides some form of standard.

Yes, I agree. The ones in the contract are useful and should be kept!

pallets/api/src/fungibles/mod.rs
Daanvdplas: I agree with you. Removed those events from the pallet.

Daanvdplas: see refactor: remove duplicated events

pallets/api/src/fungibles/mod.rs
```diff
@@ -93,6 +95,33 @@ pub mod pallet {
 	#[pallet::pallet]
 	pub struct Pallet<T>(_);
 
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
```
evilrobot-01, line 100: Missing doc comments on the the event type itself.

pallets/api/src/fungibles/mod.rs
```diff
@@ -150,11 +192,11 @@ pub mod pallet {
...
 			value: BalanceOf<T>,
 		) -> DispatchResultWithPostInfo {
-			let who = ensure_signed(origin.clone())
+			let owner = ensure_signed(origin.clone())
 				.map_err(|e| e.with_weight(Self::weight_approve(0, 0)))?;
-			let current_allowance = AssetsOf::<T>::allowance(id.clone(), &who, &spender);
-			let spender = T::Lookup::unlookup(spender);
-			let id: AssetIdParameterOf<T> = id.into();
+			let current_allowance = AssetsOf::<T>::allowance(id.clone(), &owner, &spender);
+			let spender_unlookup = T::Lookup::unlookup(spender.clone());
```
evilrobot-01, line 198: `unlookup` returns a `Source` -> `spender_source` would be a better name.

pallets/api/src/fungibles/mod.rs
```diff
@@ -150,11 +192,11 @@ pub mod pallet {
...
+			let id_param: AssetIdParameterOf<T> = id.clone().into();
```
evilrobot-01, line 199: `id_param` is not used when `Equal`, better to move it into the `Greater`/`Less` match arms to avoid the unnecessary clone when equal. Very low likelihood of the code path being run but an easy optimisation.

In the Greater case, it can be inserted directly inline into line 209 without the need for the variable.

pallets/api/src/fungibles/mod.rs
```diff
@@ -193,15 +241,24 @@ pub mod pallet {
 		/// - `spender` - The account that is allowed to spend the tokens.
 		/// - `value` - The number of tokens to increase the allowance by.
 		#[pallet::call_index(6)]
-		#[pallet::weight(AssetsWeightInfoOf::<T>::approve_transfer())]
+		#[pallet::weight(AssetsWeightInfoOf::<T>::approve_transfer() + T::DbWeight::get().reads(1))]
```
evilrobot-01, line 244: Missing proof size in this simple read addition? Sadly should have a benchmark.

pallets/api/src/fungibles/mod.rs
```diff
@@ -218,24 +275,31 @@ pub mod pallet {
...
 			value: BalanceOf<T>,
 		) -> DispatchResultWithPostInfo {
-			let who = ensure_signed(origin.clone())
+			let owner = ensure_signed(origin.clone())
 				.map_err(|e| e.with_weight(Self::weight_approve(0, 0)))?;
-			let current_allowance = AssetsOf::<T>::allowance(id.clone(), &who, &spender);
-			let spender = T::Lookup::unlookup(spender);
-			let id: AssetIdParameterOf<T> = id.into();
+			let current_allowance = AssetsOf::<T>::allowance(id.clone(), &owner, &spender);
+			let spender_unlookup = T::Lookup::unlookup(spender.clone());
```
evilrobot-01, line 281: Naming as above.

pallets/api/src/fungibles/tests.rs
```diff
@@ -9,165 +9,212 @@ use frame_support::{
...
+		let id = ASSET;
+		let from = Some(ALICE);
+		let to = Some(BOB);
+
+		create_asset_and_mint_to(ALICE, id, ALICE, value * 2);
+		let balance_before_transfer = Assets::balance(id, &BOB);
+		assert_ok!(Fungibles::transfer(signed(ALICE), id, BOB, value));
+		let balance_after_transfer = Assets::balance(id, &BOB);
+		assert_eq!(balance_after_transfer, balance_before_transfer + value);
+		System::assert_last_event(Event::Transfer { id, from, to, value }.into());
```
evilrobot-01, line 27: Minor, but could just import the `System::assert_last_event` function to avoid the `System::` prefix in all the tests.

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -58,6 +58,65 @@ mod constants {
 	pub(super) const BURN: u8 = 20;
 }
 
+pub mod events {
```
evilrobot-01, line 61: Missing doc comment for the public module. What is the point of this module? Some documentation will help users know why these events exist.

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -58,6 +58,65 @@ mod constants {
...
+		/// Transfer recipient. `None` in case of burning tokens.
+		#[ink(topic)]
+		pub to: Option<AccountId>,
+		/// Amount of tokens transferred (or minted/burned).
+		pub value: u128,
+	}
+
+	/// Event emitted when a token class is created.
+	#[ink::event]
+	pub struct Create {
```
evilrobot-01, line 92: This and remaining events are missing topics.

pop-api/src/v0/assets/fungibles.rs
evilrobot-01: I also think that the usage of these events should be demonstrated somewhere in this PR - e.g. unit or integration tests.

evilrobot-01 (changes requested):
Looking great. There are some missing doc comments and some minor optimisations which can be done easily.

Simple weight addition is questionable and proof that the contract events work by inclusion in some tests would be ideal.

pallets/api/src/fungibles/tests.rs
Daanvdplas: Because it is an associated function of the Pallet struct, rather than a free-standing function at the crate level, you cannot directly import it using frame_system::assert_last_event. Instead, it must be accessed through the Pallet struct.

The mock::System::assert_last_event import path assumes that System (which is an alias for frame_system::Pallet) has been defined in a mock module and that the assert_last_event function is accessible directly from this alias. However, the assert_last_event function is tied to the Pallet implementation for the frame_system module and is not a standalone function that can be imported directly like this.

pallets/api/src/fungibles/mod.rs
Daanvdplas: See commit `refactor: incompatible event based on standard`

pallets/api/src/fungibles/mod.rs
Daanvdplas: see commit `refactor: optimization pallet fungibles`

pallets/api/src/fungibles/mod.rs
Daanvdplas: see commit `refactor: optimization pallet fungibles`

pallets/api/src/fungibles/mod.rs
Daanvdplas: see commit `refactor: optimization pallet fungibles`

pop-api/src/v0/assets/fungibles.rs
Daanvdplas: See commit `refactor: ink topics + docs to contract events`

pallets/api/src/fungibles/mod.rs
Daanvdplas: Please see commit `fix: weight increase allowance`. This is a simple solution and not 100% accurate however before I put more time into this I would like to get your view on the approach that has been taken for `decrease_allowance`. Basically, we are reusing the weight function of `approve` for `increase_` and `decrease_allowance`. This is for all three we have to read state (`allowance`) and either call `approve_transfer`, `cancel_approval`, or both. As I think this is fine for now since we never underestimate weight, this does deserve attention before going to mainnet.

pop-api/src/v0/assets/fungibles.rs
Daanvdplas: This will take me 2 hours. Will push something later tonight

chungquantin (approved):

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -58,6 +58,83 @@ mod constants {
 	pub(super) const BURN: u8 = 20;
 }
 
+pub mod events {
```
chungquantin, line 61: Reviewed! The PR looks good to me. But I think we should have a documentation for this somewhere in the code (or in a guidebook) to mention that using these events in the contract would make the contract size larger. And if not needed, the developers should make use of the runtime event that we already exposed by default.

chungquantin (approved):

pop-api/src/v0/assets/fungibles.rs
chungquantin: Link for reference: https://use.ink/basics/events#cost-of-using-events

pop-api/src/v0/assets/fungibles.rs
Daanvdplas: EDIT: will do this tomorrow, created a separate issue for it to not delay this getting merged #181

Daanvdplas: As for docs, see commit `docs: events contract`

pop-api/src/v0/assets/fungibles.rs
Daanvdplas: docs: events contract

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -58,6 +58,15 @@ mod constants {
 	pub(super) const BURN: u8 = 20;
 }
 
+/// A set of events for use in smart contracts interacting with the fungibles API.
```
evilrobot-01, line 61: Great docs!

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -90,30 +99,48 @@ pub mod events {
...
+		#[ink(topic)]
+		pub id: AssetId,
+		#[ink(topic)]
+		/// The name of the asset.
+		pub name: Vec<u8>,
+		/// The symbol of the asset.
+		#[ink(topic)]
+		pub symbol: Vec<u8>,
+		/// The decimals of the asset.
+		#[ink(topic)]
```
evilrobot-01, line 134: I think this is overkill, why would someone want the event indexed by the number of decimals?

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -90,30 +99,48 @@ pub mod events {
...
 	#[ink::event]
 	pub struct SetMetadata {
-		id: AssetId,
-		name: Vec<u8>,
-		symbol: Vec<u8>,
-		decimals: u8,
+		/// The ID of the asset created.
+		#[ink(topic)]
+		pub id: AssetId,
+		#[ink(topic)]
```
evilrobot-01, line 127: Debatable, but probably okay.

pallets/api/src/fungibles/mod.rs
```diff
@@ -95,13 +95,12 @@ pub mod pallet {
 	#[pallet::pallet]
 	pub struct Pallet<T>(_);
 
+	/// The events that can be emitted.
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(super) fn deposit_event)]
 	pub enum Event<T: Config> {
 		/// Event emitted when allowance by `owner` to `spender` changes.
 		Approval {
-			/// Token ID.
```
evilrobot-01: Sadly I think this is problematic. I understand it was probably removed to conform to the standard, but does ommitting it effectively make it useless? The events are emitted from a pallet managing multiple assets, so without knowing which asset id had an approval, is there any point? A contract call might be able to correlate the resulting events based on the extrinsic events, but if contract makes two calls to the fungibles pallet to create an allowance, for two different assets, how can we tell from the events which value matches which asset?

evilrobot-01 (commented):
Hmmm, not sure about the id removal from the events. See inline comments for more context.

evilrobot-01:
A general heads up that all these PRs to the `api` branch do not seem to include any test coverage stats, meaning that there might still be some work to do before the branch is merged to `main`. I will take a look at that branch first to see why.

evilrobot-01:
> A general heads up that all these PRs to the `api` branch do not seem to include any test coverage stats, meaning that there might still be some work to do before the branch is merged to `main`. I will take a look at that branch first to see why.

FYI - the `daan/api` branch is missing a number of commits from main, with the ci one perhaps the most important. I suggest getting this addressed soon so that you can get a feel for how much code is untested before assuming that the initial api work is done.

image

pallets/api/src/fungibles/mod.rs
Daanvdplas: But perhaps this brings us back to this [comment](https://github.com/r0gue-io/pop-node/pull/153#pullrequestreview-2229006536). 

In short, reason for adding events to the fungible pallet:
- standard compliance - the problem is that the events are not useful though (see frank's comment). In addition, when we do add the `asset_id` the events become the same as in pallet assets (e.g. [Transferred](https://github.com/paritytech/polkadot-sdk/blob/819a5818284a96a5a5bd65ce67e69bab860d4534/substrate/frame/assets/src/lib.rs#L563))
- pallet usage - we have an indicator how often the pallet is used. More interesting is how often the API is used

Would it be oke to start without events in the pallet?

Daanvdplas:
> FYI - the daan/api branch is missing a number of commits from main, with the ci one perhaps the most important. I suggest getting this addressed soon so that you can get a feel for how much code is untested before assuming that the initial api work is done.

On it right away

pop-api/src/v0/assets/fungibles.rs
Daanvdplas: Didn't think here, thanks for pointing it out

pop-api/src/v0/assets/fungibles.rs
Daanvdplas: Left it in "A good rule of thumb is to ask yourself if somebody might want to search for this topic." 

Could be.

pallets/api/src/fungibles/mod.rs
peterwht: > Would it be oke to start without events in the pallet?

I would rather we start with something, even super simple events so we can be aware of if the API is being used on testnet.

Daanvdplas: Alright I will then add the `asset id` back in the `Transfer` and `Approval` event

pallets/api/src/fungibles/mod.rs
```diff
@@ -252,8 +318,7 @@ pub mod pallet {
 			admin: AccountIdOf<T>,
 			min_balance: BalanceOf<T>,
 		) -> DispatchResult {
-			let admin = T::Lookup::unlookup(admin);
-			AssetsOf::<T>::create(origin, id.into(), admin, min_balance)
+			AssetsOf::<T>::create(origin, id.into(), T::Lookup::unlookup(admin), min_balance)
```
peterwht, line 321: Do we want to know if somebody creates an asset through Pop API? If yes, can please add a simple event here. Otherwise I am fine if you leave it as is.

peterwht (approved):
LGTM! Left a small comment, good with whatever you decide.

pallets/api/src/fungibles/mod.rs
```diff
@@ -218,24 +277,31 @@ pub mod pallet {
...
+			let owner = ensure_signed(origin.clone())
 				.map_err(|e| e.with_weight(Self::weight_approve(0, 0)))?;
-			let current_allowance = AssetsOf::<T>::allowance(id.clone(), &who, &spender);
-			let spender = T::Lookup::unlookup(spender);
-			let id: AssetIdParameterOf<T> = id.into();
+			let current_allowance = AssetsOf::<T>::allowance(id.clone(), &owner, &spender);
+			let spender_source = T::Lookup::unlookup(spender.clone());
+			let id_param: AssetIdParameterOf<T> = id.clone().into();
 
 			if value.is_zero() {
```
evilrobot-01, line 286: Why not have this check directly after the origin check? There are unnecessary reads which occur which dont factor into this check, so more efficient to return as early as possible.

pallets/api/src/fungibles/mod.rs
```diff
@@ -150,11 +193,9 @@ pub mod pallet {
...
 		) -> DispatchResultWithPostInfo {
-			let who = ensure_signed(origin.clone())
+			let owner = ensure_signed(origin.clone())
 				.map_err(|e| e.with_weight(Self::weight_approve(0, 0)))?;
-			let current_allowance = AssetsOf::<T>::allowance(id.clone(), &who, &spender);
-			let spender = T::Lookup::unlookup(spender);
-			let id: AssetIdParameterOf<T> = id.into();
+			let current_allowance = AssetsOf::<T>::allowance(id.clone(), &owner, &spender);
 
 			let return_weight = match value.cmp(&current_allowance) {
```
evilrobot-01, line 200: This variable is called `weight` in `decrease_allowance` - suggest using the same name for consistency.

pallets/api/src/fungibles/mod.rs
evilrobot-01: Good point.

pallets/api/src/fungibles/tests.rs
```diff
@@ -9,165 +9,212 @@ use frame_support::{
...
-		create_asset_and_set_metadata(ALICE, ASSET, name, symbol, decimals);
-		assert_ok!(Fungibles::clear_metadata(signed(ALICE), ASSET));
-		assert_eq!(Assets::name(ASSET), Vec::<u8>::new());
-		assert_eq!(Assets::symbol(ASSET), Vec::<u8>::new());
-		assert_eq!(Assets::decimals(ASSET), 0u8);
+		let id = ASSET;
+
+		create_asset_and_set_metadata(ALICE, id, vec![42], vec![42], 42);
+		assert_ok!(Fungibles::clear_metadata(signed(ALICE), id));
+		assert_eq!(Assets::name(id), Vec::<u8>::new());
```
evilrobot-01, line 181: For future, could probably just check if it is empty - less verbose.

evilrobot-01 (changes requested):
Just one more optimisation and a variable rename for consistency. Peter has a good point in terms of emitting events to determine usage, so hopefully a quick add.

peterwht:
[1124]

pallets/api/src/fungibles/mod.rs
Daanvdplas: see [refactor: fungible pallet](https://github.com/r0gue-io/pop-node/pull/153/commits/3b3ae9fcf4de6115bc5a2ecb9a92a2ca2e0d0068)

pallets/api/src/fungibles/tests.rs
Daanvdplas: I like it, thanks for pointing it out!

see [refactor: fungible pallet](https://github.com/r0gue-io/pop-node/pull/153/commits/3b3ae9fcf4de6115bc5a2ecb9a92a2ca2e0d0068)

pallets/api/src/fungibles/mod.rs
Daanvdplas: Yes good one

[refactor: fungible pallet](https://github.com/r0gue-io/pop-node/pull/153/commits/3b3ae9fcf4de6115bc5a2ecb9a92a2ca2e0d0068)

pallets/api/src/fungibles/mod.rs
```diff
@@ -318,7 +327,15 @@ pub mod pallet {
...
 			min_balance: BalanceOf<T>,
 		) -> DispatchResult {
-			AssetsOf::<T>::create(origin, id.into(), T::Lookup::unlookup(admin), min_balance)
+			AssetsOf::<T>::create(
+				origin.clone(),
+				id.clone().into(),
+				T::Lookup::unlookup(admin.clone()),
+				min_balance,
+			)?;
+			let creator = ensure_signed(origin)?;
```
evilrobot-01, line 336: I'd stick this as the first statement for consistency.

evilrobot-01 (commented):
One last minor refactor then good for me.

pallets/api/src/fungibles/mod.rs
Daanvdplas: see last commit

pallets/api/src/fungibles/mod.rs
evilrobot-01: Strictly speaking, these are just dispachables and therefore callable from anywhere. There is no guarantee that the call actually came from a contract, but may be a high probability.

From an indexing perspective, if we can correlate a contract Called event with Created event with same transaction then we're fine. One would assume so.

evilrobot-01 (approved):

PR #156, opened 2024-08-04
chore: upgrade to 1.8.0

chungquantin:
- ISSUE: https://github.com/r0gue-io/pop-node/issues/141
- Release: https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.8.0
- Prerequisites: https://github.com/r0gue-io/pop-node/pull/139
## CHANGELOG
- [x] Make the `benchmark pallet` command only require a `Hasher`: https://github.com/paritytech/polkadot-sdk/pull/3244
- [x] Remove `slot_duration` from `AuraParams` (https://github.com/paritytech/polkadot-sdk/commit/ec6bf5d0a52161bda0eccd2d797dffab4213900a). Slot 
- [x] Update `jsonrpsee` version to `1.8.0`
- [x] Add `ApiVersion` to `pallet_contracts::Config` 

### [3384](https://github.com/paritytech/polkadot-sdk/pull/3384): `pallet_contracts` stabilize call_v2, instantiate_v2, lock_dependency and unlock_dependency. 

These APIs are currently unstable and are being stabilized in this PR. 
Note: `add_delegate_dependency` and `remove_delegate_dependency` have been renamed to `lock_dependency` and `unlock_dependency` respectively.

### [3243](https://github.com/paritytech/polkadot-sdk/pull/3243): Don't fail fast if the weight limit of a cross contract call is too big
Cross contracts calls will now be executed even if the supplied weight limit is bigger than the reamining weight. If the actual weight is too low they will fail in the cross contract call and roll back. This is different from the old behaviour where the limit for the cross contract call must be smaller than the remaining weight.

### [3154](https://github.com/paritytech/polkadot-sdk/pull/3154): Contracts: Stabilize caller_is_root API
Removed the #[unstable] attrribute on caller_is_root host function.

### [3184](https://github.com/paritytech/polkadot-sdk/pull/3184): Contracts: Remove no longer enforced limits from the Schedule
The limits are no longer in use and do nothing. Every builder overwritting them can just adapt their code to remove them without any consequence.

### [3415](https://github.com/paritytech/polkadot-sdk/pull/3415): [pallet-contracts] Add APIVersion to the config.
Add APIVersion to the config to communicate the state of the Host functions exposed by the pallet.


evilrobot-01:
Awesome! I especially like the changelog. 🔥

Shall we perhaps include a subsection in the changelog specifically around contracts, with links to anything that is relevant to contracts in the release notes/prdocs? This is simply to save any reviewers some time, as the person doing the upgrade would have covered the content already, it could be greatly helpful to collate them into a section in the PR description here. 

There are a lot of improvements in the contracts pallet in coming releases so would be ideal that we are all familiar with them. and their effects.

chungquantin:
@evilrobot-01 I have updated the PR to cover contracts related changes. However, seems like this upgrade made changes to the XCM integration tests, [failed with different events returned](https://github.com/r0gue-io/pop-node/actions/runs/10233460935/job/28311783524?pr=156#:~:text=RococoRelay%3A%3ARuntimeEvent,1777). I am taking a look to find commits that leads to this.

evilrobot-01:
I think this is due to the change from currency adapter to fungibles adapter.

See my other open PR which updates the integration tests to use Paseo rather than Rococo.

peterwht:
Just commenting as a useful issue for this PR: https://github.com/moondance-labs/tanssi/issues/574

peterwht (approved):
lgtm. Just needs the resolved tests from Frank's PR.

fyi for other reviewers:
- did a diff of parachain template service.rs, no differences relevant to this upgrade.

evilrobot-01:
Expect there will be compilation issues here when the merge conflicts are resolved.

I pointed out that the integration tests PR switching to Paseo runtimes meant we would be stuck on SDK 1.7, as that is what Paseo is on. Emulated tests compile into a single build artifact and therefore need the same versions. 

Only other option is using the Polkadot runtime instead, as Paseo is based on it, but that would require using a specific commit there and a jump to 1.13. I see they are now on 15, at least on main. Last release is still pre 1.13.

evilrobot-01:
My suggestion would be to first get async backing enable on the testnet, then upgrade to 1.13 before updating collators and updating runtime. These can be done in parallel.

A decision would also need to be made on how to handle the update. One PR jumping from current to .13 or several smaller PRs to branch from preceding PR to incrementally update. The latter means more smaller reviews vs one large.

Finally, the fellowship runtimes are on .15 so it could be a case of going as far as we can in one big push and then we are current and can focus elsewhere.

peterwht:
> My suggestion would be to first get async backing enable on the testnet, then upgrade to 1.13 before updating collators and updating runtime. These can be done in parallel.
> 
> A decision would also need to be made on how to handle the update. One PR jumping from current to .13 or several smaller PRs to branch from preceding PR to incrementally update. The latter means more smaller reviews vs one large.

I would prefer to move to an OmniNode, then do a large jump to 1.15. Reason being that runtime changes are usually easy to review, while node changes can be a little more tedious. The current polkadot-parachain binary is async backing enabled with lookahead.

We can use Tanssi's tool to generate release changes from 1.7 -> 1.15 so it's easy to see.

Daanvdplas:
As talked about internally, lets create a `developer` branch and create all the upcoming runtime upgrade PRs to the `developer` branch. We are current with 1.14, we merge it as a whole to `main`.

Daanvdplas:
This is done correct?

chungquantin:
@Daanvdplas Should be closed because I merged it to `developer` branch already.

Daanvdplas:
I thought so

Just checking to be sure :)

PR #163, opened 2024-08-07
refactor: generic extension crate to de-duplicate code between runtimes

chungquantin:
### Description
Removing the extension from the `devnet` and `testnet` runtimes. Make sure that adding features in the future for one runtime does not impact the other runtimes. 
### CHANGELOG
- Separate the extension from runtimes to standalone crate `pop-runtime-extension` in `runtime/extension` 
- Preview the integration of `pop-runtime-extension` crate to `pop-runtime-testnet`: https://github.com/r0gue-io/pop-node/pull/167
- Encoding / decoding dispatch error test module is currently defined inside `pop-runtime-devnet`. This can be improved in the another PR to make something like `TestSuiteBuilder` that can be reused in multiple runtimes. 
### How feature can be added in the future? 
Create a struct `Extension` and implement the `ReadState` trait and `CallFilter` trait for the `Extension` to handle the queries from the contract environment.


chungquantin:
Got several blocks trying to separate the chain extension from the runtimes. The main issue is the required traits. Let say if `Devnet` implements fungible feature, it requires `pallet_asset::Config` but `Testnet` (and probably `Prod`) won't. My first attempt is do mutual exclusiveness, having two `PopApiExtensionConfig` trait that is flipped based on the feature. But feature unification will mess it up so the approach does not work.  
```rs
// Conditionally use different implementations of `PopApiExtensionConfig` based on the active feature.
cfg_if::cfg_if! {
	if #[cfg(feature = "pop-devnet")] {
		pub use config::devnet::PopApiExtensionConfig;
	} else if #[cfg(feature = "pop-testnet")] {
		pub use config::testnet::PopApiExtensionConfig;
	}
}
```
My second attempt is to expose two different traits `devnet::PopApiExtensionConfig` and `testnet::PopApiExtensionConfig` which both requires `PopApiExtensionConfigBase`. However, this ends up with another blocker for `impl has stricter requirements`. Reason for that because `StateReadHandler` and `CallDispatchHandler` has `PopApiExtensionConfigBase` as a trait bound but we provide `devnet::PopApiExtensionConfig` which implements `pallet_asset::Config` (more requirements than the one asked by the methods)

peterwht (commented):
Nice! Left a few nitpick comments, then I think it's good to go.

runtime/devnet/src/config/api.rs
```diff
@@ -1,14 +1,23 @@
 use crate::{config::assets::TrustBackedAssetsInstance, fungibles, Runtime, RuntimeCall};
 use codec::{Decode, Encode, MaxEncodedLen};
 use frame_support::traits::Contains;
+use pop_runtime_extension::ReadStateHandler;
 
 /// A query of runtime state.
 #[derive(Encode, Decode, Debug, MaxEncodedLen)]
 #[repr(u8)]
-pub enum RuntimeRead<T: fungibles::Config> {
+pub enum RuntimeRead {
```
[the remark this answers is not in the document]
chungquantin: This must be public because RuntimeRead is passed to the ApiExtensionConfig.

runtime/devnet/src/config/contracts.rs
```diff
@@ -1,13 +1,15 @@
...
-	deposit, extensions, Balance, Balances, BalancesCall, Perbill, Runtime, RuntimeCall,
-	RuntimeEvent, RuntimeHoldReason, Timestamp,
+	deposit, Balance, Balances, BalancesCall, Perbill, Runtime, RuntimeCall, RuntimeEvent,
+	RuntimeHoldReason, Timestamp,
 };
 use frame_support::{
 	parameter_types,
 	traits::{ConstBool, ConstU32, Randomness},
 };
 use frame_system::pallet_prelude::BlockNumberFor;
```
Daanvdplas, line 10: remove whitespace

runtime/devnet/src/config/extension.rs
```diff
@@ -0,0 +1,80 @@
+#[cfg(test)]
```
Daanvdplas, line 1: This test should be in the chain extension crate. We can remove the use of `Runtime`. This was only a test for clarity but lets keep what we can keep.

extension/src/lib.rs
```diff
@@ -1,58 +1,72 @@
...
+	/// Layout of `params`: [version, pallet_index, call_index, ...Vec<u8>].
+	fn handle_read(params: T::RuntimeRead) -> Vec<u8>;
+}
+
 #[derive(Default)]
-pub struct PopApiExtension;
+pub struct ApiExtension;
+
+/// Extract (version, function_id, pallet_index, call_index) from the payload bytes.
+fn extract_env<T, E>(env: &Environment<E, BufInBufOutState>) -> (u8, u8, u8, u8)
```
Daanvdplas, line 43: Would be good if we can use this function in `test_byte_extraction`

Daanvdplas (commented):
This is really good work Tin! You got a super clean refactor here. Before we merge I'd like to get @evilrobot-01 review as well.

extension/src/lib.rs
```diff
@@ -1,58 +1,72 @@
...
-const UNKNOWN_CALL_ERROR_ENCODED: [u8; 4] = [254u8, 0, 0, 0];
-
 type ContractSchedule<T> = <T as pallet_contracts::Config>::Schedule;
 
+pub trait Config:
+	frame_system::Config<RuntimeCall: GetDispatchInfo + Dispatchable<PostInfo = PostDispatchInfo>>
+{
+	/// A query of runtime state.
+	type RuntimeRead: Decode + ReadStateHandler<Self>;
+	/// Whitelisting list of runtime calls and read state calls.
```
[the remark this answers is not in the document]
evilrobot-01: Can we use Allowlisted instead please.

Daanvdplas (approved):

extension/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+
+/// Function identifiers used in the Pop API.
+///
+/// The `FuncId` specifies the available functions that can be called through the Pop API. Each
+/// variant corresponds to a specific functionality provided by the API, facilitating the
+/// interaction between smart contracts and the runtime.
+///
+/// - `Dispatch`: Represents a function call to dispatch a runtime call.
+/// - `ReadState`: Represents a function call to read the state from the runtime.
+/// - `SendXcm`: Represents a function call to send an XCM message.
```
Daanvdplas, line 274: Can you remove this please. Old code I forgot to remove

extension/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+	/// Processes the parameters needed to execute a call within the runtime environment.
+	/// Layout of `params`: [version, pallet_index, call_index, ...Vec<u8>].
+	fn handle_read(params: T::RuntimeRead) -> Vec<u8>;
+}
+
+#[derive(Default)]
+pub struct ApiExtension;
+
+/// Extract (version, function_id, pallet_index, call_index) from the payload bytes.
+fn extract_env<T, E: Ext<T = T>>(env: &Environment<E, BufInBufOutState>) -> (u8, u8, u8, u8) {
```
Daanvdplas, line 45: Could we move this below `call()`

extension/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+
+		match result {
+			Ok(_) => Ok(RetVal::Converging(0)),
+			Err(e) => Ok(RetVal::Converging(convert_to_status_code(e, version))),
+		}
+	}
+}
+
+/// Helper method to decode the byte data to a provided type and throws error if failed.
+fn decode_checked<T: Decode>(params: &mut &[u8]) -> Result<T, DispatchError> {
```
Daanvdplas, line 108: Could we move this below `dispatch()`

extension/src/lib.rs
```diff
@@ -0,0 +1,298 @@
...
+
+/// Wrapper to enable versioning of runtime calls.
+#[derive(Decode, Debug)]
+enum VersionedDispatch<T: Config> {
+	/// Version zero of dispatch calls.
+	#[codec(index = 0)]
+	V0(T::RuntimeCall),
+}
+
+// Converts a `DispatchError` to a `u32` status code based on the version of the API the contract uses.
```
Daanvdplas, line 218: Could we move all this code and doc below the TryFrom implementation so that the types `Versioned...` and `FuncId` are nicely positioned together.

Daanvdplas (commented):
Sorry some reordering of the extension file

extension/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+///
+/// The `FuncId` specifies the available functions that can be called through the Pop API. Each
+/// variant corresponds to a specific functionality provided by the API, facilitating the
+/// interaction between smart contracts and the runtime.
+///
+/// - `Dispatch`: Represents a function call to dispatch a runtime call.
+/// - `ReadState`: Represents a function call to read the state from the runtime.
+#[derive(Debug)]
+pub enum FuncId {
+	Dispatch,
```
evilrobot-01, line 228: No doc comments for variants.

runtime/devnet/src/config/api.rs
```diff
@@ -1,14 +1,23 @@
...
 	/// Fungible token queries.
 	#[codec(index = 150)]
-	Fungibles(fungibles::Read<T>),
+	Fungibles(fungibles::Read<Runtime>),
+}
+
+impl ReadStateHandler<Runtime> for RuntimeRead {
+	fn handle_read(read: RuntimeRead) -> sp_std::vec::Vec<u8> {
```
evilrobot-01, line 16: Unnecessary path prefix

runtime/devnet/src/config/api.rs
```diff
@@ -1,14 +1,23 @@
...
 #[repr(u8)]
-pub enum RuntimeRead<T: fungibles::Config> {
+pub enum RuntimeRead {
 	/// Fungible token queries.
 	#[codec(index = 150)]
-	Fungibles(fungibles::Read<T>),
+	Fungibles(fungibles::Read<Runtime>),
+}
+
+impl ReadStateHandler<Runtime> for RuntimeRead {
```
evilrobot-01, line 15: Implementing this trait for the enum seems weird. It is the runtime that is deciding to match an enum variant to a pallet function, so it might feel more natural to implement the trait for the Runtime or for a local Extension struct.

An example that comes to mind is the XcmExecutor struct which is used with pallet-xcm. If we consider the Extension as a subsystem of pallet-contracts, would modelling it in a similar way make it easier. That is, we provide a generic Extension which the runtime wraps in its own Extension struct. Any requirements are then implemented for the Extension struct rather than the current Config trait you have implemented for the Runtime.

runtime/devnet/src/config/contracts.rs
```diff
@@ -44,6 +45,11 @@ parameter_types! {
 	pub const CodeHashLockupDepositPercent: Perbill = Perbill::from_percent(0);
 }
 
+impl pop_chain_extension::Config for Runtime {
```
evilrobot-01, line 48: As mentioned elsewhere, while I like the idea I think the implementation is a little confusing as we have all these dependent traits littered about which may make maintenance more difficult.

A rough suggestion might be to have an Extension or ChainExtension type defined and the relevant traits required implemented for that type only, so it then just effectively 'plugs in' to the ChainExtension on line 72.

I'll try find some time later today to try and mock something out.

extension/src/v0.rs
```diff
@@ -1,5 +1,5 @@
 #[cfg(test)]
```
evilrobot-01, line 1: The introduction of versioning to this crate seems off, if the intended goal was a generic chain extension and this seems very specific. I guess part of this is to eliminate duplicate code, but from a modularity perspective I wonder if versioning should be a 'layer above'. 

I am probably missing the nuances here, so will take some time to consider it further...

extension/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+	frame_system::Config<RuntimeCall: GetDispatchInfo + Dispatchable<PostInfo = PostDispatchInfo>>
+{
+	/// A query of runtime state.
+	type RuntimeRead: Decode + ReadStateHandler<Self>;
+	/// Allowlisted runtime calls and read state calls.
+	type AllowedApiCalls: Contains<Self::RuntimeCall> + Contains<Self::RuntimeRead>;
+}
+
+/// Trait for handling parameters from the chain extension environment during state read operations.
+pub trait ReadStateHandler<T: Config> {
```
evilrobot-01, line 35: Subjective, but the `Handler` suffix doesnt seem necessary to me. Rather than handling a read, isnt it just effectively doing a read.

Would `ReadState::read()` not accomplish the same goal in a much simpler and clearer way?

extension/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+	/// Processes the parameters needed to execute a call within the runtime environment.
+	/// Layout of `params`: [version, pallet_index, call_index, ...Vec<u8>].
+	fn handle_read(params: T::RuntimeRead) -> Vec<u8>;
+}
+
+#[derive(Default)]
+pub struct ApiExtension;
```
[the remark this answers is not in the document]
evilrobot-01: Rather than being a type, would using a trait with some default function implementations not provide the same result but in a more flexible way? I gave the example of the XcmExecutor elsewhere, but perhaps an ApiExtension trait could inherit from ChainExtension<T>, add the additional function in ReadStateHandler, or inherit from that as well, and provide a single trait that can then be implemented for a type in the runtime.

extension/src/lib.rs
```diff
@@ -0,0 +1,297 @@
...
+		Err(err) => {
+			log::debug!(target:LOG_TARGET, "{} failed: error: {:?}", log_prefix, err.error);
+			Err(err.error)
+		},
+	}
+}
+
+/// Wrapper to enable versioning of runtime state reads.
+#[derive(Decode, Debug)]
+enum VersionedStateRead {
```
evilrobot-01, line 204: As mentioned elsewhere, not sure about the specifics of versioning in a low level generic extension implementation. I guess we could always stick with this for now and refactor in the future.

evilrobot-01 (changes requested):
Great job! Left various comments and suggestions for improvement.

The implementation is great, but feels a little complex from a maintenance perspective - at least in terms of implementing certain traits for certain types. I left a few suggestions on how this might be improved to make it a little more modular, but may not be suitable from a practical perspective.

One last thing - the title of the PR is misleading, as duplicating a chain extension isnt really the goal. My understanding was that it was attempting to provide a standard/generic implementation to de-duplicate code and reduce the maintenance burden.

extension/Cargo.toml
```diff
@@ -0,0 +1,52 @@
+[package]
```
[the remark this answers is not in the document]
Daanvdplas: Note: we need the chain extension as a crate so that it can be used by `drink` for integration tests.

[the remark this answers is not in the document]
evilrobot-01: Understood, just suggesting moving it into the runtime directory.

Some former requirements for drink support might inform some of the suggested improvements. Sure it's a runtime implementation of a chain extension tho? E.g. dev/test

extension/src/lib.rs
chungquantin: I agree. I believe versioning should be handled on the runtime side. During my self-review, I noticed that it's not very clean that `ReadState::read()` is currently invoked only in version v0. If more versions are introduced in the future, there should be a better design to handle this, and managing versioning on the runtime side would solve this issue. However, within the scope of this PR, which aims to refactor runtime extensions into a single generic extension, I don't think this should be elaborated further here.

extension/src/lib.rs
chungquantin: Thanks for this suggestion.

runtime/devnet/src/config/contracts.rs
chungquantin: @evilrobot-01 Thank you for a very thoughtful feedback. I created a sub-PR: https://github.com/r0gue-io/pop-node/pull/176 to show the alternative approach which removes the `Config` trait so on the Runtime side, it will be cleaner when there is no involvement between the `Runtime` and the extension configuration. However, this approach makes the extension crate code a bit more complex as there are generic types passing around between functions. Could you please skim over it and let me know if this is the implementation that we expect?

extension/src/constants.rs
```diff
@@ -0,0 +1,13 @@
+use sp_runtime::DispatchError;
+
+/// Logging target for categorizing messages from the Pop API extension module.
+pub(crate) const LOG_TARGET: &str = "pop-api::extension";
+
+pub const DECODING_FAILED_ERROR: DispatchError = DispatchError::Other("DecodingFailed");
```
evilrobot-01, line 6: Inconsistent doc comments for pub consts.

extension/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+};
+use sp_core::crypto::UncheckedFrom;
+use sp_runtime::{traits::Dispatchable, DispatchError};
+use sp_std::vec::Vec;
+
+type ContractSchedule<T> = <T as pallet_contracts::Config>::Schedule;
+
+/// Handles the query from the chain extension environment for state reads.
+pub trait ReadState {
+	type StateQuery: Decode;
```
evilrobot-01, line 27: A doc comment for the associated type would be a nice finishing touch.

extension/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+
+	/// Decodes parameters into state query.
+	fn decode(params: &mut &[u8]) -> Result<Self::StateQuery, DispatchError> {
+		decode_checked(params)
+	}
+}
+
+/// Handles the query from the chain extension environment for dispatch calls.
+pub trait CallFilter {
+	type Call: Decode;
```
evilrobot-01, line 43: A doc comment for the associated type would be a nice finishing touch.

extension/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+
+	/// Allowed runtime calls from the API.
+	fn contains(t: &Self::Call) -> bool;
+}
+
+#[derive(Default)]
+pub struct ApiExtension<I>(PhantomData<I>);
```
evilrobot-01, line 50: Missing doc comment.

extension/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+		let (version, function_id, pallet_index, call_index) = extract_env(&env);
+
+		let result = match FuncId::try_from(function_id) {
+			// Read encoded parameters from buffer and calculate weight for reading `len` bytes`.
+			Ok(function_id) => {
+				// reference: https://github.com/paritytech/polkadot-sdk/blob/117a9433dac88d5ac00c058c9b39c511d47749d2/substrate/frame/contracts/src/wasm/runtime.rs#L267
+				let len = env.in_len();
+				env.charge_weight(contract_host_weight.return_per_byte.saturating_mul(len.into()))?;
+				let params = env.read(len)?;
+				log::debug!(target: LOG_TARGET, "Read input successfully");
```
evilrobot-01, line 84: Do we still need this? Seems redundant. Perhaps it should be augmented with some state to be a bit more useful.

extension/src/lib.rs
```diff
@@ -0,0 +1,311 @@
...
+			1 => Self::ReadState,
+			_ => {
+				return Err(UNKNOWN_CALL_ERROR);
+			},
+		};
+		Ok(id)
+	}
+}
+
+// Converts a `DispatchError` to a `u32` status code based on the version of the API the contract uses.
```
evilrobot-01, line 265: Doc comments need ///

extension/Cargo.toml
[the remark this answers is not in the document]
evilrobot-01: Having taken another look, and considering a possible omninode future, it probably doesnt need to be moved into the runtime directory, so resolving this.

evilrobot-01 (changes requested):
Looks good, just missing doc comments as a finishing touch.

Strange that there is no code coverage report added to this PR, nor the base branch. I'll see if I can get that resolved.

extension/src/lib.rs
chungquantin: Created an issue: https://github.com/r0gue-io/pop-node/issues/180

extension/src/v0.rs
chungquantin: Created an issue: https://github.com/r0gue-io/pop-node/issues/180

extension/src/lib.rs
```diff
@@ -20,13 +18,26 @@ use sp_core::crypto::UncheckedFrom;
...
 type ContractSchedule<T> = <T as pallet_contracts::Config>::Schedule;
 
-/// Handles the query from the chain extension environment for state reads.
+/// Type of the state reader.
 pub trait ReadState {
+	/// Query of the state read operations.
 	type StateQuery: Decode;
 
-	/// Allowed state queries from the API.
+	/// Check if state queries from the API is allowlisted.
```
evilrobot-01, line 40:
```suggestion
	/// Check if a state query is allowed.
```

extension/src/lib.rs
```diff
@@ -38,14 +49,16 @@ pub trait ReadState {
...
 }
 
-/// Handles the query from the chain extension environment for dispatch calls.
+/// Type of the dispatch call filter.
 pub trait CallFilter {
+	/// Query of the dispatch calls operations.
 	type Call: Decode;
 
-	/// Allowed runtime calls from the API.
+	/// Check if runtime calls from the API is allowlisted.
```
evilrobot-01, line 57:
```suggestion
	/// Check if runtime call is allowed.
```

evilrobot-01 (changes requested):
Two last suggestions to simplify the language on the comments and then good to go!

extension/src/lib.rs
chungquantin: Thank you!

evilrobot-01 (approved):
LGTM, thanks for your patience!

PR #176, opened 2024-08-12
refactor: remove config and add generic extension type

chungquantin:
Alternative design of the PR: https://github.com/r0gue-io/pop-node/pull/163
- Removing the `Config` trait so there is no relationship between the `Runtime` and the extension. 
- Types need to be passed to the generic extension type `ApiExtension`. 
This alternative approach removes all the connections with the `Runtime`, every configuration will be on the level of the `ApiExtension`

### DIFF
--- a/extension/src/lib.rs
+++ b/extension/src/lib.rs
@@ -10,7 +10,7 @@ use constants::*;
 use frame_support::{
 	dispatch::{GetDispatchInfo, PostDispatchInfo},
 	pallet_prelude::*,
-	traits::{Contains, OriginTrait},
+	traits::OriginTrait,
 };
 use frame_system::RawOrigin;
 use pallet_contracts::chain_extension::{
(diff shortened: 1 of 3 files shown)


extension/src/lib.rs
```diff
@@ -210,7 +224,9 @@ enum VersionedStateRead {
 
 /// Wrapper to enable versioning of runtime calls.
 #[derive(Decode, Debug)]
-enum VersionedDispatch<T: Config> {
+enum VersionedDispatch<
+	T: frame_system::Config<RuntimeCall: GetDispatchInfo + Dispatchable<PostInfo = PostDispatchInfo>>,
```
evilrobot-01, line 228: Why cant this just take an opaque RuntimeCall as a generic parameter? It doesnt need anything more than that - e.g. the whole Runtime type being passed in only to use the RuntimeCall from it. This is private, so the calling function can just specifiy T::RuntimeCall, keeping this implementation simple.

evilrobot-01 (commented):
Just one query for one last possible improvement. Its minor, but I prefer to no overexpose things. The VersionedDispatch enum only needs to know the type of he `RuntimeCall`, not the whole `Runtime` which implements some `Config` trait to then determine the `RuntimeCall`.

Far simpler to only pass the runtime call type.

extension/src/lib.rs
chungquantin: Agree with you, pushed an update to convert it to only receive RuntimeCall instead of a whole Config. This part is a missing part during the refactor of the devnet extension. Thanks for pointing it out.

PR #184, opened 2024-08-13
style: formatting

evilrobot-01:
Improves formatting and checks through standardisation. The main goal is automation as much as possible, whether possible in the dev environment or automated via CI. 

### CI
This draft adds the following checks only (i.e. no auto-fixes).

- formatting checks of Rust files using Nightly, as it provides a number of additional useful configuration options (`.rustfmt.toml`)
  - can be run locally via `cargo +nightly fmt --all` or configured in editor/ide or commit hooks
- formatting checks of Cargo manifests using [taplo](https://taplo.tamasfe.dev/)
  - can be run locally with `taplo format --check` and `taplo format` to fix
- formatting and checks of features using zepter
  - can be run locally with `zepter format features` and `zepter format features --fix` to fix

Manifest formatting is generally alphabetically by 'region', inline with Rust imports which are alphabetical and grouped by source (std library, external, internal). I feel the latter does indeed have some benefit over one large grouping. Subjective grouping of imports can make it difficult for others to immediately understand how things should be grouped and applying accordingly. Using a standard approach eliminates this concern.

### Local
A pre-push git hook is included at `./githooks` and needs to be manually installed via the instructions in the readme at the same location. It currently only runs the Rust formatting checks as listed above, but can be extended to include the other checks noted as well. 

A pre-commit hook could also be added to auto-format, but most editors/ides probably allow this to be set on save so re-running it all again on each commit might be overkill.


Daanvdplas:
`cargo +nightly fmt` already seems to do a lot of things! Zepter and Taplo will definitely come in handy for the pop cli repo as well!

Again, this gives me so much piece of mind! Thank you very much ❤️

Daanvdplas (approved):
Very excited! Thanks for the effort you put into this. I think it will have a very positive impact on our development output as a team!

.rustfmt.toml
```diff
@@ -1,24 +1,23 @@
...
-trailing_comma = "Vertical"
+newline_style = "Unix"
+normalize_comments = true
+reorder_impl_items = true
 trailing_semicolon = false
+unstable_features = true
 use_field_init_shorthand = true
-# Format comments
-comment_width = 100
+use_small_heuristics = "Max"
```
al3mart, line 21: Maybe adding a comment like:

`# Applies `max_width` value or its default (`100`) if not specified.`

al3mart (approved):
I have open a PR to extend the githooks with TOML and features formating in https://github.com/r0gue-io/pop-node/pull/229.

Happy to rebase that PR if we want to merge this first! :+1:

.rustfmt.toml
evilrobot-01: Done

runtime/testnet/src/config/xcm.rs
```diff
@@ -171,34 +174,34 @@ pub type XcmRouter = WithUniqueTopic<(
 )>;
 
 impl pallet_xcm::Config for Runtime {
+	type AdminOrigin = EnsureRoot<AccountId>;
+	// ^ Override for AdvertisedXcmVersion default
```
chungquantin, line 178: Not related to this PR actually, but we miss a dot :o 
```suggestion
	// ^ Override for AdvertisedXcmVersion default.
```

chungquantin (approved):

peterwht (approved):

PR #190, opened 2024-08-14
docs: pop api

Daanvdplas:


pop-api/src/lib.rs
```diff
@@ -1,11 +1,20 @@
...
 #![cfg_attr(not(feature = "std"), no_std, no_main)]
 
 use constants::DECODING_FAILED;
 use ink::env::chain_extension::{ChainExtensionMethod, FromStatusCode};
 #[cfg(feature = "assets")]
 pub use v0::assets;
 
+/// Module providing primitives types.
 pub mod primitives;
+/// Version zero of the API.
```
evilrobot-01, line 17:
```suggestion
/// The first version of the API.
```

pop-api/src/lib.rs
```diff
@@ -1,11 +1,20 @@
+//! The `pop-api` crate provides an API for smart contracts to interact with the Pop network runtime.
```
evilrobot-01, line 1:
```suggestion
//! The `pop-api` crate provides an API for smart contracts to interact with the Pop Network runtime.
```

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -1,3 +1,10 @@
+//! The `fungibles` module provides an API for interacting and managing with fungible assets on Pop network.
```
evilrobot-01, line 1:
```suggestion
//! The `fungibles` module provides an API for interacting and managing fungible assets on Pop Network.
```

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -1,3 +1,10 @@
...
+//!
+//! 1. PSP-22 Interface
```
evilrobot-01, line 3: This needs some context added in the form of an intro sentence, otherwise its just a list of items with some vague link to the module content.

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -8,28 +15,22 @@ use constants::*;
 use ink::{env::chain_extension::ChainExtensionMethod, prelude::vec::Vec};
 pub use metadata::*;
 
-/// Helper method to build a dispatch call `ChainExtensionMethod` for fungibles `v0`.
-///
-/// Parameters:
-/// - 'dispatchable': The index of the module dispatchable functions.
+// Helper method to build a dispatch call `ChainExtensionMethod` for fungibles `v0`.
```
evilrobot-01, line 18:
```suggestion
// Helper method to build a dispatch call.
```

Its obvious that its building that based on the return type and that its in v0 based on the module scope. Unnecessary noise imo, and prone to error due to being overly verbose and then just copying, as is the case below.

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -8,28 +15,22 @@ use constants::*;
...
+//
+// Parameters:
+// - 'dispatchable': The index of the module dispatchable functions.
```
evilrobot-01, line 21:
```suggestion
// - 'dispatchable': The index of the dispatchable function within the module.
```

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -8,28 +15,22 @@ use constants::*;
...
 fn build_dispatch(dispatchable: u8) -> ChainExtensionMethod<(), (), (), false> {
 	crate::v0::build_dispatch(FUNGIBLES, dispatchable)
 }
 
-/// Helper method to build a dispatch call `ChainExtensionMethod` for fungibles `v0`.
-///
-/// Parameters:
-/// - 'state_query': The index of the runtime state query.
+// Helper method to build a dispatch call `ChainExtensionMethod` for fungibles `v0`.
```
evilrobot-01, line 26:
```suggestion
// Helper method to build a call to read state.
```

The name of the function clearly indicates that it isnt building a dispatch call, but a call to read state. The same mistake is made in the called function (crate::v0::build_read_state and crate::v0::build_dispatch) so please also update the doc comments accordingly all the way up the stack.

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -323,9 +324,46 @@ pub fn burn(id: AssetId, account: AccountId, value: Balance) -> Result<()> {
 		.call(&(id, account, value))
 }
 
+/// The interface for managing metadata of fungible assets. It includes the PSP-22 Metadata interface
+/// for querying metadata.
+// TODO: if reviewers agree with replacing metadata related apis here, the dispatchable index has to
```
evilrobot-01, line 329: Seems to sit better in relation to the PSP22 standard. Not sure about the `dispatchable index` changing, figure it may be something to do with the ordering of dispatchables to align with the standard? As they are not part of the standard, does the index matter?

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -323,9 +324,46 @@ pub fn burn(id: AssetId, account: AccountId, value: Balance) -> Result<()> {
...
 
+	/// Set the metadata for a token with a given asset ID.
+	///
+	/// # Parameters
+	/// - `id`: The identifier of the asset to update.
+	/// - `name`: The user friendly name of this asset. Limited in length by `StringLimit`.
+	/// - `symbol`: The exchange symbol for this asset. Limited in length by `StringLimit`.
+	/// - `decimals`: The number of decimals this asset uses to represent one unit.
+	///
+	/// # Returns
```
evilrobot-01, line 342: Personally dont see the value in the returns section, especially based on the fact that the description is obvious and so is the return type. A concern would be that if it is started, it needs to be done for every public function for consistency, bring wasted efforts in terms of implementation and reviews.

pop-api/src/v0/mod.rs
```diff
@@ -6,6 +6,7 @@ use crate::{
 };
 use ink::env::chain_extension::ChainExtensionMethod;
 
+/// APIs for assets related use cases.
```
evilrobot-01, line 9:
```suggestion
/// APIs for asset-related use cases.
```

pop-api/src/v0/assets/mod.rs
```diff
@@ -1,2 +1,3 @@
+/// APIs for fungibles assets.
```
evilrobot-01, line 1:
```suggestion
/// APIs for fungible assets.
```

evilrobot-01 (changes requested):
Some obvious errors which need fixing. Have provided various suggestions to speed up the process.

pop-api/src/v0/assets/fungibles.rs
Daanvdplas: Agree, removed

pop-api/src/v0/assets/fungibles.rs
Daanvdplas: Changed back. 

As for the index, it is just based on the order of the functions in the file.

Daanvdplas:
Thanks for the review! All comments resolved

pop-api/src/v0/assets/fungibles.rs
```diff
@@ -375,6 +341,7 @@ pub mod metadata {
 	}
 }
 
+/// The interface for creating, managing and destroying fungible assets.
 pub mod asset_management {
```
evilrobot-01, line 345:
```suggestion
pub mod management {
```

Might be simpler.

evilrobot-01 (approved):
Minor adjust to asset_management module suggested, not a blocker from merging tho.

PR #194, opened 2024-08-15
chore: upgrade to 1.13.0

chungquantin:
Status: ✅ Tested on my local `devnet`, `testnet` => Block production / finalization + submit extrinsic works well. 
- Features: ✅ try-runtime ✅ runtime-benchmarks ✅ metadata-hash
- ISSUE: https://github.com/r0gue-io/pop-node/issues/141
- Release: https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.13.0
- Prerequisite: https://github.com/r0gue-io/pop-node/pull/193

### References
- https://github.com/r0gue-io/base-parachain/pull/38/
- Compare `pallet-contracts`: [`v1.12.0`](https://github.com/paritytech/polkadot-sdk/commits/polkadot-v1.10.0/substrate/frame/contracts) and [`v1.13.0`](https://github.com/paritytech/polkadot-sdk/commits/polkadot-v1.11.0/substrate/frame/contracts)
- Compare `pallet-xcm`: [`v1.12.0`](https://github.com/paritytech/polkadot-sdk/commits/polkadot-v1.10.0/substrate/frame/xcm) and [`v1.13.0`](https://github.com/paritytech/polkadot-sdk/commits/polkadot-v1.11.0/substrate/frame/xcm)
## Direct changes
- https://github.com/paritytech/polkadot-sdk/pull/4380
- https://github.com/paritytech/polkadot-sdk/pull/3952
- https://github.com/paritytech/polkadot-sdk/pull/4410
- https://github.com/paritytech/polkadot-sdk/pull/4233
- https://github.com/paritytech/polkadot-sdk/pull/4274

## Related changes
- https://github.com/paritytech/polkadot-sdk/pull/4131
- https://github.com/paritytech/polkadot-sdk/pull/4418

PR #179, opened 2024-08-12
chore: upgrade to 1.9.0

chungquantin:
Status: ✅ Tested on my local `devnet`, `testnet` => Block production / finalization + submit extrinsic works well. 

- ISSUE: https://github.com/r0gue-io/pop-node/issues/141
- Reference `base-parachain`: https://github.com/r0gue-io/base-parachain/pull/9/files
- Release: https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.9.0

- Compare `pallet-contracts`: [`v1.8.0`](https://github.com/paritytech/polkadot-sdk/commits/polkadot-v1.8.0/substrate/frame/contracts) and [`v1.9.0`](https://github.com/paritytech/polkadot-sdk/commits/polkadot-v1.9.0/substrate/frame/contracts)
## CHANGELOG
- [ ] https://github.com/paritytech/polkadot-sdk/pull/3002 (not included in this upgrade until the POV reclaim feature is merged to the system parachains, see comment)
- [x] Permissioned contract deployment on a chain
- [x] Multi-Block-Migrations,`poll` hook and new System Callbacks ([link to prdoc](https://github.com/paritytech/polkadot-sdk/pull/1781))
- [x] Removes `as [disambiguation_path]` from the required syntax in `derive_impl`

#### [#3377 - Permissioned contract deployment](https://github.com/paritytech/polkadot-sdk/pull/3377)
- Link to discussion: https://github.com/paritytech/polkadot-sdk/issues/3196

This PR introduces two new config types that specify the origins allowed to upload and instantiate contract code. However, this check is not enforced when a contract instantiates another contract.

### Async Backing Issues
- [Fixup multi-collator parachain transition to async backing #3510](https://github.com/paritytech/polkadot-sdk/pull/3510)

PR #198, opened 2024-08-16
chore: upgrade v1.14.0

chungquantin:
Status: ✅ Compiled and built successfully

- ISSUE: [https://github.com/r0gue-io/base-parachain/issues/39](https://github.com/r0gue-io/base-parachain/issues/39)
- Release: [https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.14.0](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.14.0)
- Prerequisite: 
- [x] https://github.com/r0gue-io/base-parachain/pull/41
- [x] https://github.com/r0gue-io/pop-node/pull/194
- [x] https://github.com/r0gue-io/pop-node/pull/201

## Direct changes
- Add POV reclaim feature (introduced in v1.9.0 but there is a fix PR in v1.14.0).
- Migrate XCMv5 
- Introduce the construct runtime V2
- Use aggregated types for `RuntimeFreezeReason`: [paritytech/polkadot-sdk@988103d](https://github.com/paritytech/polkadot-sdk/commit/988103d7578ad515b13c69578da1237b28fa9f36)
- [[#4730](https://github.com/paritytech/polkadot-sdk/pull/4730)]: rpc upgrade jsonrpsee to v0.23.1
- [[#4769](https://github.com/paritytech/polkadot-sdk/pull/4769)]: Use real rust type for pallet alias in `runtime` macro

## Related changes

- [[#3679](https://github.com/paritytech/polkadot-sdk/pull/3679)]: [pallet-contracts] bump wasmi to 0.32
- [[#4799](https://github.com/paritytech/polkadot-sdk/pull/4799)]: network: Upgrade `litep2p` to v0.6.0


al3mart (approved):
LGTM!

Logged https://github.com/r0gue-io/pop-node/issues/210

Daanvdplas (approved):
Both testnet and devnet run with 6 second block time. One note it that `frame-metadata-hash` is not added to the `std` feature for both runtimes

Daanvdplas (changes requested):
After comparing the runtime with the parachain template in the SDK on 1.14. 

#### Runtime
- Missing [pov reclaim signed extension](https://github.com/paritytech/polkadot-sdk/blob/364961445b74027f0c1b2a0c8661b2dfd3476aec/templates/parachain/runtime/src/lib.rs#L87) (should have been added in [1.9](https://github.com/r0gue-io/base-parachain/commit/85d7c21032bc9e75459a4be7a6855a1ab5dc7d0f#diff-0ec06ea58bd455f09ce6b3bb4c2c1c0d37bda51c1e1be2151c560c9c973959ecL106))

#### Node
-[ ReclaimHostFunctions](https://github.com/paritytech/polkadot-sdk/blob/364961445b74027f0c1b2a0c8661b2dfd3476aec/templates/parachain/node/src/command.rs#L186), should also have been added to 1.9.

The 1.9 upgrade has been done very sloppy, especially when there is a reference from the [base-parachain repo](https://github.com/r0gue-io/base-parachain/commit/85d7c21032bc9e75459a4be7a6855a1ab5dc7d0f). Everything regarding the pov reclaim has been forgotten in the [1.9 upgrade](https://github.com/r0gue-io/pop-node/pull/179/files).

Daanvdplas:
Could we add that in the description please? So that we have a reference to it

chungquantin:
As we discussed in the 1.10.0 upgrade PR: https://github.com/paritytech/polkadot-sdk/pull/5371/files to whether we should add back the POV reclaim. From my understanding, it is fair to say that we can add back the feature in this v1.14.0 upgrade because there is a PR to fix the feature in this version: https://github.com/paritytech/polkadot-sdk/pull/5371/files

Hence, we can reimplement that for both `devnet` and `testnet`. What do you think? 

cc: @AlexD10S @Daanvdplas @peterwht

Daanvdplas:
There is also this migration: https://github.com/paritytech/polkadot-sdk/blob/946afaabd8244f1256f3aecff75e23c02937bd38/polkadot/xcm/pallet-xcm/src/migration.rs#L84

And this change in [1.13](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.13.0) (search for `type MaxPageSize = ConstU32<{ 103 * 1024 }>;`)

Would love to hear @al3mart opinion on both before putting it in.

al3mart:
1)

About the PoV reclaim feature I think we are OK to include it as part of this PR if we want to have it.
It definitely seems beneficial!

2)

[`MigrateToLatestXcmVersion` ](https://github.com/paritytech/polkadot-sdk/blob/946afaabd8244f1256f3aecff75e23c02937bd38/polkadot/xcm/pallet-xcm/src/migration.rs#L84) seems something that we have been missing for longer than this migrations even, so including it now would be good.

3)

On `MaxPageSize`; our only open channel for the testnet right now is using a configuration of `maxMessageSize: 8,192`, which is well under our current value of `1 << 16`, so we should be OK.
Though, if deploying a local environment as in `pop up parachain -f networks/devnet.toml` that would spin a relay that can values for that message size up to `1,048,576`, so using a config of `type MaxPageSize = ConstU32<{ 103 * 1024 }>;` for the devnet runtime should put us just above it :ok_hand: 
For the sake of simplicity I'd say that updating the config on both runtimes to `type MaxPageSize = ConstU32<{ 103 * 1024 }>;` would be the better option.

Cargo.toml
```diff
@@ -116,53 +116,53 @@ sp-io = { version = "37.0.0", default-features = false }
...
+cumulus-pallet-xcmp-queue = { version = "0.15.0", default-features = false }
 cumulus-primitives-aura = { version = "0.14.0", default-features = false }
 cumulus-primitives-core = { version = "0.14.0", default-features = false }
-cumulus-primitives-storage-weight-reclaim = { version = "5.0.0", default-features = false }
-cumulus-primitives-utility = { version = "0.14.0", default-features = false }
-emulated-integration-tests-common = { version = "10.0.0", default-features = false }
-pallet-collator-selection = { version = "16.0.0", default-features = false }
-parachains-common = { version = "14.0.0", default-features = false }
-parachain-info = { version = "0.14.0", package = "staging-parachain-info", default-features = false }
+cumulus-primitives-storage-weight-reclaim = { version = "6.0.2", default-features = false }
```
peterwht, line 151: unnecessary if we don't add PovReclaim.

peterwht (changes requested):
So close! Great job.

As this is the last PR in the upgrade to 1.14, we should make sure all migrations are added.

try-runtime says we need a migration for XcmpQueue and pallet-contracts.

```
[2024-08-20T01:32:30Z ERROR runtime::frame-support] XcmpQueue: On chain storage version StorageVersion(4) doesn't match in-code storage version StorageVersion(5).
[2024-08-20T01:32:30Z ERROR runtime::frame-support] Contracts: On chain storage version StorageVersion(15) doesn't match in-code storage version StorageVersion(16).

```

peterwht:
Let's go ahead and just add `PovReclaim` now that it has fixes. Just to note, we may have some changes in the future https://github.com/paritytech/polkadot-sdk/pull/5234

chungquantin:
1. Migrations for contract, XCMP queue and XCM is added to the `testnet` runtime. Tested with `try-runtime` and works well.
```rs
	pallet_contracts::Migration<Runtime>,
	cumulus_pallet_xcmp_queue::migration::v5::MigrateV4ToV5<Runtime>,
	pallet_xcm::migration::MigrateToLatestXcmVersion<Runtime>,
```
2. POV reclaim feature is added for both `devnet` and `testnet`
3. Update the type parameter value of the HRMP channel max size

cc: @Daanvdplas @al3mart @peterwht 

Before merging this PR, could you guys please help me to review this PR as well: https://github.com/r0gue-io/pop-node/pull/201 which fixes the integration tests.

peterwht (approved):
LGTM!!! Glad to see this so close being done.

Daanvdplas (approved):

PR #200, opened 2024-08-16
[WIP] refactor: generic api

Daanvdplas:
Just a draft of how it might look like. Facing a problem that the `ChainExtension` type in pallet contracts is bounded by the `Default` trait. Because `Runtime` doesn't implement the `Default` trait I'm not sure how we can make the `ExtensionTrait` generic over Runtimes. 

Not sure whether it is an easy fix, still feel that it can be simplified though. Don't mind the naming at the moment. Just wanted to get something working and get feedback.

I put `dispatch` and `dispatch_call` in one function for simplicity for the refactoring.

### DIFF
--- a/extension/src/lib.rs
+++ b/extension/src/lib.rs
@@ -21,13 +21,9 @@ use sp_std::vec::Vec;
 /// Logging target for categorizing messages from the Pop API extension module.
 const LOG_TARGET: &str = "pop-api::extension";
 
-const DECODING_FAILED_ERROR: DispatchError = DispatchError::Other("DecodingFailed");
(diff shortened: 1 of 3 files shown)


chungquantin:
Just occasionally come across this PR and think that this can help: 
> Facing a problem that the ChainExtension type in pallet contracts is bounded by the Default trait.

I guess if you use the `PopExtensionApiConfig` that I had before, you can get rid of this where the `Runtime` type is separated from the `ApiExtension` but can still keep what you are having because `Runtime` is an associated type of `ApiExtension. `

evilrobot-01:
https://github.com/r0gue-io/pop-node/compare/daan/api...frank/extension

PR #201, opened 2024-08-16
fix: migrate integration tests from `paseo` to `polkadot` runtimes

chungquantin:
## CHANGELOG
- Migrate the Paseo runtimes to Polkadot runtimes as a stopgap until Paseo is updated to v1.14.0. 
- Replace deprecated method `transfer_assets_delivery_fees` to `teleport_assets_delivery_fees`.
- Replace `Withdraw` event with `Burned` and `Deposit` event with `Minted`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2552,9 +2552,9 @@ dependencies = [
 
 [[package]]
 name = "cumulus-primitives-storage-weight-reclaim"
-version = "6.0.1"
+version = "6.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d9e02d29ca51a4911585fdd5381e5f1cb4ee553e4f20b843c08917dfba881ea7"
+checksum = "29a7e13063f593f21534a7b64c96f311c40cd4d3c72649e0bd063a34506fdd70"
 dependencies = [
  "cumulus-primitives-core",
  "cumulus-primitives-proof-size-hostfunction",
@@ -3897,9 +3897,9 @@ dependencies = [
 
 [[package]]
 name = "frame-system"
-version = "36.0.1"
+version = "36.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d19f388280c14dbabcee2a817e1123c59c55847c6d07a5b5ca901b34a5910093"
+checksum = "64d6a0e7bb6503facdcc6f8e19c83cd0bfc8bbbd268522b1a50e107dfc6b972d"
 dependencies = [
  "cfg-if",
  "docify",
@@ -9934,9 +9934,9 @@ dependencies = [
 
 [[package]]
 name = "polkadot-runtime-parachains"
-version = "15.0.2"
+version = "15.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a6cb1cf1fa41c64b0b515a33d7bd388c3af0de31e5f93bd4c2b220e408f7f968"
+checksum = "cb7e68cb26f9025daaad694d8192fd0e63e92c8761c45a339dd7a5b7925a3cb6"
 dependencies = [
  "bitflags 1.3.2",
  "bitvec",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -461,7 +461,7 @@ checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
 [[package]]
 name = "asset-hub-polkadot-runtime"
 version = "1.0.0"
-source = "git+https://github.com/polkadot-fellows/runtimes?rev=4260e50#4260e50d46c13befb127211a8f7230662366998a"
+source = "git+https://github.com/polkadot-fellows/runtimes#f42acab60edf4d6ded4d9e99b1a8fbacded85669"
 dependencies = [
  "assets-common",
  "bp-asset-hub-kusama",
@@ -1100,7 +1100,7 @@ dependencies = [
 [[package]]
 name = "bp-asset-hub-kusama"
 version = "1.0.0"
-source = "git+https://github.com/polkadot-fellows/runtimes?rev=4260e50#4260e50d46c13befb127211a8f7230662366998a"
+source = "git+https://github.com/polkadot-fellows/runtimes#f42acab60edf4d6ded4d9e99b1a8fbacded85669"
 dependencies = [
  "bp-xcm-bridge-hub-router",
  "frame-support",
@@ -1114,7 +1114,7 @@ dependencies = [
 [[package]]
 name = "bp-asset-hub-polkadot"
 version = "1.0.0"
-source = "git+https://github.com/polkadot-fellows/runtimes?rev=4260e50#4260e50d46c13befb127211a8f7230662366998a"
+source = "git+https://github.com/polkadot-fellows/runtimes#f42acab60edf4d6ded4d9e99b1a8fbacded85669"
 dependencies = [
  "bp-xcm-bridge-hub-router",
  "frame-support",
@@ -1144,7 +1144,7 @@ dependencies = [
 [[package]]
 name = "bp-bridge-hub-kusama"
 version = "1.0.0"
-source = "git+https://github.com/polkadot-fellows/runtimes?rev=4260e50#4260e50d46c13befb127211a8f7230662366998a"
+source = "git+https://github.com/polkadot-fellows/runtimes#f42acab60edf4d6ded4d9e99b1a8fbacded85669"
 dependencies = [
  "bp-bridge-hub-cumulus",
  "bp-messages",
@@ -1161,7 +1161,7 @@ dependencies = [
 [[package]]
 name = "bp-bridge-hub-polkadot"
 version = "1.0.0"
-source = "git+https://github.com/polkadot-fellows/runtimes?rev=4260e50#4260e50d46c13befb127211a8f7230662366998a"
+source = "git+https://github.com/polkadot-fellows/runtimes#f42acab60edf4d6ded4d9e99b1a8fbacded85669"
 dependencies = [
  "bp-bridge-hub-cumulus",
  "bp-messages",
@@ -1716,7 +1716,7 @@ dependencies = [
 [[package]]
 name = "collectives-polkadot-runtime-constants"
 version = "1.0.0"
-source = "git+https://github.com/polkadot-fellows/runtimes?rev=4260e50#4260e50d46c13befb127211a8f7230662366998a"
+source = "git+https://github.com/polkadot-fellows/runtimes#f42acab60edf4d6ded4d9e99b1a8fbacded85669"
 
 [[package]]
 name = "color-print"
@@ -5295,7 +5295,7 @@ checksum = "c33070833c9ee02266356de0c43f723152bd38bd96ddf52c82b3af10c9138b28"
 [[package]]
 name = "kusama-runtime-constants"
 version = "1.0.0"
-source = "git+https://github.com/polkadot-fellows/runtimes?rev=4260e50#4260e50d46c13befb127211a8f7230662366998a"
+source = "git+https://github.com/polkadot-fellows/runtimes#f42acab60edf4d6ded4d9e99b1a8fbacded85669"
 dependencies = [
  "frame-support",
  "polkadot-primitives",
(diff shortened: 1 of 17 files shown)


Cargo.toml
```diff
@@ -164,7 +164,7 @@ cumulus-client-consensus-common = "0.15.0"
 cumulus-client-consensus-proposer = "0.14.0"
 cumulus-client-service = "0.15.0"
 
-# Paseo
-asset-hub-paseo-runtime = { git = "https://github.com/paseo-network/runtimes/", tag = "v1.2.5-system-chains", default-features = false }
-paseo-runtime = { git = "https://github.com/paseo-network/runtimes/", tag = "v1.2.5-system-chains", default-features = false }
-paseo-runtime-constants = { git = "https://github.com/paseo-network/runtimes/", tag = "v1.2.5-system-chains", default-features = false }
+# Paseo (note: using polkadot as stopgap until paseo updated to polkadot sdk v1.14.0)
```
Daanvdplas, line 167: Can we write TODO here as well so that it will be easier to find.

Cargo.toml
```diff
@@ -164,7 +164,7 @@ cumulus-client-consensus-common = "0.15.0"
...
+asset-hub-polkadot-runtime = { git = "https://github.com/polkadot-fellows/runtimes", rev = "4260e50", default-features = false }
```
Daanvdplas, line 168: What specific commit is this? Why are we not pointing to e.g. a tag?

integration-tests/src/chains/asset_hub_paseo/mod.rs
```diff
@@ -1,3 +1,5 @@
+// Note: using polkadot as stopgap until paseo updated to polkadot sdk v1.14.0
```
Daanvdplas, line 1: `TODO`

integration-tests/src/chains/asset_hub_paseo/mod.rs
```diff
@@ -13,27 +15,28 @@ decl_test_parachains! {
 	pub struct AssetHubPaseo {
 		genesis = genesis::genesis(),
 		on_init = {
-			asset_hub_paseo_runtime::AuraExt::on_initialize(1);
+			asset_hub_runtime::AuraExt::on_initialize(1);
 		},
-		runtime = asset_hub_paseo_runtime,
+		// Note: using polkadot as stopgap until paseo updated to polkadot sdk v1.14.0
```
Daanvdplas, line 20: `TODO`

integration-tests/src/lib.rs
```diff
@@ -95,7 +97,7 @@ fn para_receiver_assertions<Test>(_: Test) {
 	assert_expected_events!(
 		PopNetworkPara,
 		vec![
-			RuntimeEvent::Balances(pallet_balances::Event::Deposit { .. }) => {},
+			RuntimeEvent::Balances(pallet_balances::Event::Minted { ..}) => {},
```
Daanvdplas, line 100:
```suggestion
			RuntimeEvent::Balances(pallet_balances::Event::Minted { .. }) => {},
```

integration-tests/src/chains/asset_hub_paseo/mod.rs
```diff
@@ -13,27 +15,28 @@ decl_test_parachains! {
...
 }
 
 // AssetHubPaseo implementation
 impl_accounts_helpers_for_parachain!(AssetHubPaseo);
 impl_assert_events_helpers_for_parachain!(AssetHubPaseo);
-impl_assets_helpers_for_parachain!(AssetHubPaseo, Paseo);
-impl_foreign_assets_helpers_for_parachain!(AssetHubPaseo, Paseo);
+impl_assets_helpers_for_parachain!(AssetHubPaseo);
+impl_foreign_assets_helpers_for_parachain!(AssetHubPaseo, xcm::v3::Location);
 impl_xcm_helpers_for_parachain!(AssetHubPaseo);
```
Daanvdplas, line 42: Should this not be `Assethub`?

Why on line 40 & 41 not `Polkadot`?

integration-tests/src/chains/paseo/genesis.rs
```diff
@@ -1,16 +1,17 @@
 use emulated_integration_tests_common::{
-	accounts, build_genesis_storage, get_account_id_from_seed, get_from_seed, get_host_config,
-	validators,
+	accounts, build_genesis_storage, get_from_seed, get_host_config, validators,
 };
-use paseo_runtime_constants::currency::UNITS as PAS;
 use polkadot_primitives::{AssignmentId, Balance, ValidatorId};
+use polkadot_runtime as runtime;
+use polkadot_runtime_constants as runtime_constants;
+use runtime_constants::currency::UNITS as PAS;
```
Daanvdplas, line 7: Can stay the same because we are changing back to PAS

integration-tests/src/lib.rs
```diff
@@ -460,7 +458,7 @@ fn reserve_transfer_native_asset_from_para_to_system_para() {
 	let receiver_balance_after = test.receiver.balance;
 
 	let delivery_fees = PopNetworkPara::execute_with(|| {
-		xcm_helpers::transfer_assets_delivery_fees::<
+		xcm_helpers::teleport_assets_delivery_fees::<
```
Daanvdplas, line 461: I don't get why we are teleporting now and before we were doing a reserve transfer?

Pop Network would never be allowed to teleport DOT no?

Daanvdplas (commented):
Mainly questions but would be good if @evilrobot-01 can take a quick look.

integration-tests/src/lib.rs
chungquantin: `transfer_assets_delivery_fees` is removed since `asset-test-utils v11.0.0`. 

Reference: https://docs.rs/asset-test-utils/11.0.0/asset_test_utils/xcm_helpers/index.html

Cargo.toml
evilrobot-01: As they dont generate a runtime release for each bump of the SDK, this commit is the one that updates the runtimes to 1.14: https://github.com/polkadot-fellows/runtimes/commits/main/

@chungquantin could you try it without the commit to see if it all still compiles? It seems the fellowship runtimes are still on 1.14 so the latest main should suffice, relying on the lock file to pin.

integration-tests/src/chains/asset_hub_paseo/mod.rs
chungquantin: - Line 40: Since this commit: https://github.com/paritytech/polkadot-sdk/pull/3543/files, the `impl_assets_helpers_for_parachain` no longer asks for the Relaychain in the macro.
- Line 41: I follow Polkadot runtime that does not use Polkadot: https://github.com/polkadot-fellows/runtimes/blob/main/integration-tests/emulated/chains/parachains/testing/penpal/src/lib.rs#L64

integration-tests/src/chains/asset_hub_paseo/mod.rs
evilrobot-01: can actually remove this comment due to the type alias.

integration-tests/src/chains/asset_hub_paseo/mod.rs
evilrobot-01: Technically speaking the specification can be important in the context of bridges, whereby you could have Asset Hub on Polkadot and Kusama.

As we arent there yet, I would be fine with just `AssetHub`, but could perhaps be out of scope of the PR which is just fixing tests. Up to Tin on whether to address here.

Line 41 is due to the API changing with xcm v4, and AH using xcm v3. 40 is also an API change in the emulated tests macros.

integration-tests/src/chains/paseo/genesis.rs
evilrobot-01: Probably not as that dependency no longer exists.

integration-tests/src/chains/paseo/genesis.rs
```diff
@@ -20,8 +21,8 @@ fn session_keys(
 	para_assignment: AssignmentId,
 	authority_discovery: AuthorityDiscoveryId,
 	beefy: BeefyId,
-) -> paseo_runtime::SessionKeys {
-	paseo_runtime::SessionKeys {
+) -> polkadot_runtime::SessionKeys {
```
evilrobot-01: Can we use the same `runtime` alias here please?
```suggestion
) -> runtime::SessionKeys {
```

integration-tests/src/chains/paseo/genesis.rs
```diff
@@ -20,8 +21,8 @@ fn session_keys(
...
+	polkadot_runtime::SessionKeys {
```
evilrobot-01:
```suggestion
	runtime::SessionKeys {
```

integration-tests/src/lib.rs
```diff
@@ -5,7 +5,9 @@ use crate::chains::{
 	paseo::{genesis::ED as PASEO_ED, PaseoRelayPallet},
 	pop_network::PopNetworkParaPallet,
 };
-use asset_hub_paseo_runtime::xcm_config::XcmConfig as AssetHubPaseoXcmConfig;
+// Note: using polkadot as stopgap until paseo updated to polkadot sdk v1.14.0
+use asset_hub_polkadot_runtime as asset_hub_runtime;
```
evilrobot-01: Might be better to jsut use an export from asset_hub/src.rs if possible. That way the alias is defined in one place.

Cargo.lock
```diff
@@ -459,83 +459,89 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
```
evilrobot-01, line 459: Cargo.lock needs updating

integration-tests/src/lib.rs
```diff
@@ -112,11 +116,9 @@ fn para_to_system_para_sender_assertions(t: ParaToSystemParaTest) {
 	assert_expected_events!(
 		PopNetworkPara,
 		vec![
-			// Amount to reserve transfer is transferred to Parachain's Sovereign account
-			RuntimeEvent::Balances(
-				pallet_balances::Event::Withdraw { who, amount }
-			) => {
-				who: *who == t.sender.account_id,
+			// Amount to teleport is withdrawn from Sender
```
evilrobot-01, line 119:
```suggestion
			// Amount to reserve transfer is transferred to Parachain's Sovereign account
```

There is no teleport happening here as PopNet is not a trusted teleporter. I believe there was probably a change from CurrencyAdapter to FungiblesAdapter in the version uplift, which works in a different way, emitting burns/mints instead.

integration-tests/src/lib.rs
```diff
@@ -112,11 +116,9 @@ fn para_to_system_para_sender_assertions(t: ParaToSystemParaTest) {
...
+			RuntimeEvent::Balances(pallet_balances::Event::Burned { who, amount }) => {
+				who: *who == t.receiver.account_id,
```
evilrobot-01, line 121:
```suggestion
				who: *who == t.sender.account_id,
```

This is wrong and needs to be reverted.

integration-tests/src/lib.rs
```diff
@@ -132,11 +134,9 @@ fn para_to_relay_sender_assertions(t: ParaToRelayTest) {
 	assert_expected_events!(
 		PopNetworkPara,
 		vec![
-			// Amount to reserve transfer is transferred to Parachain's Sovereign account
-			RuntimeEvent::Balances(
-				pallet_balances::Event::Withdraw { who, amount }
-			) => {
-				who: *who == t.sender.account_id,
+			// Amount to teleport is withdrawn from Sender
```
evilrobot-01, line 137:
```suggestion
			// Amount to reserve transfer is transferred to Parachain's Sovereign account
```

As above.

integration-tests/src/lib.rs
```diff
@@ -132,11 +134,9 @@ fn para_to_relay_sender_assertions(t: ParaToRelayTest) {
...
+			RuntimeEvent::Balances(pallet_balances::Event::Burned { who, amount }) => {
+				who: *who == t.receiver.account_id,
```
evilrobot-01, line 139:
```suggestion
				who: *who == t.sender.account_id,
```

As above, needs fixing.

evilrobot-01 (changes requested):
Great on the whole, needs a few fixes tho please.

chungquantin:
@evilrobot-01 Thanks for your review, Frank. After re-reviewing my own code, I decided to do dependency alias instead so we can get rid of all the unnecessary changes related to crate name. Please help me to review it again!

al3mart (approved):

PR #192, opened 2024-08-14
refactor: error

Daanvdplas:
Closes #204

Additionally it removes the pop primitives dependency from the extension crate since it is no longer needed. This should have been added to #211 but forgotten.

Issue #188, opened 2024-08-14
tests: increase test coverage chain extension


Daanvdplas:
https://github.com/r0gue-io/pop-node/pull/132#issuecomment-2287155002

Besides the error handling, most of the chain extension functionality is untested. Since it is difficult to mock the `Environment` due to the private `Runtime` struct in `Environment { inner, ... }` we might need to refactor the code in order to increase the test coverage.

#202 might help.

Daanvdplas:
[sc-1156]

Daanvdplas:
Haven't been able to find a way to write tests for `call`, `read_state`, `dispatch`, `extract_env` mainly because it is hard to mock the environment. I need to create a `Runtime` which is private so still researching how to do this best.

Daanvdplas:
closed by #240

PR #211, opened 2024-08-17
test: improve test cases

Daanvdplas:
Refactors the the error handling in separate functions for easier testing between modules. The tests have also been refactored into separate tests for each error variant and more edge cases. More importantly, the tests don't test the `Error` from primitives but the raw bytes. 

In #192, `Error` is refactored to the pop api crate, note that the pop primitives crate is removed from the extension crate in this PR since here it was forgotten.

In the second commit docs have been added to the `v0` module as well.

Helping with #188

### DIFF
--- a/extension/src/lib.rs
+++ b/extension/src/lib.rs
@@ -18,16 +18,12 @@ use sp_std::vec::Vec;
 mod tests;
 mod v0;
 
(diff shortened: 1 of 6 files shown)


Daanvdplas:
No longer needed due to #218

PR #214, opened 2024-08-19
fix: invalid imported crates & crate visibility

chungquantin:
Fix existing error in the `daan/api` branch occurs in the contract & contract integration test.

### DIFF
--- a/pop-api/src/v0/assets/fungibles.rs
+++ b/pop-api/src/v0/assets/fungibles.rs
@@ -11,9 +11,9 @@ use crate::{
 	primitives::{AccountId, AssetId, Balance},
 	Result, StatusCode,
 };
-pub use asset_management::*;
 use constants::*;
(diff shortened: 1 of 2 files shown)


pop-api/src/v0/assets/fungibles.rs
```diff
@@ -485,13 +485,11 @@ mod tests {
 	use super::FungiblesError;
 	use crate::{
 		constants::{ASSETS, BALANCES},
-		primitives::error::{
```
Daanvdplas, line 488: Sure this works? I think this should only be changed when these types are removed from primitives crate

chungquantin: Apologies, this shouldn't be changed. Just revert the changes.

Daanvdplas:
Thank you Tin!

PR #218, opened 2024-08-19
refactor: generic extension

### LINKED ISSUE #202: refactor: generic extension
Handling of the version doesnt feel right in the generic extension and think that it should be a runtime level consideration. The actual types should be provided by the runtime (https://github.com/r0gue-io/pop-node/pull/132#discussion_r1717500819).

See #200 if you want to hack on it.

evilrobot-01:
Refactors the chain extension to maximise flexibility, ensuring that it is highly configurable and so that anything which is runtime specific is defined there instead - e.g. versioning. 

A high level overview of the components:

[contract environment: contract ⇄ `pop-api`] ⇄ [runtime: `pallet-contracts` ⇄ `pop-chain-extension` ⇄ `pallet-api` ⇄ `fungibles`]

The simplest analogy is that of a pipeline, where a request arrives and moves through a pipeline where it can be versioned (processed) based on other information available from the environment and then decoded into a valid runtime call. `pop-api` chooses to encode the version within the ext_id/func_id for example. The reverse also then allows for versioning by adapting any result (read) from a pallet or error into a type/shape the applies to the specific version. More specifics below.

### Errors
The extension supports conversion of a `DispatchError` into some other error type. This is to allow a `pop-api` specific error type which is also versionable, which in turn is converted into a u32 status code used by the contract. The contract can then optionally convert this status code back into a specific versioned error type if desired, with some contract size overhead.

### Reads
The results of reads also need to be versionable, so the `ReadState` function allows an additional conversion to be configured. This is so that we can simply map a current read result type from a pallet into some prior versioned result type before encoding it into the bytes expected by the contract based on the version requested.

### Versioning
Versioning is considered a runtime-specific requirement and all related logic is therefore contained within the `versioning` module within the runtime. The generic chain extension is configured to be version aware by the various types that are plugged into it.

### TODOs
- [x] unit tests

Closes #202


extension/src/reboot.rs
```diff
@@ -0,0 +1,243 @@
...
+			type C = C;
+			type Output = Output;
+
+			fn build<E: Ext<T = Self::C>>(
+				env: &mut Environment<E, BufInBufOutState>,
+			) -> Result<Self::Output> {
+				let contract_host_weight = <C as Config>::Schedule::get().host_fn_weights;
+				let len = env.in_len();
+				env.charge_weight(contract_host_weight.return_per_byte.saturating_mul(len.into()))?;
+				let params = env.read(len)?;
```
Daanvdplas, line 184: This could potentially be called in `fn call` and `params` given as a parameter to `execute` and `build`.

Daanvdplas (commented):
This fulfils all the needs that we have identified and well done with the amount of flexibility this implementation provides!

In the past we've experienced how useful this flexibility can be so very much onboard.

extension/src/reboot.rs
```diff
@@ -0,0 +1,243 @@
...
+		let mut env = env.buf_in_buf_out();
+
+		let contract_host_weight = <C as Config>::Schedule::get().host_fn_weights;
+		env.charge_weight(contract_host_weight.debug_message)?;
+
+		F::Function::execute::<E>(env)
+	}
+}
+
+// Simple trait to allow configuration of chain extension functions (workaround for Default requirement)
```
Daanvdplas, line 45: ✍️

extension/src/reboot.rs
evilrobot-01: Seems very outdated based on latest push, let me know if still needs addressing in latest version.

pallets/api/Cargo.toml
```diff
@@ -11,8 +11,12 @@ targets = ["x86_64-unknown-linux-gnu"]
 
 [dependencies]
 codec.workspace = true
+log.workspace = true
 scale-info.workspace = true
 
+# Local
+pop-chain-extension.workspace = true
```
evilrobot-01, line 18: This is a side effect of trying to isolate anything Pop API related, reusable across runtimes, within pallet-api.

The effectively makes the pop-api crate the ink! interface, with pallet-api the runtime implementation, with submodules added over time for each use case.

extension/src/functions.rs
```diff
@@ -0,0 +1,136 @@
...
+	) -> Result<RetVal> {
+		// Attempts to match a specified extension/function identifier to its corresponding function, as configured by the runtime.
+		for_tuples!( #(
+            if Tuple::matches(&env) {
+                return Tuple::execute(env)
+            }
+        )* );
+
+		// Otherwise returns error indicating an unmatched request.
+		Err(pallet_contracts::Error::<Self::Config>::NoChainExtension.into())
```
evilrobot-01, line 37: Same error raised when no active chain extension is found: https://github.com/paritytech/polkadot-sdk/blob/f239abac93b8354fec1441e39727b0706b489415/substrate/frame/contracts/src/wasm/runtime.rs#L2258

Might be okay to start, probably should be configurable.

extension/src/lib.rs
```diff
@@ -1,324 +1,109 @@
...
-	fn contains(t: &Self::Call) -> bool;
-}
-
-/// Pop API chain extension.
+/// A configurable chain extension.
 #[derive(Default)]
-pub struct ApiExtension<I>(PhantomData<I>);
-
-impl<T, I> ChainExtension<T> for ApiExtension<I>
+pub struct Extension<C: Config>(PhantomData<C>);
```
evilrobot-01, line 31: Alternative is Tin's implementation, which implements the Config trait on the runtime. The benefit is that you dont have an additional type, but I feel this is logically better, as we are configuring the chain extension plugged into pallet-contracts, plugged into the runtime. It keeps it encapsulated/logically distinct.

pallets/api/src/extension.rs
```diff
@@ -0,0 +1,54 @@
+use core::marker::PhantomData;
```
evilrobot-01, line 1: This modules aims to store anything that is both specifc to pop-api and reusable for chain extensions implementations across runtimes.

e.g.
[contract envinroment: contract <> `pop-api` ] <----> [runtime: `pop-chain-extension` <> `pallet-api` <> `fungibles`]

pallets/api/src/extension.rs
```diff
@@ -0,0 +1,54 @@
...
+pub type Extension<Functions> = pop_chain_extension::Extension<Functions>;
+/// Decodes output by prepending bytes from ext_id() + func_id()
+pub type DecodesAs<Output, Error, Logger = ()> = Decodes<Output, Error, Prepender, Logger>;
+
+/// Prepends bytes from ext_id() + func_id() to prefix the encoded input bytes to determine the versioned output
+pub struct Prepender;
+impl Processor for Prepender {
+	const LOG_TARGET: &'static str = "pop-api::extension::processor";
+
+	fn process<E: Ext, S: State>(value: &mut Vec<u8>, env: &mut Environment<E, S>) {
```
evilrobot-01, line 19: This encoding scheme is specific to pop-api, hence stored here.

runtime/devnet/src/config/api.rs
```diff
@@ -1,68 +1,545 @@
...
+		}
+	}
+
+	pub struct DecodingFailedError;
+	impl ErrorProvider for DecodingFailedError {
+		const ERROR: DispatchError = error::DECODING_FAILED_ERROR;
+	}
+
+	// TODO: refactor to move components to pallet-api
+	mod error {
```
evilrobot-01, line 173: This still needs work...

runtime/devnet/src/config/contracts.rs
```diff
@@ -64,7 +64,7 @@ impl pallet_contracts::Config for Runtime {
 	type CallStack = [pallet_contracts::Frame<Self>; 23];
 	type WeightPrice = pallet_transaction_payment::Pallet<Self>;
 	type WeightInfo = pallet_contracts::weights::SubstrateWeight<Self>;
-	type ChainExtension = pop_chain_extension::ApiExtension<Extension>;
+	type ChainExtension = api::Extension<Config>;
```
evilrobot-01, line 67: Defers to the extension defined in the local api module.

extension/src/functions.rs
Daanvdplas: This prevents adding the `UnknownCall`. In the end it is a u32 on the contract side which is turned into a clear error. Docs can support what this error means. It prevents the situation where we end up wanting to remove the previously used `UnknownCall` so sounds like a good plan.

extension/src/lib.rs
Daanvdplas: I agree that having a chain extension that implements the Config trait makes more sense.

extension/src/functions.rs
Daanvdplas: Regarding this point, there is also [DecodingFailed](https://github.com/paritytech/polkadot-sdk/blob/717bbb24c40717e70d2e3b648bcd372559c71bd2/substrate/frame/contracts/src/lib.rs#L1244).

Daanvdplas (commented):
Looking good to proceed. We might be able to not use any additional error variant other than `Other` with your proposed change to using the pallet contracts error for `NoChainExtensionId`. This prevents us wanting to delete it in the future.

extension/src/functions.rs
evilrobot-01: Changed to `DecodingFailed` in latest commit as the error description is roughly closer to the reason for the error. Note also that we have DecodingFailed in the contract-side error, with an explicit mapping to this variant when it occurs in the runtime keeping everything aligned.

In reality, as long as devs are using the pop-api crate, they should never experience this error - only those who are implementing the API.

extension/src/reboot.rs
evilrobot-01: @Daanvdplas can this comment be resolved?

Daanvdplas: yes, apologies, resolved

pallets/api/src/extension.rs
```diff
@@ -0,0 +1,121 @@
...
+	/// # Parameters
+	/// - `value` - The value to be converted.
+	/// - `env` - The current execution environment.
+	fn convert<E: Ext, S: State>(value: Self::Source, env: &mut Environment<E, S>) -> Self::Target {
+		// Defer to supplied versioned result conversion type
+		let version = env.func_id().to_le_bytes()[0];
+		log::debug!(target: Self::LOG_TARGET, "versioned result converter: result={value:?}, version={version}");
+		let converted: Target = (value, version).into();
+		log::debug!(target: Self::LOG_TARGET, "versioned result converter: converted result={converted:?}");
+		converted.into()
```
peterwht, line 119: unnecessary .into (clippy)

peterwht (commented):
Really clean PR!  The new format is great. Left a nitpicks.
Will review again after unit tests are added.

extension/src/lib.rs
```diff
@@ -1,324 +1,159 @@
...
+mod decoding;
+mod functions;
+mod matching;
 #[cfg(test)]
 mod tests;
-mod v0;
 
-/// Logging target for categorizing messages from the Pop API extension module.
-const LOG_TARGET: &str = "pop-api::extension";
+type Schedule<T> = <T as pallet_contracts::Config>::Schedule;
```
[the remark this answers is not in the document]
evilrobot-01: You sure? Still seems to be in master: https://github.com/paritytech/polkadot-sdk/blob/b2ec017c0e5e49f3cbf782a5255bb0f9e88bd6c1/substrate/frame/contracts/src/lib.rs#L333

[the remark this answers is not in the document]
chungquantin: @evilrobot-01 Not the pallet-contract config, but the `type Schedule<T>` is no longer used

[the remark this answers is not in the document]
evilrobot-01: But it is used in this PR?

Sorry but I don't follow, can you provide more context please.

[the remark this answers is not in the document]
Daanvdplas: You guys mean: https://github.com/paritytech/polkadot-sdk/pull/4233/files ?

pallets/api/src/extension.rs
```diff
@@ -0,0 +1,121 @@
...
+		value
+	}
+}
+
+/// Matches on the first byte of a function identifier only.
+pub struct IdentifiedByFirstByteOfFunctionId<T>(PhantomData<T>);
+impl<T: Get<u8>> Matches for IdentifiedByFirstByteOfFunctionId<T> {
+	fn matches<E: Ext, S: State>(env: &Environment<E, S>) -> bool {
+		let bytes = env.func_id().to_le_bytes();
+		bytes[1] == T::get()
```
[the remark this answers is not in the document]
evilrobot-01: This is dependent on the redefinition of the encoding scheme used which was to be dealt with separately to this PR.

Agree that the current naming is not representative of what it is actually doing, was more a case of demonstrating a descriptive name for the type to make it easier to follow what is going on when it is used higher up the stack.

@Daan - can you prepare an internal draft writeup of how it will work for alignment please?  I can then amend here accordingly, with that encoding change happening separately to this PR.

extension/src/lib.rs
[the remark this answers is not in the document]
chungquantin: @Daanvdplas Yes, https://github.com/r0gue-io/pop-node/blob/8072a479a3de951e560a7393e30124f91eede3a4/runtime/devnet/src/extensions.rs#L128C2-L130C60

So we need to rebase first before merging to main.

extension/src/decoding.rs
```diff
@@ -0,0 +1,55 @@
...
+
+	/// Decodes data read from contract memory.
+	///
+	/// # Parameters
+	/// - `env` - The current execution environment.
+	fn decode<E: Ext, S: BufIn>(env: &mut Environment<E, S>) -> Result<Self::Output> {
+		// Charge appropriate weight, based on input length, prior to decoding.
+		// reference: https://github.com/paritytech/polkadot-sdk/blob/117a9433dac88d5ac00c058c9b39c511d47749d2/substrate/frame/contracts/src/wasm/runtime.rs#L267
+		let len = env.in_len();
+		let weight = Schedule::<E::T>::get()
```
chungquantin, line 26: Deprecated: https://github.com/r0gue-io/pop-node/blob/8072a479a3de951e560a7393e30124f91eede3a4/runtime/devnet/src/extensions.rs#L128C2-L130C60

extension/src/lib.rs
```diff
@@ -1,324 +1,159 @@
...
-					},
-				}
-			},
-			Err(e) => Err(e),
-		};
-
-		match result {
-			Ok(_) => Ok(RetVal::Converging(0)),
-			Err(e) => Ok(RetVal::Converging(convert_to_status_code(e, version))),
+		env.charge_weight(Schedule::<Runtime>::get().host_fn_weights.debug_message)?;
```
chungquantin, line 49: Different from `main`. https://github.com/r0gue-io/pop-node/blob/8072a479a3de951e560a7393e30124f91eede3a4/runtime/devnet/src/extensions.rs#L128C2-L130C60

chungquantin (commented):
Mindblowing!

extension/src/functions.rs
```diff
@@ -0,0 +1,156 @@
...
+	/// Executes the function.
+	///
+	/// # Parameters
+	/// - `env` - The current execution environment.
+	fn execute<E: Ext<T = Self::Config>, S: BufIn + BufOut>(
+		env: &mut Environment<E, S>,
+	) -> Result<RetVal>;
+}
+
+#[impl_trait_for_tuples::impl_for_tuples(1, 3)]
```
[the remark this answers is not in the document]
evilrobot-01: Added, will be included as a commit in next push.

runtime/devnet/src/config/api/versioning.rs
```diff
@@ -0,0 +1,208 @@
...
+			},
+			CannotLookup => Error::CannotLookup,
+			BadOrigin => Error::BadOrigin,
+			Module(error) => {
+				// Note: message not used
+				let ModuleError { index, error, message: _message } = error;
+				// Map pallet-contracts::Error::DecodingFailed to Error::DecodingFailed
+				if index as usize
+					== <crate::Contracts as frame_support::traits::PalletInfoAccess>::index()
+					&& error == [11, 0, 0, 0]
```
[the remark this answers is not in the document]
evilrobot-01: Agreed, addressed in commit along with a test.

pallets/api/src/extension.rs
evilrobot-01: Thank you!

extension/src/lib.rs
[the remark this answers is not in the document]
evilrobot-01: Marked the type alias as deprecated, will be in next push.

extension/src/decoding.rs
evilrobot-01: Marked the type alias as deprecated, will be in next push.

pallets/api/src/extension.rs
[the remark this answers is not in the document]
evilrobot-01: Have refactored these into functions to make it easier to update once encoding scheme finalised.

extension/src/lib.rs
evilrobot-01: The read_state implementation at that reference does not seem to be consistent with the dispatch - is this intentional?

I would expect the weight charging for inputs (reading from contract memory) to be the same for both, with the read_state charging additional weight for the returned value. Am I missing something?

extension/src/lib.rs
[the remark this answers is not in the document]
evilrobot-01: I see now that those functions are already available on the weight info trait so adjusting to match.

extension/src/lib.rs
chungquantin: The changes simply just replace the deprecated approach in `main` (which only has a basic version of the chain extension, no pop-api introduced yet) with the new one in the scope of upgrading to 1.14.0. Hence, I believe these need to be addressed in the API PR.

evilrobot-01: Addressed in latest pushed commit.

extension/src/lib.rs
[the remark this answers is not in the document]
evilrobot-01: Addressed in latest pushed commit.

evilrobot-01:
@peterwht @chungquantin as per your comments above, I addressed them in https://github.com/r0gue-io/pop-node/pull/218/commits/2152867b8894e36a17b6773e1bc3ada1f7966437 and resolved the conversation. Please could you kindly take a look at the commit and let me know if any concerns?

extension/src/decoding.rs
```diff
@@ -50,3 +52,14 @@ impl<
 	type Error = Error;
 	const LOG_TARGET: &'static str = Logger::LOG_TARGET;
 }
+
+/// Default processor implementation which just passes through the value unchanged.
+pub struct IdentityProcessor;
```
chungquantin, line 57: Wonder why is this named `IndentityProcessor`?

evilrobot-01: Not a great name, I admit. It was partially inspired by the [Identity hasher](https://docs.substrate.io/build/runtime-storage/#hashing-algorithms) in Substrate. Any better suggestions? Could just be `Identity` instead.

chungquantin:
1. https://github.com/r0gue-io/pop-node/pull/236
2. https://github.com/r0gue-io/pop-node/pull/237

extension/src/functions.rs
```diff
@@ -0,0 +1,169 @@
...
+		// Perform any final conversion. Any implementation is expected to charge weight as appropriate.
+		let result = ResultConverter::convert(result, env).into();
+		// Charge weight before read
+		let weight = ContractWeights::<Config>::seal_input_per_byte(1); // use unit weight as write function handles multiplication
+		log::trace!(target: Logger::LOG_TARGET, "return result to contract: weight_per_byte={weight}");
+		// Charge appropriate weight for writing to contract, based on input length, prior to decoding.
+		// TODO: check parameters (allow_skip, weight_per_byte)
+		// TODO: confirm whether potential error from writing to the buffer needs to be converted to a versioned error (suspect not)
+		env.write(&result, false, Some(weight))?;
+		Ok(Converging(0))
```
Daanvdplas, line 149: Shouldn't this be:
```
match result {
			Ok(_) => Ok(Converging(0)),
			Err(e) => Error::convert(e.error, env),
		}
		```

**Comment by Daanvdplas** [comment]
As for the open questions regarding whether to convert the `DispatchError` in e.g. the `ReadState` function. My view is that in the end it is also a pallet with an index. We do account for changes in a pallet index, so why not include those as well? I think treating all pallets the same is the easiest and safest.

File: extension/src/decoding.rs (hunk 17 above)
Comment (peterwht) [reply extension/src/decoding.rs:58]: Tbh, I think it's a great name! Quite clear what it does with inspiration from math, and postfixing with Processor makes quite easy to recognize what it's doing when reading tests etc.

File: pallets/api/src/extension.rs [hunk 19]
```diff
@@ -0,0 +1,131 @@
...
+	/// - `value` - The value to be processed.
+	/// - `env` - The current execution environment.
+	fn process(mut value: Self::Value, env: &impl Environment) -> Self::Value {
+		// Resolve version, pallet and call index from environment
+		let version = version(env);
+		let (module, index) = module_and_index(env);
+		// Prepend bytes
+		value.insert(0, version);
+		value.insert(1, module);
+		value.insert(2, index);
```
peterwht, line 41: These insertions would be more efficient with using splice():
```rust
    value.splice(0..0, [index, module, version]);
```

pallets/api/src/extension.rs
```diff
@@ -0,0 +1,131 @@
...
+		let version = version(env);
+		log::debug!(target: Self::LOG_TARGET, "versioned result converter: result={value:?}, version={version}");
+		let converted: Target = (value, version).into();
+		log::debug!(target: Self::LOG_TARGET, "versioned result converter: converted result={converted:?}");
+		converted
+	}
+}
+
+fn func_id(env: &impl Environment) -> u8 {
+	// TODO: update once the encoding scheme order has been finalised: expected to be env.ext_id().to_le_bytes()[0]
```
peterwht, line 118: do we want to merge with this TODO being unresolved?

pallets/api/src/extension.rs
```diff
@@ -0,0 +1,131 @@
...
+	env.func_id().to_le_bytes()[1]
+}
+
+fn module_and_index(env: &impl Environment) -> (u8, u8) {
+	// TODO: update once the encoding scheme order has been finalised: expected to be env.func_id().to_le_bytes()[0..1]
```
peterwht, line 123: here as well

pallets/api/src/extension.rs
```diff
@@ -0,0 +1,131 @@
...
+	let bytes = env.ext_id().to_le_bytes();
+	(bytes[0], bytes[1])
+}
+
+fn version(env: &impl Environment) -> u8 {
+	// TODO: update once the encoding scheme order has been finalised: expected to be env.ext_id().to_le_bytes()[1]
```
peterwht, line 129: and here

runtime/devnet/src/config/api/versioning.rs
```diff
@@ -0,0 +1,216 @@
...
+struct V0Error(pop_primitives::v0::Error);
+impl From<DispatchError> for V0Error {
+	fn from(error: DispatchError) -> Self {
+		use pop_primitives::v0::*;
+		use sp_runtime::{ArithmeticError::*, TokenError::*, TransactionalError::*};
+		use DispatchError::*;
+		// Mappings exist here to avoid taking a dependency of sp_runtime on pop-primitives
+		Self(match error {
+			Other(_message) => {
+				// Note: lossy conversion: message not used due to returned contract status code size limitation
```
peterwht, line 104: Not for this PR, but it would be nice to have a way to write this to the contract buffer for debugging purposes. Maybe we can write to buffer, and then in the contract-side we can read and display when in debug mode only. Needs some more thought.

runtime/devnet/src/config/api/versioning.rs
```diff
@@ -0,0 +1,216 @@
...
+impl From<VersionedError> for u32 {
+	fn from(value: VersionedError) -> Self {
+		match value {
+			VersionedError::V0(error) => error.into(),
+		}
+	}
+}
+
+struct V0Error(pop_primitives::v0::Error);
+impl From<DispatchError> for V0Error {
```
peterwht, line 96: Are we maintaining this DispatchError per the discussion earlier?

runtime/devnet/src/config/api/versioning.rs
peterwht: This may actually not be necessary if we return the DispatchError directly for infrastructure-errors (not pop api  use-cases)

peterwht (commented):
Reviewed again. The architecture is really awesome. Left a few small comments.

extension/src/decoding.rs
evilrobot-01: This was already moved in the tests PR. I did shorten it to `Identity` for simplicity, which implements the `Processor` trait.

extension/src/environment.rs
```diff
@@ -0,0 +1,182 @@
...
+	/// it has no purpose.
+	///
+	/// # Note
+	///
+	/// Weight is synonymous with gas in substrate.
+	fn charge_weight(&mut self, amount: Weight) -> Result<Self::ChargedAmount>;
+
+	/// Adjust a previously charged amount down to its actual amount.
+	///
+	/// This is when a maximum a priori amount was charged and then should be partially
```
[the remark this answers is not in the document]
evilrobot-01: "a priori" - these comments are copied directly from the underlying `Environment` implementation.

runtime/devnet/src/config/api/versioning.rs
evilrobot-01: We always have to maintain `DispatchError` as it is the main error emitted by the runtime. It can potentially evolve over time through runtime upgrades, hence the ability to convert it back to a prior versioned error if required, depending on context.

pallets/api/src/extension.rs
evilrobot-01: As mentioned separately, it was agreed that this would be dealt with separately. I dont believe much has been finalised there yet.

extension/src/decoding.rs
evilrobot-01: As mentioned, this is now available at https://github.com/r0gue-io/pop-node/blob/39f16e1459e75b86d07414d7e3c70e21dc97d8e0/extension/src/decoding.rs#L60 as desired. Resolving this, let me know if anything else required.

runtime/devnet/src/config/api/mod.rs
```diff
@@ -0,0 +1,154 @@
...
+			ReadStateLogTarget,
+		>,
+	);
+
+	/// The log target.
+	const LOG_TARGET: &'static str = LOG_TARGET;
+}
+
+/// Filters used by the chain extension.
+pub struct Filter;
```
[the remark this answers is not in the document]
evilrobot-01: See https://github.com/r0gue-io/pop-node/pull/264

extension/src/lib.rs
```diff
@@ -1,324 +1,229 @@
...
-pub trait CallFilter {
-	/// Query of the dispatch calls operations.
-	type Call: Decode;
+type ContractWeights<T> = <T as pallet_contracts::Config>::WeightInfo;
 
-	/// Check if runtime call is allowed.
-	fn contains(t: &Self::Call) -> bool;
-}
+/// Encoded version of `pallet_contracts::Error::DecodingFailed`, as found within `DispatchError::ModuleError`.
+pub const DECODING_FAILED_ERROR: [u8; 4] = [11, 0, 0, 0];
```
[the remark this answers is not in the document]
evilrobot-01: There are actually no usages within the chain extension crate itself, only the api pallet/runtime. I will move it to the pallet in a commit.

peterwht (approved):

pallets/api/src/extension.rs
evilrobot-01: See https://github.com/r0gue-io/pop-node/pull/265

chungquantin (approved):

pallets/api/src/extension.rs
peterwht: awesome! Those lines were bothering for a couple PRs 😅

PR #219, opened 2024-08-19
chore(runtime): bump versions to 0.3.0

peterwht:


Daanvdplas (commented):
Only testnet runtime needs to be bumped.

peterwht:
> Only testnet runtime needs to be bumped.

why would devnet not be bumped too? Currently devnet and testnet are equivalent.

Daanvdplas (approved):
No need to remove the Migrations struct but other than that all good!

evilrobot-01 (approved):

PR #222, opened 2024-08-21
fix: missing migration

al3mart:
Adds a missing migration necessary for `testnet-runtime` `0.4.0`.
Bumps spec version of testnet-runtime to `00_04_00`


Daanvdplas (approved):
Could you bump the node version too please?

PR #226, opened 2024-08-21
chore: conditional release builds

al3mart:
- Trigger the srtool build flow if the release has been tagged with `runtime-*`.
- Trigger the release node build flow if the release has been tagged with `node-*`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,7 +5,7 @@ env:
 
 on:
   release:
-    types: [ published ]
+    types: [published]
   workflow_dispatch:
     inputs:
       ref:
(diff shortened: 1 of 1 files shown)


al3mart:
Testing these changes triggering a the action to be run with:
`gh workflow run "pop release" --ref node-conditional-test-release`

The runtime build seems to be triggered anyway.

evilrobot-01:
Need to update to differentiate between the runtimes, so we can do just a testnet release (`testnet-*` tag), or just a mainnet release (`mainnet-*` tag). This should avoid all the noise which appears in the releases.

peterwht:
Closing as it's outdated.

PR #228, opened 2024-08-22
ci: add CI script for `pop-api` tests and integration tests

chungquantin:
Link to issue: https://github.com/r0gue-io/pop-node/issues/109

- Build contracts listed under `integration-tests/contracts` when running `cargo build --features integration-tests`
- CI to run integration tests of contracts


evilrobot-01:
Hmmm, could we not have solved this with a `build.rs` that takes a dependency on `cargo-contract` to build the relevant contracts?

chungquantin:
@evilrobot-01 Thanks for the feedback. I have redesigned the CI to simplify the script and move the contract build process to the `integration-tests` crate build script. Please take a look and let me know if there are any needed changes. 

cc: @Daanvdplas

evilrobot-01:
Gave it a quick test and it failed as I did not specify the 'contracts' feature. Is there any reason that this needs to be specified explicitly? The goal is that we can just run cargo test in the integration tests folder and the build script will ensure that the contracts are built, as the integration tests are reliant on them being built.

Running it again with the feature then worked, but it took a long time on 'building' after having cleared out the caches of all contracts and integration tests. Would outputting something to the console help here possibly? Just a message stating 'building xxx contract...' for each should be enough.

peterwht:
Great job, this will be quite useful! It would be a nice addition to have all .wasms get outputted to an `artifacts` folder. That way we can persist the same artifacts without requiring a rebuild of the contracts in CI.

chungquantin:
@peterwht Contracts are archived as artifacts
@evilrobot-01 Remove `contract` feature, contracts are built whenever we run `cargo test` in the `integration-tests` folder

evilrobot-01:
Rebuilding contracts each time should be instant if no changes and ensures that any local contract changes are always included when running tests, improving our devex flow.

In the past we had to know to run all these extra commands, now we just make changes and run the tests and everything 'just works'

pop-api/integration-tests/build.rs
```diff
@@ -0,0 +1,36 @@
...
+				if path.is_dir() {
+					directories.push(path);
+				}
+			}
+		}
+	} else {
+		panic!("Failed to read contracts directory: {:?}", contracts_dir);
+	}
+	directories
+}
```
Daanvdplas, line 36:
```suggestion
use contract_build::{execute, BuildMode, BuildResult, ExecuteArgs, ManifestPath, Verbosity};
use std::{
	fs,
	path::{Path, PathBuf},
	process,
};

fn main() {
	let contracts_dir = PathBuf::from("./contracts/");
	let contract_dirs = match get_subcontract_directories(&contracts_dir) {
		Ok(dirs) => dirs,
		Err(e) => {
			eprintln!("Failed to read contracts directory: {}", e);
			process::exit(1);
		},
	};

	for contract in contract_dirs {
		if let Err(e) = build_contract(&contract) {
			eprintln!("Failed to build contract {}: {}", contract.display(), e);
			process::exit(1);
		}
	}
}

// Function to retrieve all subdirectories in a given directory.
fn get_subcontract_directories(contracts_dir: &Path) -> Result<Vec<PathBuf>, String> {
	fs::read_dir(contracts_dir)
		.map_err(|e| format!("Could not read directory '{}': {}", contracts_dir.display(), e))?
		.filter_map(|entry| match entry {
			Ok(entry) if entry.path().is_dir() => Some(Ok(entry.path())),
			Ok(_) => None,
			Err(e) => Some(Err(format!("Error reading directory entry: {}", e))),
		})
		.collect()
}

// Function to build a contract given its directory.
fn build_contract(contract_dir: &Path) -> Result<BuildResult, String> {
	let manifest_path = ManifestPath::new(contract_dir.join("Cargo.toml")).map_err(|e| {
		format!("Could not retrieve manifest path for {}: {}", contract_dir.display(), e)
	})?;

	let args = ExecuteArgs {
		build_mode: BuildMode::Release,
		manifest_path,
		verbosity: Verbosity::Quiet,
		..Default::default()
	};

	execute(args).map_err(|e| format!("Build failed for {}: {}", contract_dir.display(), e))
}
```

Daanvdplas (approved):
Thanks a lot Tin! Provided a suggested refactor.

pop-api/integration-tests/build.rs
chungquantin: Thanks Daan, looks clean!

.github/workflows/ci.yml
```diff
@@ -67,6 +67,40 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: "./.github/actions/init"
+
+      - name: Run integration tests
+        working-directory: pop-api/integration-tests
+        run: cargo test --release
+
+      - name: Copy contract WASM files to artifacts folder
```
evilrobot-01, line 92: Not sure if I see the point in this. Is it just so that we can download them and put them in the right place on dev machines to avoid a build? Seems a hassle if so and probably something that will go unused, especially as the integration tests build script will sort that automatically. I see more value in always have the latest contract wasm on a dev machine so integration tests work accordingly.

.github/workflows/ci.yml
```diff
@@ -67,6 +67,40 @@ jobs:
...
+        working-directory: pop-api/integration-tests
+        run: |
+          mkdir -p artifacts
+          cp contracts/*/target/ink/*.wasm artifacts/
+
+      - name: Archive API contract artifacts
```
evilrobot-01, line 98: As above, interested to hear the use case for this. If the build jobs are in parallel on the same worker, fear this may just slow the overall process down for something that not many will use.

pop-api/integration-tests/build.rs
```diff
@@ -0,0 +1,52 @@
...
+// Function to build a contract given its directory.
+fn build_contract(contract_dir: &Path) -> Result<BuildResult, String> {
+	let manifest_path = ManifestPath::new(contract_dir.join("Cargo.toml")).map_err(|e| {
+		format!("Could not retrieve manifest path for {}: {}", contract_dir.display(), e)
+	})?;
+
+	let args = ExecuteArgs {
+		build_mode: BuildMode::Release,
+		manifest_path,
+		verbosity: Verbosity::Quiet,
```
evilrobot-01, line 47: It might be useful to see the output, otherwise one is left left waiting. Otherwise a console log message to state that each contract is being built should be better than it just 'hanging' while it builds the contracts.

I expect this will get longer in the future as more contracts are added, so might as well finish it off nicely now so we dont have to come back to it.

evilrobot-01 (commented):
Looks fine and works as is. I personally dont see the need for storing the artifacts to would be interested to see how they will actually be used.

I still think that a console log per contract build would be better than nothing whatsoever, especially as we expect many more test contracts to be added in the near term.

.github/workflows/ci.yml
peterwht: this is valid. I take back what I said about storing the artifacts.

pop-api/integration-tests/build.rs
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// Function to build a contract given its directory.
+fn build_contract(contract_dir: &Path) -> Result<BuildResult, String> {
+	let manifest_path = ManifestPath::new(contract_dir.join("Cargo.toml")).map_err(|e| {
+		format!("Could not retrieve manifest path for {}: {}", contract_dir.display(), e)
+	})?;
+
+	let args = ExecuteArgs {
+		build_mode: BuildMode::Release,
```
evilrobot-01, line 45: Note that this prevents anything logged via `ink::env::debug_println!` from appearing in test output logs (when enabled).

.github/workflows/ci.yml
chungquantin: I think right now we don't have any specific use case for the contract artifacts yet. So would be better to simply the CI script by removing this job.

chungquantin:
Update: 
- Remove archiving artifacts in the CI script
- Build mode is changed to `Debug` mode instead so we can track the debug message of the ink! contract.
- Build artifact is added with `CodeOnly` mode because we only need `.wasm` file for the integration tests, no metadata needed.

cc: @evilrobot-01 @peterwht @Daanvdplas

Daanvdplas (approved):
LGTM!

al3mart (approved):

PR #229, opened 2024-08-22
style: extend pre-push githook

al3mart:
Adds checks for toml and features formatting.

Leaving feature propagation out of the hook, but could be added if there's any interest.


evilrobot-01:
@al3mart can you resolve conflicts so we can assess this for inclusion please?

al3mart:
@evilrobot-01 This should be ready for review now

Daanvdplas (commented):
If we do zepter, lets do all:
```
zepter lint propagate-feature --feature try-runtime --left-side-feature-missing=ignore --workspace --feature-enables-dep="try-runtime:frame-try-runtime" --locked
zepter lint propagate-feature --feature runtime-benchmarks --left-side-feature-missing=ignore --workspace --feature-enables-dep="runtime-benchmarks:frame-benchmarking" --locked
zepter lint propagate-feature --feature std --left-side-feature-missing=ignore --workspace --locked
```

Not super happy with zepter after having used it yesterday and today. Just a reminder that it won't always fix your missing features and still requires manual checking

peterwht (approved):

peterwht:
Nice! One comment not directly related to this PR. It would be nice to have an internal wiki for setting up your local environment for pop-node development.

al3mart:
@Daanvdplas I didn't include the feature propagation checks as I was just going after formatting.

I also exclude it with the following in mind:

If for whatever reason one has commented out one crate out of a certain feature propagation:
```
my-feature = [
    #"pop-runtime-devnet/my-feature",
    "pop-runtime-testnet/my-feature",
]
```

And now I run zepter checks for feature propagation on `my-feature` it will prompt me with something like:

```
crate 'pop-node'
  feature 'my-feature'
    must propagate to:
      pop-runtime-testnet
Found 1 issue (run with `--fix` to fix).
```

But if I provide the fix via zeter then we end up pushing something that would look like:
```
my-feature = [
    #"pop-runtime-devnet/my-feature",
    "pop-runtime-testnet/my-feature",
    "pop-runtime-devnet/my-feature",
]
```

It will correctly add what is missing, but would effectively make the code base a bit messier, so I believe having the CI failing would be better than pushing code the passes the lint checks but is not what we want to have in the repo ultimately.

That's why I think we should not add the propagation checks in this hook.

Daanvdplas:
The above doesn't fix those missing feature propagations, they just emit a warning, just like what you are adding.

Daanvdplas (commented):
Can you add the relevant resources in `.githooks/README` pls?

Whenever a format check fails it might be good to reference to something like the above README maybe?

Daanvdplas:
Like I said, can we have some reference to this README? I think it would be could to add in the hook. If anything of the above fails, `echo "please check the ...."`

al3mart:
The hook needs to be manually installed locally first, to find those instructions users need to read this concrete README first.

Hook installing instructions are placed in the document after a mention to install each dependency that the hook will run, I think it is fair to assume users will install deps before installing the hook.

Hopefully the current addition of the two extra checks is a happier path.

evilrobot-01:
> The hook needs to be manually installed locally first, to find those instructions users need to read this concrete README first.

Exactly. If one wants to use the hook to make their lives easier, they follow the instructions and install the dependencies. However, there is a very small risk that maybe something stops working, a dependency is uninstalled for some reason and then there is some opaque error. A mention of the readme could be useful here.

My concern here is performance. A test push on this repo results in about a 15 sec lag before failing, presumably due to the additional checks being run. How often are changes made to manifests versus Rust files. Is it tolerable to occasionally have CI fail rather than running these checks every time, potentially needlessly. Are we solving a problem that rarely happens, rather than feeling some pain and identifying that it needs resolving.

Daanvdplas:
First, thanks Ale for the last changes you made!

> My concern here is performance. A test push on this repo results in about a 15 sec lag before failing, presumably due to the additional checks being run. How often are changes made to manifests versus Rust files. Is it tolerable to occasionally have CI fail rather than running these checks every time, potentially needlessly. Are we solving a problem that rarely happens, rather than feeling some pain and identifying that it needs resolving.

Lets put this PR on halt then, see how the current CI setup works, and come back to this PR when we see a need for it.

@al3mart @evilrobot-01

evilrobot-01 (approved):
Looks good, thanks!

As mentioned elsewhere, we did indeed encounter some issues around manifest formatting recently so I admit I was wrong and that this will be useful. I am still concerned about the time it takes, but perhaps that just means its time for me to upgrade. :)

I'm not against trialling it to see if it is tolerable, as at least that way we have experience in using it.

PR #231, opened 2024-08-23
feat: pallet-revive

chungquantin:
- The integration might need to update pop-node to 1.15.0 because pallet-revive is added in the next release. Reference to upgrade pop-node to stable2407, compatible with pallet-revive: https://github.com/r0gue-io/base-parachain/pull/45
- Until the next version of the pallet-contracts (migrated to PolkaVM), we will just fork the crate and play with it on local.


peterwht:
Just opened https://github.com/r0gue-io/pop-node/pull/319 with the latest version of pallet-revive. I would say this can be closed now!

PR #235, opened 2024-08-26
fix: mutable `self` required for dispatchable ink! methods

chungquantin:
In our current `pop-api` contracts, we do not mark `self` as `mutable` because the contract storage is not being modified directly; instead, the runtime storage is mutated by the `pop-api`. However, on the client side, when using libraries that support ink! (such as PJS or Dedot), these methods are perceived as `query` rather than `tx`, as the contract itself does not undergo changes.

While it's unclear whether `pop-api` will be utilized in immutable methods in production, it is crucial to clearly state this requirement in the documentation. This will help ensure that developers understand how pop-api behaves and how it might impact the perception and execution of methods in client-side libraries.

cc: @Daanvdplas @peterwht

### DIFF
--- a/pop-api/examples/fungibles/lib.rs
+++ b/pop-api/examples/fungibles/lib.rs
@@ -60,13 +60,13 @@ mod fungibles {
 		}
 
 		#[ink(message)]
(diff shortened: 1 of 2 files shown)


Daanvdplas (approved):
Nice find!

PR #237, opened 2024-08-27
refactor: add extension to integration tests

chungquantin:
- Separate contract-related tests from the extension crate to the `tests` folder


evilrobot-01:
I'm sorry but I'm not onboard with this. Its taking something that was compartmentalised into its own discrete thing and spreading it out across the codebase, making it more tightly coupled to our repo and less reusable imo. Similarly, there is the concept of PopApiExtensionConfig in the tests. The whole point was to make this generic, so that it had nothing to do with Pop's API. The API is a realisation of using the generic chain extension, its a specialisation a layer above.

The time would have been better spent focusing on testing the actual functionality of the extension, which is far more important than where the tests are located. Lets get it fully tested, make sure its suitable for usage, sufficiently flexible and easily testable, so that it can be merged and unblock things. These sorts of things can be better assessed later.

PR #240, opened 2024-08-28
test: generic extension unit testing

chungquantin:
Test suite for modules in the generic extension crate: 

Screenshot 2024-08-30 at 00 07 06


extension/src/tests/contract.rs
```diff
@@ -0,0 +1,109 @@
...
+			// `set_code` requires root origin, expect throwing error.
+			RuntimeCall::System(Call::set_code { code: "pop".as_bytes().to_vec() }),
+			GAS_LIMIT,
+		);
+		assert_eq!(call.result.err(), Some(BadOrigin))
+	})
+}
+
+#[test]
+fn read_state_works() {
```
chungquantin, line 58: Error thrown

extension/src/mock.rs
```diff
@@ -4,13 +4,42 @@ use crate::{
...
+	// The runtime state reads available.
+	RuntimeRead,
+	// Decode inputs to the function as runtime calls
+	Decodes<RuntimeRead, DecodingFailed<Test>, RemoveFirstByte>,
+	// Accept any filtering
+	Filter,
+	// Convert the result of a read into the expected result
+	DefaultConverter<RuntimeResult>,
+>;
+
```
Daanvdplas, line 42: I like this. Can you use this in `type Functions` in this file and use it for all test cases?

chungquantin: Yes it is currently used.

extension/src/functions.rs
```diff
@@ -133,6 +133,32 @@ impl<M: Matches, C, R, D, F, RC, E, L> Matches for ReadState<M, C, R, D, F, RC,
 	}
 }
 
+/// A function that filters the provided input. Only for testing purpose.
+/// Returns supplied contract value as 1 if allowed, 0 if not allowed.
+struct InputFilter<M, C, F>(PhantomData<(M, C, F)>);
```
Daanvdplas, line 138: Thank to the types you create in mock.rs (see other comment), why create an entire new function?

Creates a lot of noise and a lot more separate tests. Perhaps we can refactor the existing tests and use some kind of `for` loop where we provide the filter and execute all the same tests. Like you did in `decoding` with all the enums. 

The tests on itself are not big so I rather have extensive tests in stead of a lot of small ones

extension/src/mock.rs
Daanvdplas: Ah shoot, nice!

extension/src/functions.rs
```diff
@@ -222,27 +248,54 @@ impl<Runtime: pallet_contracts::Config> Function for Tuple {
...
+	use sp_core::ConstU32;
+
+	enum AtLeastOneByte {}
+	impl Contains<Vec<u8>> for AtLeastOneByte {
+		fn contains(input: &Vec<u8>) -> bool {
+			input.len() > 0
+		}
+	}
+
+	type TestFunctions = (
```
chungquantin, line 265: @Daanvdplas Even though you already told me these might not be necessary, I think it is cleaner to have it for testing purpose because we only want to check the Filtering and the execution of function with filtered functions. Let me know your thoughts on this. Already simplify the code.

extension/src/decoding.rs
```diff
@@ -54,13 +89,217 @@ impl<
...
+		let env = MockEnvironment::default();
+		let result = RemoveFirstByte::process(vec![0, 1, 2, 3, 4], &env);
+		assert_eq!(result, vec![1, 2, 3, 4])
+	}
+
+	#[test]
+	fn decode_with_identity_processor_works() {
+		test_cases().iter().for_each(|t| {
+			let (input, output) = (t.clone().0, t.clone().1);
+			println!("input: {:?} -> output: {:?}", input, output);
```
evilrobot-01, line 133: Love the output logging

extension/src/functions.rs
```diff
@@ -17,30 +17,6 @@ pub trait Function {
...
-
-		// Otherwise returns error indicating an unmatched request.
-		log::error!("no function could be matched");
-		Err(pallet_contracts::Error::<Self::Config>::DecodingFailed.into())
-	}
-}
-
 /// A function for dispatching a runtime call.
 pub struct DispatchCall<M, C, D, F, E = (), L = ()>(PhantomData<(M, C, D, F, E, L)>);
 impl<
```
evilrobot-01, line 22: No coverage for the error case on line 68 when running tests in module only.

extension/src/functions.rs
```diff
@@ -167,3 +173,223 @@ impl<T: Into<Vec<u8>>> Converter for DefaultConverter<T> {
...
+
+	enum MustBeEven {}
+	impl Contains<u8> for MustBeEven {
+		fn contains(input: &u8) -> bool {
+			*input % 2 == 0
+		}
+	}
+
+	#[test]
+	fn filterting_works() {
```
evilrobot-01, line 260:
```suggestion
	fn filtering_works() {
```

extension/src/functions.rs
```diff
@@ -167,3 +173,223 @@ impl<T: Into<Vec<u8>>> Converter for DefaultConverter<T> {
...
+		fn contains<C: Contains<T>, T>(input: T, expected: bool) {
+			assert!(C::contains(&input) == expected);
```
evilrobot-01, line 262: assert_eq

extension/src/functions.rs
```diff
@@ -167,3 +173,223 @@ impl<T: Into<Vec<u8>>> Converter for DefaultConverter<T> {
...
+			// Insert one byte due to the `RemoveFirstByte` conversion configuration of `DispatchCallEverythingFuncId`.
+			let encoded_call = [0u8.encode(), call.encode()].concat();
+			let mut default_env = MockEnvironment::default();
+			// Initialize mock environment with function execution.
+			let mut env = MockEnvironment::new(
+				DispatchCallEverthingFuncId::get(),
+				encoded_call.clone(),
+				MockExt::default(),
+			);
+			assert_eq!(env.charged(), default_env.charged());
```
evilrobot-01, line 291: As above, please separate charging weight into its own test

extension/src/functions.rs
```diff
@@ -167,3 +173,223 @@ impl<T: Into<Vec<u8>>> Converter for DefaultConverter<T> {
...
+			// Invalid encoded runtime call.
+			let input = vec![0, 0, 99];
+			let mut env = MockEnvironment::new(
+				DispatchCallEverthingFuncId::get(),
+				input.clone(),
+				MockExt::default(),
+			);
+			let error = pallet_contracts::Error::<Test>::DecodingFailed.into();
+			let expected = <() as ErrorConverter>::convert(error, &mut env).err();
+			assert_eq!(Functions::execute(&mut env).err(), expected);
```
evilrobot-01, line 318: As above, please separate charging weight into its own test

extension/src/functions.rs
```diff
@@ -167,3 +173,223 @@ impl<T: Into<Vec<u8>>> Converter for DefaultConverter<T> {
...
+		contains::<AtLeastOneByte, Vec<u8>>(vec![1], true);
+		contains::<AtLeastOneByte, Vec<u8>>(vec![1, 2, 3, 4], true);
+		contains::<LargerThan100, u8>(100, false);
+		contains::<LargerThan100, u8>(101, true);
+		contains::<MustBeEven, u8>(100, true);
+		contains::<MustBeEven, u8>(101, false);
+	}
+
+	#[test]
+	fn execute_dispatch_call_function_works() {
```
evilrobot-01, line 278: Whilst you have a unit test for checking that the filter works, there are no tests to ensure that filtering within the functions works. 

Each behaviour of the function should have a different test. For example, for the dispatch function
- decodes
- charges weight
- filters
- dispatches
- adjusts weight
- converts error

Same should apply for the other function.

extension/src/functions.rs
```diff
@@ -167,3 +173,223 @@ impl<T: Into<Vec<u8>>> Converter for DefaultConverter<T> {
...
+					encoded_call.len() as u32,
+					call,
+					env.ext().address().clone()
+				)
+			);
+		})
+	}
+
+	#[test]
+	fn execute_dispatch_call_function_invalid_input_works() {
```
evilrobot-01, line 307:
```suggestion
	fn execute_dispatch_call_function_invalid_input_fails() {
```

extension/src/functions.rs
```diff
@@ -167,3 +173,223 @@ impl<T: Into<Vec<u8>>> Converter for DefaultConverter<T> {
...
+			assert_eq!(Functions::execute(&mut env).err(), expected);
+			assert_eq!(env.charged(), read_from_buffer(input.len() as u32,));
+		})
+	}
+
+	#[test]
+	fn execute_read_state_function_works() {
+		let read = RuntimeRead::Ping;
+		// Insert one byte due to the `RemoveFirstByte` conversion configuration of `ReadStateEverythingFuncId`.
+		let encoded_read = [0u8.encode(), read.encode()].concat();
```
evilrobot-01, line 327: Why not just create another function variant that doesnt require the additional byte to avoid all the noise in these tests?

evilrobot-01 (commented):
Just a partial review. 

Looks great so far, left some comments.

extension/src/functions.rs
chungquantin: @evilrobot-01  Before we actually have the tests to cover the filtering in the function: https://github.com/r0gue-io/pop-node/pull/240/commits/fcf440b23710b099ec804ef9477bf927f059a15f

But it seems like not very necessary. And the function filtering is does covered in the contract testing. So I am not sure if we want to include the removed tests.

extension/src/tests/contract.rs
```diff
@@ -0,0 +1,164 @@
...
+
+#[test]
+fn noop_function_works() {
+	new_test_ext().execute_with(|| {
+		// Instantiate a new contract.
+		let contract = instantiate(CONTRACT.clone());
+		let noop_result = call(contract, NoopFuncId::get(), (), GAS_LIMIT);
+		// Successfully return data.
+		let return_value = noop_result.result.unwrap();
+		let decoded = <Result<Vec<u8>, u32>>::decode(&mut &return_value.data[..]).unwrap();
```
peterwht, line 146: why does this include in the first byte, while a few tests above it excludes it?

extension/src/tests/contract.rs
```diff
@@ -0,0 +1,164 @@
...
+		assert_eq!(noop_result.storage_deposit, StorageDeposit::Charge(0));
+	})
+}
+
+#[test]
+fn invalid_func_id_fails() {
+	new_test_ext().execute_with(|| {
+		// Instantiate a new contract.
+		let contract = instantiate(CONTRACT.clone());
+		let result = call(contract, INVALID_FUNC_ID, (), GAS_LIMIT);
```
peterwht, line 157: 💜 the const for INVALID_FUNC_ID

extension/src/tests/mod.rs
```diff
@@ -0,0 +1,120 @@
...
+fn extension_call_works() {
+	let input = vec![2, 2];
+	let mut env = MockEnvironment::new(NoopFuncId::get(), input.clone(), MockExt::default());
+	let mut extension = Extension::<Config>::default();
+	assert!(matches!(extension.call(&mut env), Ok(Converging(0))));
+	assert_eq!(env.charged(), overhead_weight(input.len() as u32))
+}
+
+#[test]
+fn extension_for_unknown_function_works() {
```
peterwht, line 34: Change postfix to `_fails` please. If the weight is the thing that works, please adjust the name to something like `charge_weight_for_unknown_function_works"

peterwht (commented):
Amazing work!!! Really great coverage, the small test sizes made reviewing and understanding everything much easier.

Left some comments, but I can tell it's super close to being ready.

extension/src/tests/utils.rs
```diff
@@ -0,0 +1,180 @@
...
+	gas_limit: Weight,
+) -> ContractExecResult<Balance, EventRecord> {
+	log::debug!("call: func_id={func_id}, input={input:?}");
+	let result = Contracts::bare_call(
+		ALICE,
+		contract,
+		0,
+		gas_limit,
+		None,
+		[function_selector("call"), (func_id, input.encode()).encode()].concat(),
```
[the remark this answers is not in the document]
peterwht: Reason being it will help make future tests easier to write.

extension/src/tests/contract.rs
```diff
@@ -0,0 +1,164 @@
...
+	#[test]
+	fn read_state_works() {
+		new_test_ext().execute_with(|| {
+			// Instantiate a new contract.
+			let contract = instantiate(CONTRACT.clone());
+			// Successfully return data.
+			let read_result =
+				call(contract, ReadContractFuncId::get(), RuntimeRead::Ping, GAS_LIMIT);
+			let return_value = read_result.result.unwrap();
+			let decoded = <Result<Vec<u8>, u32>>::decode(&mut &return_value.data[1..]).unwrap();
```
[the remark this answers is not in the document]
evilrobot-01: That still needs to be solved - the same exists in all all the integration tests which use ExecReturnValue. I tried to do some debugging within contracts internals but wasnt able to hit a breakpoint in the right place. A guess might be that it is wrapped in an Ok(..)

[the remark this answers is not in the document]
chungquantin: @evilrobot-01 As you added that first byte, could you help me to explain why does we need that? When I debug the code, saw the first extra byte but not very sure where it is coming from.

extension/src/tests/mod.rs
evilrobot-01: Agree with the postfix, but might be more useful to start tests that logically belong to the same component with the same prefix, so remain somewhat together when sorted. It might be more useful to see all tests for a function grouped together before then the test case and result in the test name.

extension/src/tests/utils.rs
```diff
@@ -0,0 +1,180 @@
...
+
+/// Instantiating the contract.
+pub(crate) fn instantiate(contract: Vec<u8>) -> AccountId {
+	let result = Contracts::bare_instantiate(
+		ALICE,
+		0,
+		GAS_LIMIT,
+		None,
+		Code::Upload(contract),
+		function_selector("new"),
```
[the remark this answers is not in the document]
evilrobot-01: I dont see the need. This contract only passes through data to the chain extension and making such a change means that every usage of instantiate now needs an additional parameter.

extension/src/tests/utils.rs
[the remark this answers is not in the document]
evilrobot-01: As above, the whole point of the way the contract is written is to take a func id and bytes, so its as dumb as possible. It acts as a conduit to the chain extension which is under test. Adding a "call" parameter value to every usage means a whole lot of extra code which is unnecessary. We can refactor when there is a need for another, there isnt now.

extension/src/tests/contract.rs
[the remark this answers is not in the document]
evilrobot-01: It is simply a side effect of passing some bytes to the contract, which then passes them as a parameter to the chain extension, which is scale encoding them.

https://github.com/r0gue-io/pop-node/blob/e36c83327dc23bb3c5f5cda7df3a3c3d79822ca6/extension/contract/lib.rs#L28

Ordinarily the contract (api) would be constructing the input using the actual types, so the encoding should be correct and not require any pre-processing before decoding.  Put another way, its a work around only applicable to the testing the chain extension via that contract, which does nothing more than pass through the inputs and outputs, hence the name.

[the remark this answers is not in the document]
evilrobot-01: To my first comment above, the same exists in https://github.com/r0gue-io/pop-node/blob/e36c83327dc23bb3c5f5cda7df3a3c3d79822ca6/pop-api/integration-tests/src/fungibles/utils.rs#L10.

extension/src/tests/utils.rs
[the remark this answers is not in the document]
evilrobot-01: Reverted this change in my PR, it was a whole layer of indirection for something which will probably never be required, adding to code churn and more code to review.

extension/src/tests/contract.rs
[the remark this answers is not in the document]
evilrobot-01: Created an issue at https://github.com/r0gue-io/pop-node/issues/263

extension/src/tests/contract.rs
evilrobot-01: Confirmed. The Noop function called does nothing and just returns `Ok(RetVal::Converging(0))` so there is no data written.

peterwht (approved):

PR #244, opened 2024-08-29
feat(mainnet): send fees to PotStake account, update ED and fees to match system chains

peterwht:


evilrobot-01 (changes requested):
A few minor things I think.

runtime/mainnet/src/lib.rs
```diff
@@ -366,15 +370,14 @@ impl pallet_balances::Config for Runtime {
 parameter_types! {
 	/// Relay Chain `TransactionByteFee` / 10
 	pub const TransactionByteFee: Balance = fee::TRANSACTION_BYTE_FEE;
+	pub SudoAddress: AccountId = AccountId::from_ss58check("15NMV2JX1NeMwarQiiZvuJ8ixUcvayFDcu1F9Wz1HNpSc8gP").unwrap();
```
evilrobot-01, line 373: Can we update https://github.com/r0gue-io/pop-node/blob/da62fc26c33975e7bcc73ee87ddbb1642f76dc05/node/src/chain_spec.rs#L385 and https://github.com/r0gue-io/pop-node/blob/da62fc26c33975e7bcc73ee87ddbb1642f76dc05/node/src/chain_spec.rs#L213 to use this value please?

evilrobot-01 (changes requested):
Just one last improvement please, but an important one.

PR #250, opened 2024-08-30
refactor: extension unit tests

Daanvdplas:
Refactored the tests so that all the functionality is tested in a more accumulative way. Starting as small as possible and then bubbling up, also reusing helper functions for weight that help readers understand how it works.


chungquantin (approved):
This is clean! Thanks for your PR.

Issue #251, opened 2024-08-30
feat: make drink generic to support pop-api


Daanvdplas:
We want to offer the developers a test suite for interacting with the pop chain extension. Luckily there is drink!.

# Goal

Create an example contract (preferably with some logic around the api to see how what the devex is like - e.g. `Error`) using drink! for testing.

# Steps

- [x] Fork the drink! repository with the [following changes](https://github.com/Daanvdplas/drink) to make drink! mostly generic.
- [x] Move the required code from [ink sandbox](https://github.com/use-ink/ink/tree/master/crates/e2e/sandbox) (e.g. the `Sandbox trait) to the forked drink repository. This is not dependent of any ink code, see `Cargo.toml`, so this should be straightforward. There are some useful macros that can be used as well.
- [x] Now in the pop-node repo we want to create a `PopSandbox` ([here you can see how](https://github.com/Daanvdplas/fungibles_api_example/blob/main/lib.rs)) which the developers can then use for their drink tests. One thing that I haven't tested yet is including the chain extension. The code says it handles it and there is also an[ example ](https://github.com/Daanvdplas/drink/tree/main/examples)in the drink repository, so this should be doable.
- [x] When this all works we want to implement the example contract. This contract will be a very important exercise. Not only will we test the api, the error handling, the test suite, exposure to ink!, but also will be used for the first resource (most probably tutorial) for the pop api.

chungquantin:
Created a new private repository (will be converted to public later) for pop-drink: https://github.com/r0gue-io/pop-drink

PR #255, opened 2024-09-02
chore(devnet): make drink generic over runtime

chungquantin:
Link to issue: #251 

> [!NOTE]
> Design consideration: Successfully create a `PopSandbox` located under the folder `pop-sandbox` (Considered refactoring in the future for a reasonable structure). 

> [!Caution]
> Why do we need a separate `pop-sandbox` instead of using `create_sandbox!` macro to construct the runtime and sandbox for it? The approach was [implemented before](https://github.com/r0gue-io/pop-node/pull/255/commits/95bd23d8c1eac8ca4c5604be13f9e389e12c6086#diff-cea0a528709440338a1ccbd5b0f2752ae2df1c32e1b6c1488afbe8155997210d:~:text=create_sandbox!(PopDevnetSandbox%2C%20PopRuntime)%3B) but got an issue with the HostFunction [cumulus-primitives-proof-size-hostfunction](https://crates.io/crates/cumulus-primitives-proof-size-hostfunction/0.9.0) when replacing the customized `Sandbox` in the drink test.

Left a few TODOs about the current issues. 
- [ ] (`fix`) A contract that use `Pop API` will throw `DeploymentReverted` or `ContractTrapped`. This perhaps can be an issue with the chain extension. 
- [x] (`chore`) Next task would be rebasing the branch to `daan/api` (1.14.0) and upgrade the `pop-drink` repository to 1.14.0 as well.


chungquantin:
Closed to rebase to v1.14.0

PR #264, opened 2024-09-05
feat: include runtime base call filter in extension call filter

evilrobot-01:
The extension should not allow any calls which are explicitly prohibited by the base call filter defined on the system pallet. An example is all the force_* calls within the balances pallet which we explicitly prevent due to the nature of the token of the chain.

The PR simply calls the base call filter first before then determining whether the call is explicitly denied by the extension configuration.


chungquantin (approved):
LGTM!

peterwht (approved):
Nice and clean 🧼

PR #266, opened 2024-09-06
feat(devnet): runtime non-fungibles extensions

chungquantin:
Implement runtime pallet functionality extensions of the non-fungibles use case

> [!Important]
> Questionable approach in the `set_attribute` method and `ItemAttribute` state query: How can we handle the namespace? Should we dynamically identify the namespace based on the provided origin address? For `ItemAttribute`, should we simplify it by returning all attributes of the collection item or only one single key-value per call.


chungquantin:
Weight benchmarking result for `BalanceOf`. This proves that `owned_in_collection` weight is decided 
```rs
	/// The range of component `i` is `[0, 5000]`.
	fn owned_in_collection(i: u32, ) -> Weight {
		// Proof Size summary in bytes:
		//  Measured:  `0`
		//  Estimated: `0`
		// Minimum execution time: 2_000_000 picoseconds.
		Weight::from_parts(3_905_593, 0)
			// Standard Error: 26
			.saturating_add(Weight::from_parts(469, 0).saturating_mul(i.into()))
	}
```

cc: @Daanvdplas

PR #267, opened 2024-09-06
chore: resolve conflicts for `daan/api`

chungquantin:
Rebase and update the branch `daan/api` to the latest version of the `master` branch.


extension/src/functions.rs
```diff
@@ -350,23 +354,28 @@ mod tests {
 			})
 		}
 
-		#[test]
+#[test]
 		fn dispatch_call_adjusts_weight() {
```
chungquantin, line 358: > [!IMPORTANT]
> Behavior of weight charges in the pallet-contracts is updated which impacts directly to the `dispatch_call_adjust_weight_works` test. 
> - Reference: https://github.com/paritytech/polkadot-sdk/commit/b6231c79ca708d9c9280210f41ca38fd816c8ad9
> - No change to pre-dispatch weight charged

pop-api/integration-tests/src/fungibles/mod.rs
```diff
@@ -569,7 +569,7 @@ fn burn_works() {
 		start_destroy_asset(addr.clone(), asset);
 		assert_eq!(
 			burn(addr.clone(), asset, BOB, amount),
-			Err(Module { index: 52, error: [16, 0] })
+			Err(Module { index: 52, error: [17, 0] })
```
chungquantin, line 572: > [!Note]
> Index of `burn()` method is updated from `16` to `17`

peterwht (approved):
Amazing!! So close to getting this fully merged

extension/src/mock.rs
```diff
@@ -80,31 +82,39 @@ impl pallet_balances::Config for Test {
...
+	type RuntimeEvent = RuntimeEvent;
+	type RuntimeHoldReason = RuntimeHoldReason;
+	type Schedule = MySchedule;
+	type Time = Timestamp;
+	type UnsafeUnstableInterface = ();
+	// UnstableInterface;
+	type UploadOrigin = EnsureSigned<Self::AccountId>;
+	type WeightInfo = ();
+	type WeightPrice = ();
+	// Self;
```
evilrobot-01, line 117:
```suggestion

```

evilrobot-01 (approved):
Seems fine, a lot of noise due to the large difference between main and the target branch. Will be easier to navigate once merged so think we should proceed as is and be more thorough with the final review.

PR #269, opened 2024-09-08
feat: psp22 interfaces

evilrobot-01:
Was suprised to see that there were no trait definitions for the PSP22 standard within the ink! library to allow users to easily apply the standard to their contracts, whilst leveraging the api for the implementation.

This is just a quick prototype, but suggest that it is incorporated (with proper documentation inline with the rest of the crate) and should then be used in the examples. We should also provide an example contract which does exactly this, so anyone can just deploy an instance creating/pointing to an asset identifier. The next asset id added recently will be especially useful here, so all the contract constructor would need is metadata. An additional improvement to the pallet-fungibles would be to make the id parameter for create optional to make it more efficient.


peterwht:
This is awesome to see! Was thinking about this as well. Would be even cooler / better devex if someone can just do `impl PSP22 for MyContract` that uses default implementations to access the runtime. This is more similar to what other ecosystems do.

evilrobot-01:
The ink! trait definition isn't there yet sadly, no associated types and no default impls. Would be awesome if that existed tho, one for the i!A perhaps.

Daanvdplas:
This is definitely something thats need to be added, great job noting it. My suggestion is to finish this PR without the example, and include the latter in #255

Daanvdplas (commented):
This is definitely something thats need to be added, great job noting it. My suggestion is to finish this PR without the example, and include the latter in #255

evilrobot-01:
It should be included in some contract/test at the very least.

Deferring actual usage of things to the very end might mean only discovering a usability issue at the end. A quick prototype of a wrapper contract taking token properties via constructor to create a token, then implementing the trait to expose access to the token, isn't hard.

The existing constructor contract within the integration tests can probably be adapted pretty easily.

chungquantin:
Duplicate: https://github.com/r0gue-io/pop-node/pull/297

PR #273, opened 2024-09-10
test: primitives

Daanvdplas:


primitives/src/lib.rs
```diff
@@ -150,4 +153,49 @@ pub mod v0 {
...
 		}
 	}
+
+	#[cfg(test)]
+	mod tests {
+		use super::*;
+
+		#[test]
+		fn test_error_u32_conversion_with_all_variants() {
+			let error_variants = vec![
```
chungquantin, line 163: Minor suggestion, variable is redundant and code can be simplify + kept consistent with other tests. 
```suggestion
	        #[test]
		fn test_error_u32_conversion_with_all_variants() {
			// Test conversion for all Error variants
			vec![
				Error::Other,
				Error::CannotLookup,
				Error::BadOrigin,
				Error::Module { index: 1, error: [2, 3] },
				Error::ConsumerRemaining,
				Error::NoProviders,
				Error::TooManyConsumers,
				Error::Token(TokenError::FundsUnavailable),
				Error::Arithmetic(ArithmeticError::Underflow),
				Error::Transactional(TransactionalError::LimitReached),
				Error::Exhausted,
				Error::Corruption,
				Error::Unavailable,
				Error::RootNotAllowed,
				Error::DecodingFailed,
				Error::Unknown { dispatch_error_index: 1, error_index: 2, error: 3 },
			]
			.into_iter()
			.for_each(|error| {
				let status_code = u32::from(error.clone());
				let decoded_error = Error::from(status_code);
				assert_eq!(decoded_error, error);
			});
		}
```

chungquantin (approved):

primitives/src/lib.rs
```diff
@@ -16,8 +19,8 @@ pub mod v0 {
 	mod error {
 		use super::*;
 
-		/// Reason why a Pop API call failed.
-		#[derive(Encode, Decode, Debug, Eq, PartialEq)]
+		/// Reason why a call failed.
+		#[derive(Encode, Decode, Debug, Eq, PartialEq, Clone)]
```
evilrobot-01, line 23: Do contracts need to be able to clone? Would think not. This is probably going to increase contract size somewhat, just so that you can clone it for tests. Please move all the clone derives under the std feature as on line 24.

It might be worth considering which of the other derives should be moved as well.

evilrobot-01 (changes requested):
Potential size issues introduced and I dont think the conversion from an error to a status code is adequately tested, with indirect coverage only.

primitives/src/lib.rs
```diff
@@ -150,4 +153,47 @@ pub mod v0 {
...
 	}
+
+	#[cfg(test)]
+	mod tests {
+		use super::*;
+
+		#[test]
+		fn test_error_u32_conversion_with_all_variants() {
+			// Test conversion for all Error variants
+			vec![
```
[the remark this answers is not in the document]
Daanvdplas: This is a great suggestion, thank you

primitives/src/lib.rs
Daanvdplas: My thought process was that they would "probably" need them, also for testing but under the `std` features they will be able to. Better to have as little features enabled and based on user feedback see if something needs changing.

evilrobot-01 (approved):
Added conversion function could have been a one liner and an explicit test would have been nice but fine as is.

PR #270, opened 2024-09-10
feat(devnet): add ismp

peterwht:
This PR adds ISMP to the devnet runtime. The purpose is to get ISMP added to devnet to start deeper work to interface ISMP with contracts, and create more comprehensive testing.

It includes:

**Devnet runtime:**
- ISMP pallet
- removed: Demo pallet

**Node**
- ISMP Inherent
- ISMP RPC

**Test**
- removed: Script to test GET request + response

Note:
In order to use ISMP, it must be built with `--features ismp` enabled. When this feature is enabled, both testnet and mainnet runtimes will be disabled and error with "ISMP is not enabled". If the feature is not enabled, devnet, testnet, and mainnet will be built. However, ISMP will not work for devnet.

Because the node requires that the runtimes implement the runtime API, we must feature-gate with the ISMP feature. Unfortunately, because trait aliases are experimental (nightly) feature-gating is rather verbose allowing us to remain on stable.

Updated CI PR here: https://github.com/r0gue-io/pop-node/pull/276

PR #276, opened 2024-09-11
feat(CI): add ismp feature to CI, and choosing runtime for release

peterwht:
Related: #270 
Note: 270 is the base branch. Please only review:
- .github/workflows/ci.yml
- .github/workflows/release.yml

No other files changed.

Updates the CI to _also_  build the runtimes with `--features ismp` enabled. It also modifies the release CI to only build one selected runtime or the runtime for a release (see below). If the runtime is devnet, it will build the node with `--features ismp`.

**This means the release CI will now only build ONE runtime.** Considering this is for releases, I do not see a reason we would need to build several runtimes. A testnet and mainnet release should be distinct. There is no reason to have a devnet release. There is also the option for `none`, which skips the srtool build altogether.

Screenshot 2024-09-13 at 4 04 03 PM

## Build Runtime on Release
When a release is published, the CI will build that specific runtime only. This requires that the release name abides by a specific format. This format requires the runtime name as the prefix:
```
<runtime_name_prefix>-<*can be whatever>
// testnet-v0.1.0
// devnet-v0.1.0
// mainnet-v0.1.0
```

The CI's running can be confirmed here:
**release.yml**
- Devnet release: https://github.com/peterwht/pop-node/releases/tag/devnet-v01-CI-Test
- Automatic CI run: https://github.com/peterwht/pop-node/actions/runs/10802403850
- Manual dispatch: https://github.com/peterwht/pop-node/actions/runs/10803352479
- node only: https://github.com/peterwht/pop-node/actions/runs/10856692574

**ci.yml**
- https://github.com/peterwht/pop-node/actions/runs/10802399442

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,16 @@ jobs:
         run: |
           cargo check --release --locked --features=runtime-benchmarks,try-runtime
 
+  check-ismp:
+    needs: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: "./.github/actions/init"
+      - name: Check Build with ISMP
+        run: |
+          cargo check --release --locked --features=ismp,runtime-benchmarks,try-runtime
+
   clippy:
     needs: lint
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


peterwht:
[sc-1326]

.github/workflows/release.yml
```diff
@@ -16,45 +16,73 @@ on:
         description: Build node
         default: true
       runtime:
-        type: boolean
-        description: Build runtimes deterministically
-        default: true
+        description: 'Runtime to build (devnet, testnet, mainnet)'
+        required: true
```
evilrobot-01, line 20: If this is required and a choice, does it mean that a runtime will always build when just wanting to build the node, which in turn will build all the runtimes?

peterwht: Good point! See the latest commit and updated PR description.

evilrobot-01: Rather than 'none', would it not be simpler to remove `required: true` and use `github.event.inputs.runtime != ''`?

peterwht: For the choice input it still requires that something is selected even if required is removed. So, I set the default to be "" (empty string), and added a new choice "".

evilrobot-01 (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -62,6 +72,24 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: --release --locked --features=runtime-benchmarks
```
Daanvdplas, line 74: Newline

Daanvdplas (approved):

.github/workflows/ci.yml
```diff
@@ -62,6 +72,24 @@ jobs:
...
+  clippy-ismp:
```
evilrobot-01, line 75:
```suggestion

  clippy-ismp:
```

Issue #279, opened 2024-09-11
test: gas contracts


Daanvdplas:
https://github.com/r0gue-io/pop-node/pull/132#discussion_r1750842354

PR #277, opened 2024-09-11
chore(devnet): add sandbox & test contracts with drink

chungquantin:
## Description
- Link to issue: #251 
- Drink crate: https://github.com/r0gue-io/pop-drink

Implementation of the [`pop_drink::Sandbox`](https://github.com/r0gue-io/pop-drink) struct for the Pop Network runtimes (located in `pop-node/runtime`) required for the quasi testing with `drink`.

This PR creates a `pop-sandbox` crate that implements the `pop-drink::Sandbox` runtime environment for the `Devnet` => `DevnetSandbox`.

In the context of quasi-testing with pop-drink, a sandbox refers to an isolated runtime environment that simulates the behavior of a full node, without requiring an actual node. It can emulate key processes (where runtime pallets are involved) such as block initialization, execution, and block finalization. 

## Tasks
- [ ] (`release`) Publish `pop-drink` crate to registry. 
- [ ] (`release`) Publish `pop-sandbox` crate to registry. 
- [x] `DevnetSandbox` implementation in `pop-sandbox`
- [x] README page for `pop-sandbox` crate.
- [x] Example production-ready contracts with drink test using Pop API: https://github.com/r0gue-io/pop-node/pull/300
- [x] Need utility functions for `drink` test: https://github.com/r0gue-io/pop-node/pull/300

## References
- Previous commit history: https://github.com/r0gue-io/pop-node/pull/255

PR #216, opened 2024-08-19
test: assert events emitted from contract 

chungquantin:
Link to issue: https://github.com/r0gue-io/pop-node/issues/181
Refactor fungible `integration-tests` contract to emit events by default and catch those events in the integration test cases.

PR #280, opened 2024-09-11
refactor: api integration tests

Daanvdplas:
Contains all changes coming from comments for `pop-api` in #132 . Please not that in the comments I used #279 (incorrect) and #280. They both count for this PR.

Changes include:
- Example refactor
- refactor (pop api): ChainExtensionMethodApi type + small doc changes.
- refactor (integration tests): 
1. remove clone
2. terminology to token & token_id
3. refactor env_logger
4. refactor helper functions from pallet assets
5. probably some more small ones.

### DIFF
--- a/pop-api/examples/fungibles/lib.rs
+++ b/pop-api/examples/fungibles/lib.rs
@@ -1,14 +1,9 @@
 #![cfg_attr(not(feature = "std"), no_std, no_main)]
 
-/// Local Fungibles:
-/// 1. PSP-22 Interface
-/// 2. PSP-22 Metadata Interface
-/// 3. Asset Management
-///
 use ink::prelude::vec::Vec;
 use pop_api::{
 	assets::fungibles::{self as api},
-	primitives::AssetId,
+	primitives::TokenId,
 	StatusCode,
 };
 
@@ -29,30 +24,20 @@ mod fungibles {
 			Default::default()
 		}
 
-		/// 1. PSP-22 Interface:
-		/// - total_supply
-		/// - balance_of
-		/// - allowance
-		/// - transfer
-		/// - transfer_from
-		/// - approve
-		/// - increase_allowance
-		/// - decrease_allowance
-
 		#[ink(message)]
-		pub fn total_supply(&self, id: AssetId) -> Result<Balance> {
+		pub fn total_supply(&self, id: TokenId) -> Result<Balance> {
 			api::total_supply(id)
 		}
 
 		#[ink(message)]
-		pub fn balance_of(&self, id: AssetId, owner: AccountId) -> Result<Balance> {
+		pub fn balance_of(&self, id: TokenId, owner: AccountId) -> Result<Balance> {
 			api::balance_of(id, owner)
 		}
 
 		#[ink(message)]
 		pub fn allowance(
 			&self,
-			id: AssetId,
+			id: TokenId,
 			owner: AccountId,
 			spender: AccountId,
 		) -> Result<Balance> {
@@ -60,32 +45,31 @@ mod fungibles {
 		}
 
 		#[ink(message)]
-		pub fn transfer(&mut self, id: AssetId, to: AccountId, value: Balance) -> Result<()> {
+		pub fn transfer(&mut self, id: TokenId, to: AccountId, value: Balance) -> Result<()> {
 			api::transfer(id, to, value)
 		}
 
 		#[ink(message)]
 		pub fn transfer_from(
 			&mut self,
-			id: AssetId,
+			id: TokenId,
 			from: AccountId,
 			to: AccountId,
 			value: Balance,
-			// In the PSP-22 standard a `[u8]`, but the size needs to be known at compile time.
 			_data: Vec<u8>,
 		) -> Result<()> {
 			api::transfer_from(id, from, to, value)
 		}
 
 		#[ink(message)]
-		pub fn approve(&mut self, id: AssetId, spender: AccountId, value: Balance) -> Result<()> {
+		pub fn approve(&mut self, id: TokenId, spender: AccountId, value: Balance) -> Result<()> {
 			api::approve(id, spender, value)
 		}
 
 		#[ink(message)]
 		pub fn increase_allowance(
 			&mut self,
-			id: AssetId,
+			id: TokenId,
 			spender: AccountId,
 			value: Balance,
 		) -> Result<()> {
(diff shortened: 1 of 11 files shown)


pop-api/examples/fungibles/lib.rs
```diff
@@ -95,91 +79,34 @@ mod fungibles {
...
 
 	#[cfg(test)]
 	mod tests {
 		use super::*;
 
 		#[ink::test]
-		fn default_works() {
-			PopApiFungiblesExample::new();
-		}
+		fn default_works() {}
```
chungquantin, line 110: Maybe we should just remove the test module. FYI, this is replaced with `drink` test in this PR: https://github.com/r0gue-io/pop-node/pull/277. Hence, does not matter wether we remove it or not.

pop-api/integration-tests/src/fungibles/mod.rs
```diff
@@ -488,89 +445,74 @@ fn clear_metadata_works() {
...
+		let balance_after_burn = Assets::balance(token, &BOB);
 		assert_eq!(balance_after_burn, balance_before_burn - amount);
-		// Asset is not live, i.e. frozen or being destroyed.
-		start_destroy_asset(addr.clone(), asset);
-		assert_eq!(
-			burn(addr.clone(), asset, BOB, amount),
-			Err(Module { index: 52, error: [17, 0] })
-		);
+		// Token is not live, i.e. frozen or being destroyed.
+		pallet_assets_start_destroy(&addr, token);
```
chungquantin, line 515: Instead of repeating the methods with `pallet_assets` as prefix like this, why don't we create a `pallet_assets` module and put those methods inside it?

pop-api/examples/fungibles/lib.rs
Daanvdplas: Will keep it for now, can be removed it with your PR

pop-api/integration-tests/src/fungibles/mod.rs
```diff
@@ -330,122 +304,106 @@ fn token_metadata_works() {
...
+		pallet_assets_freeze(&addr, token);
 		assert_eq!(
-			set_metadata(addr.clone(), TOKEN_ID, vec![0], vec![0], 0u8),
+			set_metadata(&addr, TOKEN_ID, vec![0], vec![0], 0u8),
 			Err(Module { index: 52, error: [16, 0] }),
 		);
-		thaw_asset(addr.clone(), asset);
+		pallet_assets_thaw(&addr, token);
 		// TODO: calling the below with a vector of length `100_000` errors in pallet contracts
 		//  `OutputBufferTooSmall. Added to security analysis issue #131 to revisit.
```
peterwht, line 395: not sure this is an issue, just means pallet-contracts properly exists with values too large.

pop-api/src/lib.rs
```diff
@@ -16,49 +17,16 @@ pub mod primitives;
 /// The first version of the API.
 pub mod v0;
 
-/// A result type used by the API, with the `StatusCode` as the error type.
+type ChainExtensionMethodApi = ChainExtensionMethod<(), (), (), false>;
```
peterwht, line 20: can you please add a comment for each parameter is?

peterwht (approved):
Nice refactor! I'm going to approve, but please address Tin's comments as I agree with them

pop-api/integration-tests/src/fungibles/mod.rs
Daanvdplas: Much cleaner, thank you!

pop-api/examples/fungibles/lib.rs
```diff
@@ -29,63 +24,52 @@ mod fungibles {
...
-		/// - allowance
-		/// - transfer
-		/// - transfer_from
-		/// - approve
-		/// - increase_allowance
-		/// - decrease_allowance
-
 		#[ink(message)]
-		pub fn total_supply(&self, id: AssetId) -> Result<Balance> {
+		pub fn total_supply(&self, id: TokenId) -> Result<Balance> {
```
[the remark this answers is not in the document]
Daanvdplas: Sharp, thank you!

pop-api/integration-tests/src/fungibles/mod.rs
Daanvdplas: Something to keep in mind though because whenever the developer wants to set metadata and uses a value that is too large. In stead of pallet assets that returns an error it is pallet contracts.

pop-api/src/lib.rs
Daanvdplas: Not sure how valuable it is as most of our values are set to `()`. Thus we explain what each generic argument is but we set it to `()`, this might only cause confusion. Only in the actual api function the types are provided, but by calling functions:
```
build_read_state(BALANCE_OF)
		.input::<(TokenId, AccountId)>()
		.output::<Result<Balance>, true>()
		.handle_error_code::<StatusCode>()
		.call(&(token, owner))
```

chungquantin:
Rebased this PR https://github.com/r0gue-io/pop-node/pull/216 to this branch. My suggestion is to merge the api event testing branch #216 first to this branch. I already rebased 216 to `daan/api`.

cc: @Daanvdplas

chungquantin (approved):
Looks good!

pop-api/integration-tests/src/fungibles/mod.rs
evilrobot-01: Would expect that it might fail with out of gas realistically.

pop-api/integration-tests/src/lib.rs
```diff
@@ -3,39 +3,40 @@
 use frame_support::{
 	assert_ok,
 	traits::fungibles::{
-		approvals::Inspect as ApprovalInspect, metadata::Inspect as MetadataInspect, Inspect,
+		approvals::Inspect as ApprovalInspect, metadata::Inspect as MetadataInspect,
```
evilrobot-01, line 6: FYI - if trait just needs to be in scope, and not explicitly used, can just use `as _` as a simpler approach which avoids naming clashes.

pop-api/integration-tests/src/lib.rs
```diff
@@ -46,13 +47,12 @@ fn new_test_ext() -> sp_io::TestExternalities {
...
 
-fn load_wasm_module<T>(path: &str) -> std::io::Result<(Vec<u8>, <T::Hashing as Hash>::Output)>
+fn load_wasm_module<T>(path: &str) -> std::io::Result<Vec<u8>>
 where
 	T: frame_system::Config,
 {
 	let wasm_binary = std::fs::read(path)?;
-	let code_hash = T::Hashing::hash(&wasm_binary);
-	Ok((wasm_binary, code_hash))
+	Ok(wasm_binary)
```
evilrobot-01, line 55: Fine as is, but if you do another pass this could be simplified to just be 'std::fs::read(path)' as the only statement in the function. Not sure the T generic type parameter is required either.

pop-api/integration-tests/src/lib.rs
```diff
@@ -82,8 +82,7 @@ fn bare_call(
 
 // Deploy, instantiate and return contract address.
 fn instantiate(contract: &str, init_value: u128, salt: Vec<u8>) -> AccountId32 {
-	let (wasm_binary, _) =
-		load_wasm_module::<Runtime>(contract).expect("could not read .wasm file");
+	let wasm_binary = load_wasm_module::<Runtime>(contract).expect("could not read .wasm file");
```
evilrobot-01, line 85: If this is the only usage, we dont need a whole function when `std::fs::read(contract).expect("could not read .wasm file")` would be enough.

pop-api/integration-tests/src/fungibles/utils.rs
```diff
@@ -1,325 +1,340 @@
 use super::*;
 
-fn do_bare_call(function: &str, addr: AccountId32, params: Vec<u8>) -> ExecReturnValue {
+type AssetId = TokenId;
```
evilrobot-01, line 3: Move to assets submodule?

evilrobot-01 (approved):
Some very minor improvements suggested, wouldnt be against merging as is however.

pop-api/integration-tests/src/fungibles/utils.rs
```diff
@@ -309,21 +310,15 @@ pub(super) mod assets {
 	}
 
 	pub(crate) fn token_name(asset_id: AssetId) -> Vec<u8> {
-		<pallet_assets::Pallet<Runtime, TrustBackedAssetsInstance> as MetadataInspect<
-			AccountId32,
-		>>::name(asset_id)
+		Assets::name(asset_id)
```
evilrobot-01, line 313: One line functions seem redundant.

Daanvdplas: I prefer to keep it consistent and everything is called via the `assets` module

evilrobot-01: I mean why wrap it in a function 'token_name()' in the utils pallet - its unnecessary complexity.

test() -> utils -> token_name() -> Assets::name()
vs
test() -> Assets::name()

Daanvdplas: You are right, one liner helper functions do not make sense. Sorted in last commit

PR #283, opened 2024-09-12
feat: read weights

evilrobot-01:


peterwht:
https://github.com/r0gue-io/pop-node/pull/284

Issue #127, opened 2024-07-25
refactor: reading state charging weight


Daanvdplas:
Read state weight implementation with benchmarks and test coverage (https://github.com/r0gue-io/pop-node/pull/132#discussion_r1749049525).

Including proof size ([SE](https://substrate.stackexchange.com/questions/11842/calculate-proof-size-of-a-function-that-reads-state)).

>The Proof size weight of a storage read is the maximal size of the storage item being read. We have the MaxEncodedLen trait to track this. So you need contextual information about the type that is being read and then use MaxEncodedLen::max_encoded_len() on it.
Depending on what the chain extension does, it may or may not be easily possible to know what key is being accessed. Smart contracts are normally uses an optimistic approach and revert on error, whereas FRAME requires to know the worst case effort in advance. These are just fundamentally different computational models.
If you can: write a benchmark. The only way to currently write FRAME benchmarks it to use a pallet; for example the minimal pallet or some mocked types just to make the benchmark compile.
If you want to know how to charge it exactly, that would be env.charge_weight(Weight::from_parts(0, proof_size)).

evilrobot-01:
https://github.com/r0gue-io/pop-node/pull/122#issuecomment-2252000985

Daanvdplas:
Your suggested approach won't allow for super clean code, it would require to match on the enum to obtain the weight, in stead of just calling the `read_state` function in the api pallet. An alternative solution is to have the `read_state` function return raw bytes plus the amount of reads / weight.

evilrobot-01:
I disagree. https://github.com/r0gue-io/pop-cli/blob/b4874220ca58ebb38453c7ab83670f7e0711c674/crates/pop-parachains/src/templates.rs#L168 shows an example of attaching a custom property to an enum using strum, with https://github.com/r0gue-io/pop-cli/blob/b4874220ca58ebb38453c7ab83670f7e0711c674/crates/pop-parachains/src/templates.rs#L194 showing how it can then be obtained. This could be reimplemented as a custom enum property which is statically defined on the enum, and can be used to charge the defined amount of reads/weight BEFORE execution of the read. Your suggestion of attaching it to the output of `read_state` wont work as charging after a read opens an attack vector.

Yes, a match may be required, but so what. The important thing is accurate weight charging before a read. This can be defined within an additional `read_weight` function, which could even potentially be defined as a trait with a default implementation. Further improvements could refactor any match out completely.

[...]

PR #284, opened 2024-09-12
feat(pallet-api/fungibles): benchmark read functions

peterwht:
Solves https://github.com/r0gue-io/pop-node/issues/127


pallets/api/src/fungibles/tests.rs
```diff
@@ -342,3 +342,98 @@ fn set_metadata_token(
...
+	#[test]
+	fn ensure_asset_variants_match() {
+		let ReadWeightInfo { total_supply, token_exists, .. } = ReadWeightInfo::new();
+
+		assert_eq!(total_supply, token_exists);
+	}
+
+	// Proof size is based on `MaxEncodedLen`, not hardware.
+	#[test]
+	fn ensure_expected_proof_size_does_not_change() {
```
peterwht, line 420: If you want to double check  this, please try on your machine!
```
./target/release/pop-node \
benchmark \
pallet \
--chain=dev \
--wasm-execution=compiled \
--pallet=fungibles \
--steps=50 \
--repeat=20 \
--json \
--template \
./scripts/pallet-weights-template.hbs \
--output=./pallets/api/src/fungibles/weights.rs \
--extrinsic=
```

pallets/api/src/fungibles/benchmarking.rs
```diff
@@ -103,5 +106,70 @@ mod benchmarks {
 		Ok(())
 	}
 
+	#[benchmark]
```
Daanvdplas, line 109: Can we have those in the same order as in the `Read` and `ReadResult` enum pls

pallets/api/src/fungibles/tests.rs
Daanvdplas: Would be nice if we also assert to the maxencodedlen of the structs from pallet assets maybe? How you did it previously

pallets/api/src/fungibles/mod.rs
```diff
@@ -490,9 +490,17 @@ pub mod pallet {
 		///
 		/// # Parameters
 		/// - `request` - The read request.
-		fn weight(_request: &Self::Read) -> Weight {
-			// TODO: match on request and return benchmarked weight
-			T::DbWeight::get().reads(1_u64)
+		fn weight(request: &Self::Read) -> Weight {
+			use Read::*;
+			match request {
```
Daanvdplas, line 495: Also here the order of the variants as in the `Read` enum, consistency.

Daanvdplas (commented):
Really nice! Left a few nitpicks but then good to go!

pallets/api/src/fungibles/tests.rs
peterwht: please see latest commit

evilrobot-01: Yeah. Tbh when I saw your comment last night I thought the same, so had the same misunderstanding.

My view is that if we rely on the accuracy of benchmarks/weights for dispatchables, recreating it for reads seems unnecessary. There is a framework there for benchmarking and getting in-depth results out of it, so its reasonable to rely on it as that is what underpins the ecosystem. Fine for some additional sanity checks tho.

pallets/api/src/fungibles/benchmarking.rs
```diff
@@ -103,5 +106,69 @@ mod benchmarks {
...
+	fn total_supply() {
+		#[block]
+		{
+			Pallet::<T>::read(Read::TotalSupply(TokenIdOf::<T>::zero()));
+		}
+	}
+	#[benchmark]
+	// Storage: `AssetAccount`
+	fn balance_of() {
+		let owner: AccountIdOf<T> = account("Alice", 0, SEED);
```
evilrobot-01, line 120: Minor, but `owner: account("Alice", 0, SEED)` inline below might be simpler. It conveys the same in a more compact style. Same with the others benchmarks.

pallets/api/src/fungibles/tests.rs
evilrobot-01: Whilst I admire the level of detail, clarity and accuracy of these tests, my only reservation would be whether this would become an expectation moving forward. My concern would be the additional time required to implement for all the other APIs to remain consistent versus the value it brings.

evilrobot-01 (approved):
Minor suggestion.

Also think the level of detail of tests needs considering, as mentioned inline.

pallets/api/src/fungibles/tests.rs
Daanvdplas: Same, therefore I think it is best to remove. Thank for the clear detailed docs and tests though Peter, was able to immediately grasp the detailed workings of it!

Daanvdplas: Thanks for the clear detailed docs and tests though Peter, was able to immediately grasp the detailed workings of it!

peterwht: It was quite tedious to do! Although thankful for the learnings :) The last test that verifies the proof size remains the same will effectively provide the same testing outcome (make sure the max_encoded_len does not change) please see the latest commit, reverting the changes.

evilrobot-01 (approved):

pallets/api/src/fungibles/benchmarking.rs
peterwht: see latest commit

PR #307, opened 2024-09-26
feat: return `UnApproved` in `decrease_allowance` instead of `saturating_sub`

chungquantin:
`Fungibles::decrease_allowance()` saturating_sub the value, it should `checked_sub` and throw an error instead. Hence, we don't have to handle `InsufficientAllowance` on the contract side.

Benefit of this approach is to make the implementation of PSP22 example contract cleaner.

### DIFF
--- a/pallets/api/src/fungibles/mod.rs
+++ b/pallets/api/src/fungibles/mod.rs
@@ -339,16 +339,16 @@ pub mod pallet {
 			let token_param: TokenIdParameterOf<T> = token.clone().into();
 
 			// Cancel the approval and approve `new_allowance` if difference is more than zero.
+			let new_allowance = match current_allowance.checked_sub(&value) {
+				Some(allowance) => allowance,
+				None => return Err(AssetsErrorOf::<T>::Unapproved.into()),
+			};
 			AssetsOf::<T>::cancel_approval(
 				origin.clone(),
 				token_param.clone(),
 				spender_source.clone(),
 			)
 			.map_err(|e| e.with_weight(WeightOf::<T>::approve(0, 1)))?;
-			let new_allowance = match current_allowance.checked_sub(&value) {
-				Some(allowance) => allowance,
-				None => return Err(AssetsErrorOf::<T>::Unapproved.into()),
-			};
 			let weight = if new_allowance.is_zero() {
 				WeightOf::<T>::approve(0, 1)
 			} else {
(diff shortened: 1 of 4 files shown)


pallets/api/src/fungibles/mod.rs
```diff
@@ -344,7 +345,10 @@ pub mod pallet {
 				spender_source.clone(),
 			)
 			.map_err(|e| e.with_weight(WeightOf::<T>::approve(0, 1)))?;
-			let new_allowance = current_allowance.saturating_sub(value);
+			let new_allowance = match current_allowance.checked_sub(&value) {
```
Daanvdplas, line 348: This should error before the `cancel_approval` above so that we are not calling it for no reason.

pallets/api/src/fungibles/tests.rs
```diff
@@ -286,6 +286,12 @@ fn decrease_allowance_works() {
 			Ok(Some(WeightInfo::approve(0, 0)).into())
 		);
 		assert_eq!(Assets::allowance(token, &owner, &spender), value);
+		// `UnApproved` error returned if current allowance is decreased more than the owner
+		// balance.
```
Daanvdplas, line 290: More than the allowance I expect?

pop-api/integration-tests/src/fungibles/mod.rs
```diff
@@ -267,9 +267,14 @@ fn decrease_allowance_works() {
 			Err(Module { index: 52, error: [16, 0] }),
 		);
 		assets::thaw(&addr, token);
+		// Error returned if current allowance is decreased more than the owner balance.
```
Daanvdplas, line 270: more than the current allowance I assume again?

Daanvdplas (commented):
Nice! A few points for attention

pop-api/integration-tests/src/fungibles/mod.rs
```diff
@@ -267,7 +267,7 @@ fn decrease_allowance_works() {
 			Err(Module { index: 52, error: [16, 0] }),
 		);
 		assets::thaw(&addr, token);
-		// Error returned if current allowance is decreased more than the owner balance.
+		// `UnApproved` error returned if current allowance is decreased more than the allowance.
```
Daanvdplas, line 270: "Unapproved" error is returned if the current allowance is less than value

pallets/api/src/fungibles/tests.rs
```diff
@@ -286,8 +286,7 @@ fn decrease_allowance_works() {
 			Ok(Some(WeightInfo::approve(0, 0)).into())
 		);
 		assert_eq!(Assets::allowance(token, &owner, &spender), value);
-		// `UnApproved` error returned if current allowance is decreased more than the owner
-		// balance.
+		// `UnApproved` error returned if current allowance is decreased more than the allowance.
```
Daanvdplas, line 289: "Unapproved" error is returned if the current allowance is less than `value`

Daanvdplas (approved):

PR #312, opened 2024-09-27
feat: token name & symbol return `Option`

Daanvdplas:
The psp22 standard returns an `Option<String>`. In stead of having the contract needing to add additional logic, we handle it in the pallet.

### DIFF
--- a/pallets/api/src/fungibles/tests.rs
+++ b/pallets/api/src/fungibles/tests.rs
@@ -47,7 +47,7 @@ mod encoding_read_result {
 
 	#[test]
 	fn token_name() {
-		let mut name = Some(vec![42, 42, 42, 42, 42]);
+		let mut name = Some("some name".as_bytes().to_vec());
 		assert_eq!(ReadResult::TokenName::<Test>(name.clone()).encode(), name.encode());
 		name = None;
 		assert_eq!(ReadResult::TokenName::<Test>(name.clone()).encode(), name.encode());
@@ -55,7 +55,7 @@ mod encoding_read_result {
 
 	#[test]
 	fn token_symbol() {
-		let mut symbol = Some(vec![42, 42, 42, 42, 42]);
+		let mut symbol = Some("some symbol".as_bytes().to_vec());
 		assert_eq!(ReadResult::TokenSymbol::<Test>(symbol.clone()).encode(), symbol.encode());
 		symbol = None;
 		assert_eq!(ReadResult::TokenSymbol::<Test>(symbol.clone()).encode(), symbol.encode());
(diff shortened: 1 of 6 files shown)


pallets/api/src/fungibles/tests.rs
```diff
@@ -47,13 +47,17 @@ mod encoding_read_result {
 
 	#[test]
 	fn token_name() {
-		let name = vec![42, 42, 42, 42, 42];
+		let mut name = Some(vec![42, 42, 42, 42, 42]);
```
chungquantin, line 50: Do we truly need a variable? It is not reused anywhere except the next line.

pallets/api/src/fungibles/tests.rs
```diff
@@ -47,13 +47,17 @@ mod encoding_read_result {
...
+		assert_eq!(ReadResult::TokenName::<Test>(name.clone()).encode(), name.encode());
+		name = None;
 		assert_eq!(ReadResult::TokenName::<Test>(name.clone()).encode(), name.encode());
 	}
 
 	#[test]
 	fn token_symbol() {
-		let symbol = vec![42, 42, 42, 42, 42];
+		let mut symbol = Some(vec![42, 42, 42, 42, 42]);
```
chungquantin, line 58: Do we truly need a variable? It is not reused anywhere except the next line.

pop-api/integration-tests/src/fungibles/mod.rs
```diff
@@ -295,18 +295,16 @@ fn token_metadata_works() {
 		let decimals: u8 = 69;
 
 		// Token does not exist.
-		assert_eq!(token_name(&addr, TOKEN_ID), Ok(Assets::name(TOKEN_ID)));
-		assert_eq!(token_name(&addr, TOKEN_ID), Ok(Vec::<u8>::new()));
-		assert_eq!(token_symbol(&addr, TOKEN_ID), Ok(Assets::symbol(TOKEN_ID)));
-		assert_eq!(token_symbol(&addr, TOKEN_ID), Ok(Vec::<u8>::new()));
+		assert_eq!(token_name(&addr, TOKEN_ID), Ok(None));
+		assert_eq!(token_symbol(&addr, TOKEN_ID), Ok(None));
 		assert_eq!(token_decimals(&addr, TOKEN_ID), Ok(Assets::decimals(TOKEN_ID)));
```
chungquantin, line 300: Should we also remove this to keep the consistency?

chungquantin (approved):
Some nitpitcks but not very important.

pallets/api/src/fungibles/tests.rs
Daanvdplas: Yes I think it is personally cleaner.

pop-api/integration-tests/src/fungibles/mod.rs
Daanvdplas: Removed

pallets/api/src/fungibles/tests.rs
evilrobot-01, line 50: `let name = "some name".as_bytes()` wouldve been nicer, more descriptive. Looping through an array of the cases might also neaten this up slightly

```
for name in [Some("Pop".as_bytes()), None]
{ assert_eq!(..) }
```

Out of scope, but using actual strings in other tests might be nicer, but not a requirement.

pallets/api/src/fungibles/tests.rs
evilrobot-01, line 58: As above.

pallets/api/src/fungibles/tests.rs
```diff
@@ -504,11 +508,14 @@ fn token_metadata_works() {
 			ReadResult::TokenDecimals(Default::default())
```
evilrobot-01, line 508: Would be nicer to use None in the above examples rather than default to be more explicit: `assert_eq!(Fungibles::read(TokenName(TOKEN)), ReadResult::TokenName(None));`, as in the integration tests.

Comment only, not a requirement.

evilrobot-01 (approved):
Looks good, made some minor suggestions for improving the tests which are quick and easy to implement if you so choose.

al3mart (approved):

Issue #313, opened 2024-09-28
ci(bug): `api-integration-tests` duplicate lang item in crate `sp_io` 


chungquantin:
Got this error on my local when trying to run the integration-tests of pop-api in `main`. See CI error in other branches that also failed at this job.

chungquantin:
Found a solution for this
```
cargo update staging-xcm --precise 14.1.0
```
But the `Cargo.lock` is included in the `.gitignore`, not sure how to change this yet,

evilrobot-01:
Can you not add/specify the dependency as "=14.1.0"?

PR #317, opened 2024-10-01
fix(ci): api-integration-tests

chungquantin:
https://github.com/r0gue-io/pop-node/issues/313


Daanvdplas (approved):

Daanvdplas (commented):
Can we add a TODO above

evilrobot-01 (approved):
Looks fine, would be better with a comment so that someone might try to remove it in the future rather than leaving it as is forever.

evilrobot-01 (approved):

PR #311, opened 2024-09-27
feat(pallet): follow psp22 standard for `decrease_allowance` and `burn`

chungquantin:
Discussion: https://github.com/r0gue-io/pop-node/pull/300#discussion_r1777416957
- For `decrease_allowance`: `Fungibles::decrease_allowance()` saturating_sub the value, it should `checked_sub` and throw an error instead. Hence, we don't have to handle `InsufficientAllowance` on the contract side.
- For `burn`: `InsufficientBalance` case is not returned in the `burn` pallet api if the `value` exceeds the minted value. In `decrease_balance` method of `pallet-assets`, it is also `saturating_sub`.

PR #300, opened 2024-09-19
feat: psp22 trait contract example with drink tests

chungquantin:
## Description
Example contract for the fungible use case that applies the PSP22 traits from https://github.com/r0gue-io/pop-node/pull/297. The PR includes: 
- Quasi testing with `pop-drink` and `pop-sandbox`: https://github.com/r0gue-io/pop-drink
- Remove all the deprecated example contracts

Issue #321, opened 2024-10-02
ci(devnet): auto re-running benchmark for pallets on push 


chungquantin:
Perhaps only rerun benchmark for pallets under `pallets` folder.

mahmudsudo:
can i take on this ?

chungquantin:
Yes @mahmudsudo

PR #322, opened 2024-10-02
feat: psp22 standard for `decrease_allowance` and `burn` pallet methods

chungquantin:
Discussion: https://github.com/r0gue-io/pop-node/pull/300#discussion_r1777416957
- For `decrease_allowance`: `Fungibles::decrease_allowance()` saturating_sub the value, it should `checked_sub` and throw an error instead. Hence, we don't have to handle `InsufficientAllowance` on the contract side.
- For `burn`: `InsufficientBalance` case is not returned in the `burn` pallet api if the `value` exceeds the minted value. In `decrease_balance` method of `pallet-assets`, it is also `saturating_sub`.

NOTE: Original PR https://github.com/r0gue-io/pop-node/pull/311 includes file permission changes which is hard to revert and makes the commit history unclean. Hence, I created this PR to substitute the original on and move all comments / resolve comments left by reviewers.


pallets/api/src/fungibles/weights.rs
```diff
@@ -2,7 +2,7 @@
 //! Autogenerated weights for `fungibles`
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 40.0.0
-//! DATE: 2024-09-13, STEPS: `50`, REPEAT: `20`, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! DATE: 2024-10-01, STEPS: `50`, REPEAT: `20`, LOW RANGE: `[]`, HIGH RANGE: `[]`
```
chungquantin, line 5: > @evilrobot-01: Well done for re-running! It might be nice to have devnet benchmarking run automatically on PRs in the future, so that changes included in the comments above the weight function signatures in terms of storage reads/writes can be checked.

Created an issue for this: https://github.com/r0gue-io/pop-node/issues/321

pallets/api/src/fungibles/tests.rs
```diff
@@ -416,10 +417,22 @@ fn burn_works() {
 		let from = BOB;
 		let total_supply = value * 2;
 
-		// Check error works for `Assets::burn()`.
-		assert_noop!(Fungibles::burn(signed(owner), token, from, value), AssetsError::Unknown);
+		// "BalanceLow" error is returned if token is not created.
```
chungquantin, line 420: > @evilrobot-01: Whilst the assets error assertions were to ensure that errors are properly handled, which is the case on line 424, this line could have been left in.

Resolved but need to update the returned error and the comment.

pallets/api/src/fungibles/tests.rs
```diff
@@ -272,20 +272,28 @@ fn decrease_allowance_works() {
 				BadOrigin.with_weight(WeightInfo::approve(0, 0))
 			);
 		}
+		assets::create_mint_and_approve(owner, token, owner, value, spender, value);
```
chungquantin, line 275: > @evilrobot-01: The changes to this test demonstrate why a test per behaviour might be preferred. Such an introduction of a feat should of just had a new test added for a new behaviour, which would have been much cleaner to both write and review.

Daanvdplas (approved):

PR #324, opened 2024-10-03
refactor: `PSP22Error` conversion & assertion

chungquantin:
Refactoring the pop-api error assertions to make it readable instead of comparing with raw bytes. This PR relies on the PR in pop-drink: https://github.com/r0gue-io/pop-drink/pull/14

## About the implementation

Screenshot 2024-10-08 at 00 43 37

Currently, in the example PR, the pop-api error assertion involves converting `pop_api::primitives::Module` to a `u32`. Our goal is to simplify the assertion to make it more readable, aiming for the following conversion chain:

```rs
runtime::<Pallet>::Error -> pop_api::primitives::<version>::Error -> u32
```

### How to assert the `PSP22Error <> pop_drink::Error`?
```rs
use drink::devnet::error::{v0::{assert_runtime_err, RuntimeError},  Assets, AssetsError::AssetNotLive};

assert_runtime_err!(
	transfer(&mut session, BOB, AMOUNT / 2),
	RuntimeError::Module(Assets(AssetNotLive)),
);
```
### If the PalletError `assertion` failed
```rs
---- tests::burn_fails_with_no_permission stdout ----
thread 'tests::burn_fails_with_no_permission' panicked at tests.rs:910:13:
assertion `left == right` failed
  left: Module(RuntimeError::Assets(NoPermission))
 right: Module(RuntimeError::Assets(IncorrectStatus))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
### If the DispatchError `assertion` failed (other than PalletError)
```rs
---- tests::mint_fails_with_arithmetic_overflow stdout ----
thread 'tests::mint_fails_with_arithmetic_overflow' panicked at tests.rs:644:5:
assertion `left == right` failed
  left: Api(Arithmetic(Overflow))
 right: Api(Arithmetic(Underflow))
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #327, opened 2024-10-04
refactor: primitive error conversion

chungquantin:
Moving the conversion code `DispatchError -> V0Error` for `pop_primitives::Error` in `devnet/versioning.rs` to `pop_primitives` to handle `DispatchError -> pop_primitives::Error`.

We need this refactor for for the pop-api refactor in example contract to work. For more context, please take a look at: 
- refactor pop-api error in example contract: https://github.com/r0gue-io/pop-node/pull/324#:~:text=feat%3A%20show%20reviewers,in%20contract%20size.
- pop-drink error type conversion: https://github.com/r0gue-io/pop-drink/pull/14

### DIFF
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -8,10 +8,14 @@ version = "0.0.0"
 [dependencies]
 codec.workspace = true
 scale-info.workspace = true
+sp-runtime.workspace = true
 
 [dev-dependencies]
 enum-iterator = "2.1.0"
 
 [features]
 default = [ "std" ]
-std = [ "codec/std", "scale-info/std" ]
+
+std = [ "codec/std", "scale-info/std", "sp-runtime/std" ]
+
+runtime = []
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -15,5 +15,5 @@ enum-iterator = "2.1.0"
 
 [features]
 default = [ "std" ]
-runtime = []
+runtime = [  ]
 std = [ "codec/std", "scale-info/std", "sp-runtime/std" ]
--- a/primitives/Cargo.toml
+++ b/primitives/Cargo.toml
@@ -15,7 +15,5 @@ enum-iterator = "2.1.0"
 
 [features]
 default = [ "std" ]
-
-std = [ "codec/std", "scale-info/std", "sp-runtime/std" ]
-
 runtime = []
+std = [ "codec/std", "scale-info/std", "sp-runtime/std" ]
--- a/primitives/src/lib.rs
+++ b/primitives/src/lib.rs
@@ -7,6 +7,8 @@ use codec::{Decode, Encode};
 use enum_iterator::Sequence;
 #[cfg(feature = "std")]
 use scale_info::TypeInfo;
+#[cfg(feature = "runtime")]
+use sp_runtime::{DispatchError, ModuleError};
 pub use v0::*;
 
 /// The identifier of a token.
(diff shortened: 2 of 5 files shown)


primitives/Cargo.toml
```diff
@@ -8,10 +8,12 @@ version = "0.0.0"
...
 scale-info.workspace = true
+sp-runtime.workspace = true
 
 [dev-dependencies]
 enum-iterator = "2.1.0"
 
 [features]
 default = [ "std" ]
-std = [ "codec/std", "scale-info/std" ]
+runtime = [  ]
```
peterwht, line 18: would recommend putting some placeholder in here to make zepter and taplo happy.

primitives/src/lib.rs
```diff
@@ -107,6 +109,58 @@ pub mod v0 {
 			}
 		}
 
+		#[cfg(feature = "runtime")]
+		impl From<DispatchError> for Error {
+			fn from(error: DispatchError) -> Self {
+				use sp_runtime::{
+					ArithmeticError::*, DispatchError::*, TokenError::*, TransactionalError::*,
+				};
+				// Mappings exist here to avoid taking a dependency of sp_runtime on pop-primitives
```
Daanvdplas, line 118: Remove

primitives/src/lib.rs
```diff
@@ -7,6 +7,8 @@ use codec::{Decode, Encode};
 use enum_iterator::Sequence;
 #[cfg(feature = "std")]
 use scale_info::TypeInfo;
+#[cfg(feature = "runtime")]
```
Daanvdplas, line 10: Drink importing this crate with `runtime` feature feels odd.

```suggestion
#[cfg(all(feature = "runtime", feature = "std"))]
```

primitives/src/lib.rs
```diff
@@ -195,5 +249,59 @@ pub mod v0 {
 				assert_eq!(error, DecodingFailed,);
 			});
 		}
+
+		// Compare all the different `DispatchError` variants with the expected `Error`.
+		#[test]
+		#[cfg(feature = "runtime")]
```
Daanvdplas, line 255: This should not be required if the "std" feature is added.

runtime/devnet/src/config/api/versioning.rs
```diff
@@ -103,17 +103,9 @@ struct V0Error(pop_primitives::v0::Error);
 impl From<DispatchError> for V0Error {
 	fn from(error: DispatchError) -> Self {
```
Daanvdplas, line 104: Preferably this should just be:
```suggestion
	fn from(error: DispatchError) -> Self {
	    Self(error.into())
	}
```

Makes me wonder whether to remove that whole `DecodingFailed` variant as the developer should never receive this error anyway @evilrobot-01 .

runtime/devnet/src/config/api/versioning.rs
```diff
@@ -103,17 +103,9 @@ struct V0Error(pop_primitives::v0::Error);
...
 		use pop_primitives::v0::*;
-		use sp_runtime::{ArithmeticError::*, TokenError::*, TransactionalError::*};
 		use DispatchError::*;
 		// Mappings exist here to avoid taking a dependency of sp_runtime on pop-primitives
```
Daanvdplas, line 107: Remove

Daanvdplas (commented):
Still some rough edges tmo.

primitives/Cargo.toml
Daanvdplas: How?

primitives/src/lib.rs
chungquantin: Why do we need `std`?  These imports are only required for the conversion, and the conversion is only enabled for `runtime` feature.

primitives/src/lib.rs
chungquantin: Without `runtime` feature enabled, the conversion `Error::from` does not work because `From<DispatchError>` is only enabled when there is `runtime`.

evilrobot-01:
Forgive me if I am missing something, but if drink is implying/requiring a dependency on a specific runtime, then why can't we just reuse the error conversion defined within the runtime rather than moving into primitives?

PR #346, opened 2024-10-19
refactor(devnet): remove pallet-contracts remnants

evilrobot-01:
Just removes a few pallet-contracts remnants from the devnet runtime.

Note: the second commit restores the name of the pallet in the runtime, which restores the Contracts UI within PJS. Whether it works is another story...


evilrobot-01:
No longer applicable.

PR #347, opened 2024-10-20
chore: timestamp settings for tests

Daanvdplas:
Required change for drink test environment. This PR has to be merged to have the reviewers test #300


chungquantin (approved):
No concern from my end!

AlexD10S (approved):

evilrobot-01:
I think we need to be mindful of benchmarks with something like this. Unlikely we will benchmark on devnet runtime tho.

PR #354, opened 2024-10-24
feat(revive): add `to_account_id` host function for converting from a `h160` to `h256` account

evilrobot-01:
Required for a compatible implementation of self.env().caller() for ink!


peterwht (approved):
Niiiiiiice!

PR #357, opened 2024-10-25
feat: builder incentives api

evilrobot-01:
Had to refactor the instantiate function to take data as a parameter so I could call a constructor with params.


AlexD10S (approved):
Great work!

PR #351, opened 2024-10-23
feat: sponsorships

al3mart:
Introduce the PoC for sponsored transactions.

Means to keeps track of sponsorship relations between a sponsor account and a beneficiary account such that sponsors can cover transaction fees for the beneficiaries.

This PoC introduces a simple pallet that maintains sponsorships in a double map
```
/// Register of sponsorships.
/// - K1: Account acting as sponsor.
/// - K2: Sponsored account.
/// - V: If specified, the sponsored amount.
#[pallet::storage]
pub type Sponsorships<T: Config> = StorageDoubleMap<
	_,
	Twox64Concat,
	AccountIdOf<T>,
	Twox64Concat,
	AccountIdOf<T>,
	BalanceOf<T>,
	OptionQuery,
>;
```

Note that the value stored is of type `Weigth` with the idea that each relation could have different limits for how much a sponsor should cover for its beneficiary. Although, that is not explored in this PoC.

Implements two extrinsics to modify said state:

- `pub fn sponsor_account(origin: OriginFor<T>, beneficiary: T::AccountId) -> DispatchResult {}`
- `pub fn remove_sponsorship_for(origin: OriginFor<T>, beneficiary: T::AccountId) -> DispatchResult {}`
- `pub fn set_sponsorship_amount( origin: OriginFor<T>, beneficiary: AccountIdOf<T>, new_amount: BalanceOf<T>) ->  DispatchResult {}`

And the struct `Sponsored` implements `SignedExtension` and it is expected to wrap another implementer of  `SignedExtension` that handles the deduction of the relevant transaction fees.
`Sponsored` implementation takes care of verifying the sponsorship exists and forwards to the wrapper extension the correct account to deduct fees from.
In this PoC the only possible interaction that can be sponsored is a call to a contract.

This pallet is exposed to ink! contracts via pop-api. A contract is included showing how an account can be registered to be sponsor in the example contract "sponsorships": `pop-api/examples/sponsorships/lib.rs`.

PR #362, opened 2024-10-31
feat: sponsorships

al3mart:
Introduce the PoC for sponsored transactions.

Means to keeps track of sponsorship relations between a sponsor account and a beneficiary account such that sponsors can cover transaction fees for the beneficiaries.

This PoC introduces a simple pallet that maintains sponsorships in a double map
```
/// Register of sponsorships.
/// - K1: Account acting as sponsor.
/// - K2: Sponsored account.
/// - V: If specified, the sponsored amount.
#[pallet::storage]
pub type Sponsorships<T: Config> = StorageDoubleMap<
        _,
        Twox64Concat,
        AccountIdOf<T>,
        Twox64Concat,
        AccountIdOf<T>,
        BalanceOf<T>,
        OptionQuery,
>;
```

Note that the value stored is of type `Weigth` with the idea that each relation could have different limits for how much a sponsor should cover for its beneficiary. Although, that is not explored in this PoC.

Implements two extrinsics to modify said state:

- `pub fn sponsor_account(origin: OriginFor<T>, beneficiary: T::AccountId) -> DispatchResult {}`
- `pub fn remove_sponsorship_for(origin: OriginFor<T>, beneficiary: T::AccountId) -> DispatchResult {}`
- `pub fn set_sponsorship_amount( origin: OriginFor<T>, beneficiary: AccountIdOf<T>, new_amount: BalanceOf<T>) ->  DispatchResult {}`

And the struct `Sponsored` implements `SignedExtension` and it is expected to wrap another implementer of  `SignedExtension` that handles the deduction of the relevant transaction fees.
`Sponsored` implementation takes care of verifying the sponsorship exists and forwards to the wrapper extension the correct account to deduct fees from.
In this PoC the only possible interaction that can be sponsored is a call to a contract.

This pallet is exposed to ink! contracts via pop-api. A contract is included showing how an account can be registered to be sponsor in the example contract "sponsorships": `pop-api/examples/sponsorships/lib.rs`.


pallets/sponsorships/src/lib.rs
```diff
@@ -0,0 +1,245 @@
...
+		/// Parameters
+		/// - `beneficiary`: Account to be removed from a sponsorship.
+		#[pallet::call_index(1)]
+		#[pallet::weight(<T as Config>::WeightInfo::remove_sponsorship())]
+		pub fn remove_sponsorship_for(
+			origin: OriginFor<T>,
+			beneficiary: AccountIdOf<T>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+			if <Sponsorships<T>>::take(&who, &beneficiary).is_some() {
```
al3mart, line 141: Original comment at: https://github.com/r0gue-io/pop-node/pull/351#discussion_r1816353978

What would happen to the weight here? Is it just lost?

pallets/sponsorships/src/sponsored.rs
```diff
@@ -0,0 +1,205 @@
...
+	<<T as pallet_revive::Config>::Currency as frame_support::traits::fungible::Inspect<
+		<T as frame_system::Config>::AccountId,
+	>>::Balance: TryFrom<U256>,
+	<<T as pallet_revive::Config>::Time as frame_support::traits::Time>::Moment: Into<U256>,
+{
+	fn is_contracts_call(
+		call: &<T as frame_system::Config>::RuntimeCall,
+	) -> Option<AccountIdOf<T>> {
+		match call.is_sub_type() {
+			Some(pallet_revive::Call::<T>::call { dest, .. }) => {
```
al3mart, line 66: Original comment from https://github.com/r0gue-io/pop-node/pull/351#discussion_r1816359073



Ideally this would be passed in via a generic type parameter which resolves a runtime call to an account id so no tight coupling to revive and configured at runtime level.

No changes required.

AlexD10S (approved):
Great! I tested both sponsorship and incentives, and is looking good!

PR #366, opened 2024-11-01
docs: api readme

Daanvdplas:


pop-api/src/v0/fungibles/README.md
```diff
@@ -1,162 +1,19 @@
-# Fungibles API
+## Fungibles API
 
-The `fungibles` module provides an API for interacting and managing fungible tokens.
+The `fungibles` module provides an api for interacting and managing fungible tokens.
```
al3mart, line 3: https://en.wikipedia.org/wiki/API

pop-api/README.md
```diff
@@ -1,142 +1,64 @@
...
+### Key Features
 
-The Pop API ink! library provides APIs for tailored use cases. The apis are versioned ensuring compatibility across different runtime updates and preserves the functionality of existing contracts.
+- **Versioned Interface**: Provides backward compatibility, ensuring that existing contract functionality remains stable
+  as new features are added to the runtime.
+- **Error Handling**: Supports rich, versioned error types, enabling contracts to receive and interpret any runtime
+  error, making troubleshooting and development easier.
+- **Use Cases**:
+    - [Fungibles](./src/v0/fungibles/README.md): Interacting and managing fungible tokens.
+    - Coming soon:
```
al3mart, line 24: I am not sure we should include this here. There might be better ways of communicating a feature roadmap than in this README.

Until we have something better, I believe it would be better indicating if the specific feature is integrated the API now or it is in development instead of Dec / TBD / Exploring.

al3mart (changes requested):

pop-api/README.md
Daanvdplas: So what are you saying? `In development`?

al3mart: Yeah, that sounds good to me

pop-api/src/v0/fungibles/README.md
Daanvdplas: Changed this already while you were doing the review

PR #371, opened 2024-11-05
fix(testnet): remove old extension code

Daanvdplas:
Tip: right click on the file in rust rover and do `compare with`. Doing this for all the runtime files made me easily find these missing removals.


al3mart (approved):
Thanks!

PR #376, opened 2024-11-06
feat(messaging): callbacks

evilrobot-01:
Adds optional callback specification when using the messaging API. See the messaging contract within pop-api/integration-tests/contracts for usage.


al3mart (approved):
Awesome! This is super cool.

There are some clippy warnings about missing documentation that I think could be easily added.

Aside from that left a comment as I was curious about one detail you pointed out. :)

PR #394, opened 2024-11-27
docs: extension readme

Daanvdplas:


chungquantin (approved):
Nice to have basic examples (similar to the one in mock test). In overall, I'm happy with the doc, very well-written!

Daanvdplas:
You mean adding examples to the readme?

chungquantin:
> You mean adding examples to the readme?

Looking at the existing code in the testnet runtime, there are a lot of noises aside from the direct configuration of the module. 

Something like the code snippet below could be useful for someone that interested to use this in their code. 

```rs
struct ExampleFunction;

...

pub(crate) type Functions = (
	ExampleFunction<WithFuncId<ExampleFuncId>, Runtime>,
);

#[derive(Default)]
pub struct Config;
impl super::Config for Config {
	type Functions = Functions;

	const LOG_TARGET: &'static str = "pop-chain-extension";
}
```

Issue #303, opened 2024-09-25
refactor: apply modular approach to fungible example contract


chungquantin:
Reference: https://github.com/Cardinal-Cryptography/PSP22/blob/38e0b72ab165dc13cd420d9918dff7ae8c931021/lib.rs#L3-L7

PR #401, opened 2024-12-08
chore: apply modular approach to fungible example contract

ndkazu:
### Description
Issue #303
Simplify the contract code by using a modular design, similar to what is shown here: https://github.com/Cardinal-Cryptography/PSP22/blob/38e0b72ab165dc13cd420d9918dff7ae8c931021/lib.rs#L3-L7
## Method
Create a trait `Api` in a different file `traits.rs` with all the function calls using `api::` in `lib.rs`

### DIFF
--- a/pop-api/examples/fungibles/lib.rs
+++ b/pop-api/examples/fungibles/lib.rs
@@ -1,5 +1,6 @@
 #![cfg_attr(not(feature = "std"), no_std, no_main)]
-
+mod traits;
+pub use traits::{Api, ApiImpl};
 use ink::prelude::{string::String, vec::Vec};
 use pop_api::{
 	primitives::TokenId,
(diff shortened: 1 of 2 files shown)


ndkazu:
Work in progress. First commit is already implementing the method described in some parts.

chungquantin:
Hi @ndkazu, thanks for amking progress on this. I would say creating another example 'fungile-modular' in the same folder is better. But let's keep the changes made to the fungible exampel for the code review purpose and refactor after it is approved.

chungquantin:
Discussed with the team and we decided that the issue for modular approach is no longer needed. My suggestion if you are still interested to contribute is to craft a few examples for the PSP22 contract: https://github.com/r0gue-io/pop-node/issues. Let me know if you need any help!

PR #403, opened 2024-12-10
refactor: test check_approval

Daanvdplas:
There were number of missing checks and the for loop was making it very hard to understand. Didn't want to spend too much time on it but I think this is better to follow.

### DIFF
--- a/pallets/nfts/src/tests.rs
+++ b/pallets/nfts/src/tests.rs
@@ -1954,20 +1954,33 @@ fn approval_lifecycle_works() {
 }
 
 #[test]
-fn check_approval_without_deadline_works() {
+fn check_approval_works_without_deadline_works() {
 	new_test_ext().execute_with(|| {
 		let collection_id = 0;
-		let item_id = 42;
 		let collection_owner = account(1);
+		let item_id = 42;
 		let item_owner = account(2);
 		let delegate = account(3);
 
 		Balances::make_free_balance_be(&item_owner, 100);
+		// TODO: we should return a clear error for smart contracts.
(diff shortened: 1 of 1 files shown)


pallets/nfts/src/tests.rs
```diff
@@ -1954,49 +1954,110 @@ fn approval_lifecycle_works() {
...
 		let collection_id = 0;
-		let item_id = 42;
 		let collection_owner = account(1);
+		let item_id = 42;
 		let item_owner = account(2);
 		let delegate = account(3);
 
 		Balances::make_free_balance_be(&item_owner, 100);
+		// TODO: we should return a clear error for smart contracts.
+		// assert_noop!(Nfts::check_collection_approval(&collection_id, &None, &delegate),
```
chungquantin, line 1967: Not very clear to me what is the action plan for this commented out code. Could you explain more?

Daanvdplas: I think we should add a check and error if the collection doesn't exist.

pallets/nfts/src/tests.rs
```diff
@@ -1954,49 +1954,110 @@ fn approval_lifecycle_works() {
 }
 
 #[test]
-fn check_approval_without_deadline_works() {
+fn check_approval_works_without_deadline_works() {
 	new_test_ext().execute_with(|| {
 		let collection_id = 0;
-		let item_id = 42;
 		let collection_owner = account(1);
+		let item_id = 42;
```
chungquantin, line 1961: Note it here because I think my variables did not order alphabetically. Will double check to reorder them!

PR #404, opened 2024-12-11
chore: resolve feedback comments

chungquantin:


pallets/nfts/src/lib.rs
```diff
@@ -1403,7 +1406,7 @@ pub mod pallet {
 		///
 		/// Origin must be the `ForceOrigin`.
 		///
-		/// - `owner`: The account granting approval for delegated transfer.
+		/// - `owner`: The owner of the approvals granted by the `origin`.
```
Daanvdplas, line 1409: This still doesn't make sense. The owner is not granted by the origin. We can change it back as I can not think of a better setence.

pallets/nfts/src/lib.rs
```diff
@@ -1485,7 +1488,7 @@ pub mod pallet {
 		/// Origin must be `ForceOrigin`.
 		///
 		/// Arguments:
-		/// - `owner`: The account cancelling approval for delegated transfer.
```
Daanvdplas, line 1488: The owner who granted approval for delegated transfer

pallets/nfts/src/tests.rs
```diff
@@ -2329,48 +2350,58 @@ fn force_cancel_collection_approvals_work() {
...
-			None
-		);
+			)));
 
-		assert_noop!(
-			Nfts::transfer(RuntimeOrigin::signed(delegate), collection_id, item_id, account(4)),
-			Error::<Test>::NoPermission
-		);
+			assert_noop!(
+				Nfts::transfer(
```
Daanvdplas, line 2396: This is ultimately a test for `transfer`, not for `cancel_approval`

pallets/nfts/src/tests.rs
```diff
@@ -2466,138 +2513,45 @@ fn approvals_limit_works() {
...
-		let (collection, locked_collection) = (0, 1);
-		let item = 42;
+		let (collection_id, locked_collection_id) = (0, 1);
 		let collection_owner = account(1);
-		let item_owner = account(2);
 		let delegate = account(3);
+		let item_id = 42;
+		let item_owner = account(2);
+
+		for origin in [root(), none()] {
```
Daanvdplas, line 2522: Can we add some comments here to explain these different origin tests.

Would also be good to clarify that this test is for `approve_collection_..` and `force_approve_collection_..`. Same for the other two cases (cancel_collection_approval & clear_collection_approvals)

pallets/nfts/src/tests.rs
```diff
@@ -2206,6 +2263,12 @@ fn cancel_approval_works() {
 		);
 
 		assert_ok!(Nfts::cancel_approval(RuntimeOrigin::signed(account(2)), 0, 42, account(3)));
+		assert!(events().contains(&Event::<Test>::ApprovalCancelled {
```
Daanvdplas, line 2266: Again, this should not be tested here. Is this event emitted in `check_approval`?

chungquantin: This is inside cancel_approval

Issue #378, opened 2024-11-07
chore: example contract for `psp22` trait cross contract call 


chungquantin:
- Example to show cross contract call to the Psp22 example contract.

Issue #337, opened 2024-10-10
feat(testnet): examples using psp22 contract 


chungquantin:
We would want to showcase the power of Pop API through and example contracts. With the upcoming #108 , we need some example contracts that use the Fungible contract to build a real DApps. 

## References
- https://github.com/use-ink/ink-examples

ndkazu:
@chungquantin , can this be considered as one example contract using the Fungible contract:
https://github.com/r0gue-io/pop-node/tree/main/pop-api/examples/fungibles

chungquantin:
@ndkazu the contract in examples/fungible is the base contract that follows the PSP22 standard. In the "Use Cases" section of that contract, you can find several examples of how the contract can be used. The PSP22 contract in the example crate can be applied to any case that requires a fungible mechanism:

- Staking: Implementing a staking pool contract where participants deposit fungible tokens to stake, with rewards distributed based on the amount staked.
- Governance: Enabling token holders to participate in decentralized decision-making, such as voting on proposals or network upgrades.
- Rewards Distribution: Distributing tokens as incentives for actions like participating in a network, completing tasks, or using a service.

It’s worth noting that this does not have to be just an example contract but can be used as a real DApps contract as well. In summary, the idea is to have more contracts that leverage examples/fungible for fungible token management across a variety of use cases. So I would recommend you to start with something like an AMM, Staking or DAO contract.

ndkazu:
@chungquantin , `examples/fungibles` being a contract, I assume a new `dao` contract would make use of a cross-contract call, the problems I see is that `fungibles` is a private module, and/or the contract reference is not re-exported : `pub use self::fungible_contract::FungibleContractRef;`. Am I missing something?

chungquantin:
@ndkazu Here is the example to use the fungible contract using `contract_ref!` macro https://github.com/r0gue-io/sub0-ink-challenges-solution/blob/7b48d58a2e727530400b70e3ee7b137458de00d1/challenges/challenge-7/lib.rs#L45

Daanvdplas:
>  The problem I see is that the fungible_contract reference is not re-exported : pub use self::fungibles::FungibleRef

That is a good point though.

Be mindful though that based on your use case you might not need the `Psp22` contract. Lets say for your use case you only need to make a transfer. Then it is much better to just use the `api::transfer` in your contract.

ndkazu:
[...]

PR #407, opened 2024-12-13
feat: psp22 example DAO contract 

ndkazu:
### Description

Issue #337 
This contract implements a Decentralized Autonomous Organization using Psp22. 
The key functionalities include:

- **Membership Management:** It maintains a registry of DAO members.
- **Proposal Lifecycle:** The contract manages the creation, voting, and execution of proposals. Each proposal includes details like description, voting period, vote tallies, execution status, beneficiary, and amount to be awarded.
- **Voting Mechanism:** It implements a voting system where members can cast votes with their balance on proposals. The contract tracks voting periods and maintains vote counts for each proposal.
- **Token Integration:** The DAO is associated with a specific Psp22 token_id. 
- **Governance Parameters:** governance parameters such as voting periods are customizable.
- **Vote Tracking:** The contract keeps track of when members last voted.
- **Proposal Execution:** Once a proposal's voting period ends and it passes, the contract handles the execution of the proposal: Transfer funds to the chosen beneficiary.]

## ToDo
- **Tests** 
- [x]  Create a new Dao works
- [x]   A new member joins the Dao works
- [x]   Member creates a proposal works
- [x]   Members vote system works
- [x]   Proposal enactment works
- [x]   Voting twice for the same proposal fails
- [x]   Vote fails if not a member
- [x]   Vote fails if voting period ended
- [x]   Proposal enactment fails if proposal was rejected
- [x]   Same proposal consecutive claim fails
- **Others**
- [x] Events
- [x] in-code documentation

## Remark

Considering the fact that  [another issue](https://github.com/r0gue-io/pop-node/issues/378) is asking for psp22 examples using cross contract call, I decided to use `api::...` instead of the macro `contract_ref!`.


ndkazu:
@chungquantin , I wrote a first test, but it's failing with an error that I don't fully understand: `Encountered unexpected missing chain extension method: UnregisteredChainExtension`. Can you help please?

chungquantin:
> @chungquantin , I wrote a first test, but it's failing with an error that I don't fully understand: `Encountered unexpected missing chain extension method: UnregisteredChainExtension`. Can you help please?

Are you using [pop-drink](https://github.com/r0gue-io/pop-drink) to write the test?

ndkazu:
> > @chungquantin , I wrote a first test, but it's failing with an error that I don't fully understand: `Encountered unexpected missing chain extension method: UnregisteredChainExtension`. Can you help please?
> 
> Are you using [pop-drink](https://github.com/r0gue-io/pop-drink) to write the test?

No, should I? I will look into it.

ndkazu:
> > > @chungquantin , I wrote a first test, but it's failing with an error that I don't fully understand: `Encountered unexpected missing chain extension method: UnregisteredChainExtension`. Can you help please?
> > 
> > 
> > Are you using [pop-drink](https://github.com/r0gue-io/pop-drink) to write the test?
> 
> No, should I? I will look into it.

I am now using `pop-drink`, I am also using the fungibles example tests to navigate through this. However, at the following line:
`#[drink::contract_bundle_provider]`
I keep getting the following error:
` message: Error building contract: lib section not found`
I will need some help to move forward.

Daanvdplas:
> Encountered unexpected missing chain extension method: UnregisteredChainExtension

This is because when a contract uses the `pop-api`, it interacts with Pop Network, in this case the `fungibles api`; functionality provided by the chain to create, manage and interact with fungible tokens.

As a result, the tests need something like a chain running. This is where `pop-drink` comes in, it provides a `pop` runtime for your contract to interact with, and thus be able to write and run tests for your contract.

Daanvdplas:
>  message: Error building contract: lib section not found
I will need some help to move forward.

Add the following two lines in your `Cargo.toml` (before `features` section):
```
[lib]
path = "src/lib.rs"
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+#[cfg(test)]
+mod tests;
+
+#[ink::contract]
+mod dao {
+	use super::*;
+
+	/// Structure of the proposal used by the Dao governance sysytem
+	#[derive(scale::Decode, scale::Encode, Debug)]
+	#[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
```
Daanvdplas, line 23:
```suggestion
	#[ink::scale_derive(Encode, Decode, TypeInfo)]
	#[cfg_attr(feature = "std", derive(ink::storage::traits::StorageLayout))]

```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+		// Amount of tokens to be awarded to the beneficiary
+		amount: Balance,
+
+		// Identifier of the proposal
+		proposal_id: u32,
+	}
+
+	/// Representation of a member in the voting system
+	#[derive(scale::Decode, scale::Encode)]
+	#[cfg_attr(feature = "std", derive(scale_info::TypeInfo, ink::storage::traits::StorageLayout))]
```
Daanvdplas, line 55:
```suggestion
	#[ink::scale_derive(Encode, Decode, TypeInfo)]
	#[cfg_attr(feature = "std", derive(ink::storage::traits::StorageLayout))]

```

Daanvdplas (commented):
The contract looks great! I left some fixes to make your contract compile and left comment regarding the use of Strings in contracts. Don't forget to use `cargo test --release` to run the `pop drink` tests.

Super awesome that you are developing an example contract. Please note down all the things that are difficult in your journey so that we can improve them!!

chungquantin:
[sc-1859]

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+		/// - `beneficiary` - The account that will receive the payment
+		/// if the proposal is accepted.
+		/// - `amount` - Amount requested for this proposal
+		/// - `description` - Description of the proposal
+		#[ink(message)]
+		pub fn create_proposal(
+			&mut self,
+			beneficiary: AccountId,
+			amount: Balance,
+			description: String,
```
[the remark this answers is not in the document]
ndkazu: Present solution: I am using a `Vec<u8>` type for the description, and a `STRINGLIMIT` constant. The function `create_proposal` takes a `String` type, and after conversion, we check that we are not above the string limit, or we return an error
```suggestion
			beneficiary: AccountId,
```

Daanvdplas:
I can't help you now unfortunately but if I were you I would clone pop-drink locally and see what error is returned by adding the other enum variants [here](https://github.com/r0gue-io/pop-drink/blob/cfbc763962e2cf09240adbfb4be81bfa6150c52b/crates/pop-drink/src/lib.rs#L192) and see which [error](https://github.com/r0gue-io/pop-drink/blob/cfbc763962e2cf09240adbfb4be81bfa6150c52b/crates/drink/drink/src/session/error.rs#L11) is returned to help you figure out what is causing the issue.

ndkazu:
> I can't help you now unfortunately but if I were you I would clone pop-drink locally and see what error is returned by adding the other enum variants [here](https://github.com/r0gue-io/pop-drink/blob/cfbc763962e2cf09240adbfb4be81bfa6150c52b/crates/pop-drink/src/lib.rs#L192) and see which [error](https://github.com/r0gue-io/pop-drink/blob/cfbc763962e2cf09240adbfb4be81bfa6150c52b/crates/drink/drink/src/session/error.rs#L11) is returned to help you figure out what is causing the issue.
> 
> Please let us know what the error is so we can help you to debug.
> 
> I will find some time later today to help!

I found the source of the problem, and if you have an explanation for why it's happening, I am very interested to know about it:
The parameter `description` in my `create_proposal` function had the type `String`. By replacing this type by `Vec<u8>` and during the appropriate conversions,  the problem was solved.  I couldn't find a similar  example anywhere (Missed it probably) so I think this could be a very useful example for future contract builders.

ndkazu:
For tests involving `block_number` the workflow is unclear:
- In the contract: what is the difference between `self.env().block_number()` and `ink::env::block_number::<Environment>()` ???
- In testing environment: `ink::env::block_number::<ink::env::DefaultEnvironment>()` is changed by `ink::env::block_number::<ink::env::DefaultEnvironment>();`, but is not consistent with the contract `block_number` (obtained through an event for example).

As a result, I don't see how to affect the contract `block_number` in the testing environment.

ndkazu:
@Daanvdplas  & @chungquantin , everything is working as expected now, thanks to you!
2 more tests to go!!

ndkazu:
I added all the tests, so I think a review is needed at this point. :pray:

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+		/// Enact a proposal approved by the Dao members
+		///
+		/// # Parameters
+		/// - `proposal_id` - Identifier of the proposal
+		#[ink(message)]
+		pub fn execute_proposal(&mut self, proposal_id: u32) -> Result<(), Error> {
+			let vote_end = self.proposals.get(proposal_id).ok_or(Error::ProposalNotFound)?.vote_end;
+
+			// Check the voting period
+			if self.env().block_number() <= vote_end {
```
chungquantin, line 233:
```suggestion
			if self.env().block_number() <= proposal.vote_end {
```

chungquantin: Please see line 230, this variable is unnecessary

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+				return Err(Error::VotingPeriodNotEnded);
+			}
+
+			// If we've passed the checks, now we can mutably borrow the proposal
+			// let proposal_id_usize = proposal_id as usize;
+			let mut proposal = self.proposals.get(proposal_id).ok_or(Error::ProposalNotFound)?;
```
chungquantin, line 239: See line 230, why we need to read the `proposals` storage again? This add extra cost to the smart contract function.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+
+			if proposal.executed == true {
+				return Err(Error::ProposalAlreadyExecuted);
+			}
+
+			if proposal.yes_votes > proposal.no_votes {
+				let contract = self.env().account_id();
+
+				// Execute the proposal
+				let treasury_balance = api::balance_of(self.token_id, contract).unwrap_or_default();
```
chungquantin, line 249: What happens if there is no token exist? Can simply refactoring to combine with the code at line 250 to throw an inline error.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+				if treasury_balance < proposal.amount {
+					// Not enough tokens in treasury to distribute
+					return Err(Error::TryAgainLater);
```
chungquantin, line 252:
```suggestion
					return Err(Error::InsufficientBalance);
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+				return Err(Error::ProposalAlreadyExecuted);
+			}
+
+			if proposal.yes_votes > proposal.no_votes {
+				let contract = self.env().account_id();
+
+				// Execute the proposal
+				let treasury_balance = api::balance_of(self.token_id, contract).unwrap_or_default();
+				if treasury_balance < proposal.amount {
+					// Not enough tokens in treasury to distribute
```
chungquantin, line 251:
```suggestion
```

If we need a comment here, that means the error is not expressive enough and need to be updated.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+			Ok(())
+		}
+
+		#[ink(message)]
+		pub fn get_member(&mut self, account: AccountId) -> Member {
+			self.members.get(account).unwrap_or(Member { voting_power: 0, last_vote: 0 })
+		}
+
+		#[ink(message)]
+		pub fn get_block_number(&mut self) -> Option<BlockNumber> {
```
chungquantin, line 316: Is this required? We don't need a getter method for a block number.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+				to: Some(contract),
+				value: amount,
+			});
+
+			Ok(())
+		}
+
+		#[ink(message)]
+		pub fn get_member(&mut self, account: AccountId) -> Member {
+			self.members.get(account).unwrap_or(Member { voting_power: 0, last_vote: 0 })
```
chungquantin, line 312:
```suggestion
			self.members.get(account).unwrap_or(Default::default())
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+		}
+
+		#[ink(message)]
+		pub fn get_block_number(&mut self) -> Option<BlockNumber> {
+			Some(self.env().block_number())
+		}
+
+		#[ink(message)]
+		pub fn positive_votes(&mut self, proposal_id: u32) -> Option<Balance> {
```
chungquantin, line 321: Wrong implementation. If there is no proposal exists, you need to return `None` instead of `unwrap`.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+			let proposal = &self.proposals.get(proposal_id).unwrap();
+			Some(proposal.yes_votes)
+		}
+
+		#[ink(message)]
+		pub fn negative_votes(&mut self, proposal_id: u32) -> Option<Balance> {
```
chungquantin, line 327: Same with `positive_votes`

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+			let proposal = &self.proposals.get(proposal_id).unwrap();
+			Some(proposal.no_votes)
+		}
+	}
+
+	#[derive(Debug, PartialEq, Eq)]
+	#[ink::scale_derive(Encode, Decode, TypeInfo)]
+	pub enum Error {
```
chungquantin, line 335: Requires comments for the `Error` because it is public.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+		ArithmeticOverflow,
```
chungquantin, line 336: This is not needed. 
```suggestion
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+		ProposalNotFound,
+		VotingPeriodEnded,
+		NotAMember,
```
chungquantin, line 339:
```suggestion
		MemberNotFound,
```
or 
```suggestion
		UnknownMember,
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+		AlreadyVoted,
+		VotingPeriodNotEnded,
+		ProposalAlreadyExecuted,
```
chungquantin, line 342:
```suggestion
		ProposalExecuted,
```
to keep it consistent with `ProposalRejected`

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+		ProposalRejected,
+		StringLimitReached,
```
chungquantin, line 344: Suggests a better error.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+		TryAgainLater,
```
chungquantin, line 345:
```suggestion
```
Not a good error.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+		Psp22(Psp22Error),
+	}
+
+	impl From<Psp22Error> for Error {
+		fn from(error: Psp22Error) -> Self {
+			Error::Psp22(error)
+		}
+	}
+
+	impl From<Error> for Psp22Error {
```
chungquantin, line 356: If your implementation is correct then you don't need this conversion.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+	#[ink(storage)]
+	pub struct Dao {
+		// Funding proposals
+		proposals: Mapping<u32, Proposal>,
+
+		// Mapping of AccountId to Member structs, representing DAO membership.
+		members: Mapping<AccountId, Member>,
+
+		// Mapping tracking the last time each account voted.
+		last_votes: Mapping<AccountId, BlockNumber>,
```
chungquantin, line 81: I don't see this field is used anywhere. Why do we need it?

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+				return Err(Error::AlreadyVoted);
+			}
+
+			if approve {
+				proposal.yes_votes = proposal.yes_votes.saturating_add(member.voting_power);
+			} else {
+				proposal.no_votes = proposal.no_votes.saturating_add(member.voting_power);
+			}
+
+			debug_assert!(proposal.yes_votes > 0);
```
chungquantin, line 209: This is incorrect. Imagine if someone use this contract for production use, this line of code will panic instead of returning a meaningful error.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+
+			self.proposals.remove(proposal_id);
```
chungquantin, line 211: Why `remove`? `insert` overrides a pre-existing value with the new one. https://docs.rs/ink_storage/latest/ink_storage/struct.Mapping.html#method.insert

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+				// Execute the proposal
+				let treasury_balance = api::balance_of(self.token_id, contract).unwrap_or_default();
+				if treasury_balance < proposal.amount {
+					// Not enough tokens in treasury to distribute
+					return Err(Error::TryAgainLater);
+				}
+
+				api::transfer(self.token_id, proposal.beneficiary, proposal.amount)
+					.map_err(Psp22Error::from)?;
+				self.env().emit_event(Transfer {
```
chungquantin, line 257: This is incorrect, the api is not expected to used directly but the PSP22 contract. I saw you followed the code in the psp22 example contract but that is not the expectation.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+				});
+				self.env().emit_event(Approval {
+					owner: contract,
+					spender: contract,
+					value: proposal.amount,
+				});
+
+				proposal.executed = true;
+
+				self.proposals.remove(proposal_id);
```
chungquantin, line 270: no remove.

chungquantin (changes requested):
Unfortunately there are a lot of codes need to be changed first. I haven't reviewed the tests and docs, the overall logic is incorrect. The DAO example contract is expected to cross contract calling the `PSP22` contract not using the pop-api fungible directly. See https://github.com/r0gue-io/sub0-ink-challenges/blob/main/challenges/challenge-7/lib.rs

pop-api/examples/dao/src/lib.rs
Daanvdplas: debug is not used in production build

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+			}
+
+			let member = self.members.get(caller).ok_or(Error::NotAMember)?;
+
+			if member.last_vote >= proposal.vote_start {
+				return Err(Error::AlreadyVoted);
+			}
+
+			if approve {
+				proposal.yes_votes = proposal.yes_votes.saturating_add(member.voting_power);
```
chungquantin, line 204: No action needed. This could be improved by refactoring a bit: 

```rs
let mute votes = match approve {
  true => proposal.yes_votes,
  false => proposal.no_votes
};

votes.saturating_add(member.voting_power);

```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+		pub vote_end: BlockNumber,
+
+		// Balance representing the total votes for this proposal
+		pub yes_votes: Balance,
+
+		// Balance representing the total votes against this proposal
+		pub no_votes: Balance,
+
+		// Flag that indicates if the proposal was executed
+		pub executed: bool,
```
chungquantin, line 45: Will be cleaner to have an enum type for the `ProposalStatus`.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,364 @@
...
+
+		// AccountId of the recipient of the proposal
```
chungquantin, line 47:
```suggestion
		// The recipient of the proposal
```

pop-api/examples/dao/src/lib.rs
ndkazu: Oups! forgot to update this field in the `vote` function, corrected!

ndkazu:
@chungquantin can you test `cargo test` in the dao folder please. for me it's hanging with the ` Blocking waiting for file lock on build directory`.

ndkazu:
> @chungquantin can you test `cargo test` in the dao folder and let me know if it works for you please. for me it's hanging with the ` Blocking waiting for file lock on build directory`.
Hello, I have the same problem with the `fungible` example: `cargo test` hangs forever at the following step:
```rust
   Compiling pop-drink v0.1.0 (https://github.com/r0gue-io/pop-drink#cfbc7639)
 [==] Checking clippy linting rules
    Blocking waiting for file lock on build directory
    Building [=======================> ] 785/786: fungibles(test)      
```
If this is only me, then my dev environment might be the problem, but I need confirmation on this. cannot run any test at the moment... :disappointed:

chungquantin:
@ndkazu pop-drink tests must be run with `--release` flag

```
cargo test --release
```

ndkazu:
> @ndkazu pop-drink tests must be run with `--release` flag
> 
> ```
> cargo test --release
> ```

Thanks! forgot about that one.

ndkazu:
> Aside from that, right now the proposal is only specific to the transfer functionality, it would be great to see the proposal that can serves more use cases. Maybe you can consider using `RuntimeCall`.

@chungquantin , I've been looking into [chain_extension](https://github.com/use-ink/ink-examples/tree/main/rand-extension) and [call_runtime](https://github.com/use-ink/ink-examples/blob/main/call-runtime/lib.rs) examples and from what I understood (correct me if I'm wrong...), you need to know which extrinsic in which pallet you're going to use to make the set-up => customizable runtime calls as proposals for the dao governance do not seem to be possible from what I see. If you have any suggestion on this front...

ndkazu:
@chungquantin  & @Daanvdplas,  review requested: struggling with the use of a RuntimeCall for proposal enactment.

ndkazu:
> @chungquantin & @Daanvdplas, review requested: struggling with the use of a RuntimeCall for proposal enactment.

RuntimeCall version now working => Good step for a code review.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+
+		// Proposals created in the history of the Dao
+		proposals_created: u32,
+	}
+
+	/// Defines an event that is emitted
+	/// every time a member voted.
+	#[derive(Debug)]
+	#[ink(event)]
+	pub struct Voted {
```
chungquantin, line 154:
```suggestion
	pub struct Vote {
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+		pub status: ProposalStatus,
+
+		// Identifier of the proposal
+		pub proposal_id: u32,
+
+		// Information relative to voting
+		pub votes_infos: Option<Votes>,
+
+		// Information relative to proposal execution if approved
+		pub transaction_infos: Option<Transaction>,
```
chungquantin, line 61:
```suggestion
		pub transaction: Option<Transaction>,
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+		pub voting_power: Balance,
+
+		// Keeps track of the last vote casted by the member
+		pub last_vote: BlockNumber,
+	}
+
+	#[derive(Debug, Clone)]
+	#[ink::scale_derive(Encode, Decode, TypeInfo)]
+	#[cfg_attr(feature = "std", derive(ink::storage::traits::StorageLayout))]
+	pub struct Votes {
```
chungquantin, line 101:
```suggestion
	pub struct VoteRound {
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+		// Beginnning of the voting period for this proposal
+		pub vote_start: BlockNumber,
```
chungquantin, line 103:
```suggestion
		pub start: BlockNumber,
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+
+		// End of the voting period for this proposal
+		pub vote_end: BlockNumber,
```
chungquantin, line 106:
```suggestion
		pub end: BlockNumber,
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+		last_votes: Mapping<AccountId, BlockNumber>,
+
+		// Duration of the voting period
+		voting_period: BlockNumber,
+
+		// Identifier of the Psp22 token associated with this DAO
+		token_id: TokenId,
+
+		// Proposals created in the history of the Dao
+		proposals_created: u32,
```
chungquantin, line 147:
```suggestion
		proposal_count: u32,
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+		VotingPeriodNotEnded,
+
+		/// This proposal has already been Executed
+		ProposalExecuted,
+
+		/// This proposal has been Rejected
+		ProposalRejected,
+
+		/// The proposal description is too long
+		ExceedeMaxDescriptionLength,
```
chungquantin, line 424:
```suggestion
		MaxDescriptionLengthReached,
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+
+		/// There are not enough funds in the Dao treasury
+		NotEnoughFundsAvailable,
+
+		/// The contract creation failed, a new contract is needed
+		WrongContract,
+
+		/// The Runtime Call failed
+		CallRuntimeFailed,
```
chungquantin, line 433:
```suggestion
		ProposalExecutionFailed,
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+		#[ink(message)]
+		pub fn create_proposal(
+			&mut self,
+			beneficiary: AccountId,
+			amount: Balance,
+			mut description: Vec<u8>,
+		) -> Result<(), Error> {
+			let caller = self.env().caller();
+			let contract = self.env().account_id();
+			self.proposals_created = self.proposals_created.saturating_add(1);
```
chungquantin, line 209: I suggest moving this line to the bottom of the function, above the event.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+
+			if description.len() >= u8::MAX.into() {
+				return Err(Error::ExceedeMaxDescriptionLength);
+			}
+
+			let mut proposal =
+				Proposal { proposal_id: self.proposals_created, ..Default::default() };
+			proposal.description.append(&mut description);
+			let transaction_infos = Transaction { beneficiary, amount };
```
chungquantin, line 218:
```suggestion
			let transaction = Transaction { beneficiary, amount };
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+			proposal.transaction_infos = Some(transaction_infos);
```
chungquantin, line 219:
```suggestion
			proposal.transaction = Some(transaction);
```

chungquantin (commented):
So far so good! Haven't reviewed the test yet.

pop-api/examples/dao/src/lib.rs
ndkazu: I use it to change the proposal below.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,455 @@
...
+				let _treasury_balance = match api::balance_of(self.token_id, contract) {
+					Ok(val) if val > transaction_infos.amount => val,
+					_ => {
+						return Err(Error::NotEnoughFundsAvailable);
+					},
+				};
+
+				// RuntimeCall transfer, you must comment api::transfer_from() below
+				let _ = self.env()
+					.call_runtime(&RuntimeCall::Fungibles(FungiblesCall::TransferFrom {
```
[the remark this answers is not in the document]
ndkazu: @chungquantin , can you point me to an example for this please.

ndkazu:
Hello @chungquantin &  @Daanvdplas , I am still trying to solve this and and need some feedback please: the `RuntimeCall` enum conversion to bytes requires the trait Serialize to be implemented. the problem is that this trait is not implemented for `ink_primitive::AccountId`, which itself is used in the call I am trying to test...
I am just trying to do an enum-->bytes-->string conversion to be able to run the test.

Daanvdplas:
One additional point is that I don't understand the incentive for joining the dao. If I buy myself in and all the dao's tokens are spent to e.g. projects, what do I win from this? It could be a charity funding dao perhaps, this would have to be included in the use case however.

I don't want to overcomplicate the example at all, but the use case should make sense and be complete otherwise I think it would demotivate people to look into the example.

ndkazu:
> What an amazing work! Bravo for writing the first contract using the pop api and write tests using pop drink, you are the first one!!
> 
> Before doing a proper review on the contract there needs to be agreed on what the example contract should be. For an example contract it is important to define a clear use case. Right now there are few options we can finalise towards:
> 
> 1. A dao where people can vote on proposals where the native token is transferred to a given account (makes me think of the tips pallet from the sdk)
> 2. A dao where people can vote on proposals where any kind of RuntimeCall is called
> 3. ...
> 
> Option 1 could for example turn into one of the following examples:
> 
> * Community Fund Management: Manage a treasury where members can propose and vote on funding community projects.
> * Grant Distribution: Allocate funds to developers or projects based on proposals and voting.
> 
> Option 2 gets a lot more complex to find an easy to explain use case. But perhaps OpenGov in a contract. This however will require more work and might be too complex for an example.
> 
> It doesn't have to be complicated or perfect at all, but people should understand what the contract example is about by looking at the title or the introduction, for them to get excited to learn about this specific example :)
> 
> I will be following this issue closer so that we can agree on the use case for the example. Then perhaps it requires some changes, but then I am ready to give it a review!!
> 
> Super excited to see this example in the repo!

Thank you for the feedback, @Daanvdplas!! I agree that as an example, the contract should be relatively simple, using too many advanced features in one contract could be daunting for people "Just diving in...". 
In the previous  [commit ](https://github.com/r0gue-io/pop-node/blob/c357a0755706a9bbe01adc3522854370f366cf74/pop-api/examples/dao/src/lib.rs)I'm actually going in the **option_1** direction. The latest commit is the one going in the **option_2** direction ⇒ maybe it should be 2 different PRs?
As for the incentive to join the PR, I feel like it is a bit like what is being done [here ](https://github.com/polkadot-fellows/runtimes/issues/375) with **the Optimistic Project Funding Pallet** in substrate (still in work...): Members of the community vote for the funding of projects they want to support. The funds themselves come from a percentage of the inflation. So basically this contract (**Option_1**) is a simplified/rough version of the OPF_pallet (A Staking contract would be used to provide an incentive...).

Daanvdplas:
How do you become a member of the community that holds the funds?

ndkazu:
> How do you become a member of the community that holds the funds?

In OPF you only have to be a DOT holder: you lock  some funds as voting power during the voting period.
I also wanted to add that in this kind of Dao (**Option_1**), the incentive is mainly provided by the project requesting funding and not by the Dao itself: What kind of services/Token use-case/benefits,  will they provide to the Dao_Token holders and the Eco-system? 
I just finished implementing the customized runtime calls, which corresponds to **option_2**, but we need to make a decision in order to finalize the example documentation.

Daanvdplas:
I think we should go with option 1. Perhaps we can find a popular dao contract example that functions the same / similar and see how they tell the story of the use case and tell it in a way that it doesn't matter that it is perfectly thought out, as that is the feeling I have now (regarding the incentive to join this dao). Another solution is to add ideas for how to solve this and provide it in the readme!

Let me know if you agree with taking option 1 then I can do a review.

ndkazu:
> I think we should go with option 1. Perhaps we can find a popular dao contract example that functions the same / similar and see how they tell the story of the use case and tell it in a way that it doesn't matter that it is perfectly thought out, as that is the feeling I have now (regarding the incentive to join this dao). Another solution is to add ideas for how to solve this and provide it in the readme!
> 
> Let me know if you agree with taking option 1 then I can do a review.

I opened another [PR ](https://github.com/r0gue-io/pop-node/pull/422)for **option_1** as the latest commit here is oriented toward **option_2**, we can keep it for later.

pop-api/examples/dao/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+serde = { version = "1.0.114", default-features = false }
+sp-io = { version = "23.0.0", default-features = false, features = ["disable_allocator", "disable_oom", "disable_panic_handler"] }
+sp-runtime = { version = "24.0.0", default-features = false }
+
+[dev-dependencies]
+drink = { package = "pop-drink", git = "https://github.com/r0gue-io/pop-drink" }
+env_logger = { version = "0.11.3" }
+serde_json = "1.0.114"
+
+# TODO: due to compilation issues caused by `sp-runtime`, `frame-support-procedural` and `staging-xcm` this dependency
```
chungquantin, line 21: Not a blocker but these crates [were removed in `main`](https://github.com/r0gue-io/pop-node/commit/8fb843fa7bba93885369fd733fa1531715bdb01e#diff-b6d07bdd90a1404acafadd2f52c4b65a417c7b87dd6c5f7c9804026d06a388d1L20) so this will be resolved by rebasing to latest commits in `main`.

pop-api/examples/dao/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+## Dao Workflow
+- **Subscription**: The first step is membership: users use their funds* to join the Dao, become members, and determine their voting power, as membership gives them the right to vote on the use of the Dao Treasury.
+- **Proposal**: Dao members can create spending proposals for the treasury. At the moment, the voting period for the proposal is given by the proposal creator, but this could be a Dao parameter, determined by the creator of the Dao contract.
+- **Vote**: Members of the Dao can vote for or against a given proposal, through the selection of the corresponding proposal ID. The vote has to be cast within the voting period of the selected proposal.
+- **Proposal enactment**: After the end of the voting period, If the proposal has been accepted by the Dao Members, the proposal can be enacted, i.e. funds can be claimed/transferred to the account specified as the beneficiary in the proposal. Any member can claim the reward for the winning proposal.
+
+*It is assumed that after Dao & associated token creation, potential members own the Dao Token through airdrop, token purchase, etc... (These mechanisms are not part of the contract). 
+
```
chungquantin, line 22:
```suggestion

## Test with Pop Drink

Since this contract interacts directly with Pop’s runtime through the Pop API, it requires [Pop Drink](https://github.com/r0gue-io/pop-drink) for testing. See how the contract is tested in [tests](./tests.rs).

## Support

Be part of our passionate community of Web3 builders. [Join our Telegram](https://t.me/onpopio)!

Feel free to raise issues if anything is unclear, you have ideas or want to contribute to Pop! Examples using the fungibles API are always welcome!

For any questions related to ink! you can also go to [Polkadot Stack Exchange](https://polkadot.stackexchange.com/) or
ask the [ink! community](https://t.me/inkathon/1).
```

pop-api/examples/dao/README.md
```diff
@@ -0,0 +1,53 @@
+# PSP22 DAO contract example
+
```
chungquantin, line 2: A brief description for the contract here would be very helpful. Reference from [PSP22 example contract](https://github.com/r0gue-io/pop-node/tree/main/pop-api/examples/fungibles#psp22-fungible-token-with-pop-api). And if it possible, a mention that this DAO contract is built on top of Pop API is super meaningful to us.

pop-api/examples/dao/README.md
```diff
@@ -0,0 +1,53 @@
...
+## Description
+This contract implements a Decentralized Autonomous Organization using Psp22.
```
chungquantin, line 4:
```suggestion
This contract implements a Decentralized Autonomous Organization with token-based voting system built with Pop API.
```

pop-api/examples/dao/README.md
```diff
@@ -0,0 +1,53 @@
...
+The key functionalities include:
+- **Membership Management**: It maintains a registry of DAO members.
+- **Proposal Lifecycle**: The contract manages the creation, voting, and execution of proposals. Each proposal includes details like description, voting period, vote tallies, execution status, beneficiary, and amount to be awarded.
+- **Voting Mechanism**: It implements a voting system where members can vote with their balance on proposals. The contract tracks voting periods and maintains vote counts for each proposal.
+- **Token Integration**: The DAO is associated with a specific Psp22 token_id.
```
chungquantin, line 9: Would be better to add a short description how the Pop API is used for the token-based voting system. I don't think we need to explain the relationship between token <> DAO here.

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,474 @@
...
+		pub proposal_id: u32,
+
+		// Information relative to voting
+		pub round: Option<VoteRound>,
+
+		// Information relative to proposal execution if approved
+		pub transaction: Option<Transaction>,
+
+		// Call top be executed
+		pub call: Option<Vec<u8>>,
```
chungquantin, line 72: Nice 💪

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,474 @@
...
+		/// the voting power of the member.
+
+		#[ink(message)]
+		pub fn join(&mut self, amount: Balance) -> Result<(), Error> {
+			let caller = self.env().caller();
+			let contract = self.env().account_id();
+			api::transfer_from(self.token_id, caller, contract, amount)
+				.map_err(Psp22Error::from)?;
+			let member =
+				self.members.get(caller).unwrap_or(Member { voting_power: 0, last_vote: 0 });
```
chungquantin, line 389:
```suggestion
				self.members.get(caller).unwrap_or(Member::default());
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,474 @@
...
+
+		/// Allows a user to become a member of the Dao
+		/// by transferring some tokens to the DAO's treasury.
+		/// The amount of tokens transferred will be stored as the
+		/// voting power of this member.
+		///
+		/// # Parameters
+		/// - `amount` - Balance transferred to the Dao and representing
+		/// the voting power of the member.
+
```
chungquantin, line 381:
```suggestion
```

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,474 @@
...
+		#[ink(message)]
+		pub fn join(&mut self, amount: Balance) -> Result<(), Error> {
```
chungquantin, line 383: As I understand, existing members can use this method to add more voting power?

pop-api/examples/dao/src/lib.rs
```diff
@@ -0,0 +1,474 @@
...
+		/// The Runtime Call is missing or invalid
+		NoValidRuntimeCall,
+
+		/// PSP22 specific error
+		Psp22(Psp22Error),
+	}
+
+	impl From<Psp22Error> for Error {
+		fn from(error: Psp22Error) -> Self {
+			Error::Psp22(error)
```
chungquantin, line 460:
```suggestion
			Self::Psp22(error)
```

pop-api/examples/dao/src/tests.rs
```diff
@@ -0,0 +1,549 @@
...
+
+	let now = session.sandbox().block_number();
+	session.set_actor(CHARLIE);
+	// Charlie vote
+	assert_ok!(vote(&mut session, 1, true));
+	assert_last_contract_event!(
+		&session,
+		Vote { who: Some(account_id_from_slice(&CHARLIE)), when: Some(now) }
+	);
+	// Alice vote
```
chungquantin, line 318:
```suggestion
	// ALICE votes
```

To keep it consistent with the below comments.

pop-api/examples/dao/src/tests.rs
```diff
@@ -0,0 +1,549 @@
...
+		from: account_id_from_slice(&contract),
+		to: account_id_from_slice(&BOB),
+		value: amount,
+	})
+	.encode(&mut call_f);
+	assert_ok!(create_proposal(&mut session, BOB, amount, description, call_f));
+
+	let now = session.sandbox().block_number();
+	session.set_actor(CHARLIE);
+	// Charlie vote
```
chungquantin, line 312:
```suggestion
	// CHARLIE votes
```

To keep it consistent with the below comments.

pop-api/examples/dao/src/tests.rs
```diff
@@ -0,0 +1,549 @@
...
+		Vote { who: Some(account_id_from_slice(&CHARLIE)), when: Some(now) }
+	);
+	// Alice vote
+	session.set_actor(ALICE);
+	assert_ok!(vote(&mut session, 1, true));
+	assert_last_contract_event!(
+		&session,
+		Vote { who: Some(account_id_from_slice(&ALICE)), when: Some(now) }
+	);
+	// BOB vote
```
chungquantin, line 325:
```suggestion
	// BOB votes
```

pop-api/examples/dao/src/tests.rs
```diff
@@ -0,0 +1,549 @@
...
+			admin: account_id_from_slice(&contract),
+		}
+	);
+
+	// Another proposal created by Bob
+	let description_bis =
+		"Funds for creation of another Dao contract".to_string().as_bytes().to_vec();
+	session.set_actor(BOB);
+	let mut call_f = Vec::new();
+	let call0 = RuntimeCall::Fungibles(FungiblesCall::TransferFrom {
```
chungquantin, line 146:
```suggestion
	RuntimeCall::Fungibles(FungiblesCall::TransferFrom {
```

Seems like the variable is not used anywhere

pop-api/examples/dao/src/tests.rs
```diff
@@ -0,0 +1,549 @@
...
+		token: TOKEN,
+		from: account_id_from_slice(&contract),
+		to: account_id_from_slice(&ALICE),
+		value: amount,
+	})
+	.encode(&mut call_f);
```
chungquantin, line 152:
```suggestion
	.encode();
```

pop-api/examples/dao/src/tests.rs
```diff
@@ -0,0 +1,549 @@
...
+		session,
+		"get_proposal",
+		vec![proposal_id.to_string()],
+		None,
+	)
+	.unwrap()
+}
+
+fn prepare_dao(session: &mut Session<Pop>, contract: AccountId) -> Result<(), Error> {
+	assert_ok!(session.sandbox().mint_into(&TOKEN, &ALICE, AMOUNT));
```
chungquantin, line 536: Would be cleaner to have a for loop for this.

chungquantin (changes requested):

chungquantin:
@ndkazu The example contract looks great and quite close to finish. I left a few comments to improve the README and the drink tests.

pop-api/examples/dao/src/lib.rs
ndkazu: True

pop-api/examples/dao/src/tests.rs
ndkazu: I get the error message: `this method takes 1 argument but 0 arguments were supplied` if I don't do this

ndkazu:
@chungquantin & @Daanvdplas , I corrected both PRs.

Daanvdplas:
Closing this to narrow scope

PR #409, opened 2024-12-15
chore: ink v5.1

Daanvdplas:
In v5.0 there was an ugly warning from the linter which is fixed in v5.1. Also good to uplift the version.

### DIFF
--- a/pop-api/src/v0/fungibles/errors.rs
+++ b/pop-api/src/v0/fungibles/errors.rs
@@ -129,6 +129,7 @@ mod tests {
 		},
 		StatusCode,
 	};
+	use ink::scale::{Decode, Encode};
 
 	fn error_into_status_code(error: Error) -> StatusCode {
 		let mut encoded_error = error.encode();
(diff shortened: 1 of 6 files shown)


chungquantin:
Should we use git path to latest main branch then?

Daanvdplas:
> Should we use git path to latest main branch then?

Why?

chungquantin:
> > Should we use git path to latest main branch then?
> 
> Why?

To avoid updating the version every time there's a new update. Anyway, this could brings unexpected issues so ignore my comment.

pop-api/src/v0/fungibles/errors.rs
```diff
@@ -130,6 +129,7 @@ mod tests {
 		},
 		StatusCode,
 	};
+	use ink::scale::{Decode, Encode};
```
chungquantin, line 133: This change seems unnecessary

chungquantin (approved):
LGTM! Contracts compiled and built.

pop-api/src/v0/fungibles/errors.rs
Daanvdplas: This is required to run the tests

AlexD10S (approved):
Everything looks great! I tested deploying a smart contract and interacting with it, and everything is working perfectly.

PR #412, opened 2024-12-18
chore: upgrade stable2409

Daanvdplas:
Uplifts pop to 2409. Some noteworthy changes had to be made for ismp (upstream changes) and the sdk. I left comments at the relevant places to have a discussion.

Notes:
- Pallet revive will be added in a separate PR.
- In some Cargo.toml files dependencies were added to the features. This was done while trying to find the cause of the compilation error (pop common). I left them in as they might cause issues in the future.


integration-tests/Cargo.toml
```diff
@@ -102,6 +106,7 @@ std = [
 	"sp-consensus-grandpa/std",
 	"sp-core/std",
 	"sp-runtime/std",
+	"tracing-subscriber/std",
 	"westend-runtime-constants?/std",
 	"westend-runtime?/std",
 	"xcm-executor/std",
```
Daanvdplas, line 112: These dependencies were added when trying to find a compilation error. Leaving them in as they might cause problems in the future.

pallets/api/src/fungibles/tests.rs
```diff
@@ -341,7 +341,7 @@ fn start_destroy_works() {
 		// Check that the token is not live after starting the destroy process.
 		assert_noop!(
 			Assets::mint(signed(ALICE), token, ALICE, 10 * UNIT),
-			AssetsError::AssetNotLive
+			TokenError::UnknownAsset
```
Daanvdplas, line 344: Test was failing due to the following https://github.com/paritytech/polkadot-sdk/pull/3286

Concerning to me is that this would have broken a smart contract. Presumably contracts won't often match on a specific error but if so, the contract could behave unexpected.

runtime/common/Cargo.toml
```diff
@@ -27,6 +27,15 @@ polkadot-primitives.workspace = true
...
 ]
-std = [ "frame-support/std", "sp-runtime/std" ]
+std = [
+	"codec/std",
+	"frame-support/std",
+	"parachains-common/std",
+	"polkadot-primitives/std",
+	"scale-info/std",
+	"sp-runtime/std",
+]
```
Daanvdplas, line 41: Adding these dependencies resolved the compilation issues I had.

pop-api/integration-tests/src/fungibles/mod.rs
```diff
@@ -514,9 +514,9 @@ fn mint_works() {
 		assert_eq!(balance_after_mint, balance_before_mint + amount);
 		// Account can not hold more tokens than Balance::MAX.
 		assert_eq!(mint(&addr, token, &BOB, Balance::MAX,), Err(Arithmetic(Overflow)));
-		// Token is not live, i.e. frozen or being destroyed.
+		// Token is being destroyed.
 		assets::start_destroy(&addr, token);
-		assert_eq!(mint(&addr, token, &BOB, amount), Err(Module { index: 52, error: [16, 0] }));
+		assert_eq!(mint(&addr, token, &BOB, amount), Err(Token(UnknownAsset)));
```
Daanvdplas, line 519: Same error that changed as above.

runtime/devnet/src/config/contracts.rs
```diff
@@ -16,6 +16,7 @@ fn schedule<T: pallet_contracts::Config>() -> pallet_contracts::Schedule<T> {
 	pallet_contracts::Schedule {
 		limits: pallet_contracts::Limits {
 			runtime_memory: 1024 * 1024 * 1024,
+			validator_runtime_memory: 2 * 1024 * 1024 * 1024,
```
Daanvdplas, line 19: The integrity test was failing:
```
failures:

---- __construct_runtime_integrity_test::runtime_integrity_tests stdout ----
thread '__construct_runtime_integrity_test::runtime_integrity_tests' panicked at /Users/R0GUE/.cargo/registry/src/index.crates.io-6f17d22bba15001f/pallet-contracts-38.0.0/src/lib.rs:707:13:
Maximal storage size 143942352 exceeds the storage limit 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

@peterwht do you remember why you put this in the runtimes. The SDK uses the [default](https://github.com/paritytech/polkadot-sdk/blob/743dc632fd6115b408376a6e4efe815bd804cd52/substrate/bin/node/runtime/src/lib.rs#L1344)

runtime/devnet/src/config/ismp.rs
```diff
@@ -17,7 +17,7 @@ impl pallet_ismp::Config for Runtime {
 	type Coprocessor = Coprocessor;
 	type Currency = Balances;
 	type HostStateMachine = HostStateMachine;
-	type Mmr = pallet_ismp::NoOpMmrTree<Self>;
+	type OffchainDB = ();
```
Daanvdplas, line 20: Confirmed with polytope labs team that this is oke to use in stead of the [Mmr pallet](https://github.com/polytope-labs/hyperbridge/blob/a2cb646d3aee28e04755c704bba1cf7a3672b333/parachain/runtimes/nexus/src/ismp.rs#L113)

Daanvdplas, line 20: Confirmed with polytope labs team that this is oke to use in stead of the [Mmr pallet](https://github.com/polytope-labs/hyperbridge/blob/a2cb646d3aee28e04755c704bba1cf7a3672b333/parachain/runtimes/nexus/src/ismp.rs#L113)

runtime/devnet/src/config/ismp.rs
```diff
@@ -46,7 +46,7 @@ impl Get<StateMachine> for HostStateMachine {
 #[derive(Default)]
 pub struct Router;
 impl IsmpRouter for Router {
-	fn module_for_id(&self, id: Vec<u8>) -> Result<Box<dyn IsmpModule>, Error> {
-		Err(Error::ModuleNotFound(id))
+	fn module_for_id(&self, id: Vec<u8>) -> Result<Box<dyn IsmpModule>, anyhow::Error> {
+		Err(anyhow::anyhow!("Module not found: {:?}", id))
```
Daanvdplas, line 50: Changed to `anyhow` error type

pallets/api/src/fungibles/tests.rs
Daanvdplas: This error only changed for the `mint` function. 

The reason why it is concerning to me is that I don't have the feeling parity is seeing this as a breaking change, meaning that they'll make such changes more often, meaning that we have to create more versions.

runtime/devnet/src/config/contracts.rs
```diff
@@ -67,7 +68,7 @@ impl pallet_contracts::Config for Runtime {
 	type MaxDebugBufferLen = ConstU32<{ 2 * 1024 * 1024 }>;
 	type MaxDelegateDependencies = ConstU32<32>;
 	type MaxStorageKeyLen = ConstU32<128>;
-	type MaxTransientStorageSize = ConstU32<{ 1024 * 1024 }>;
+	type MaxTransientStorageSize = ConstU32<{ 1 * 1024 * 1024 }>;
```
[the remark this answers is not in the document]
Daanvdplas: I remembered being added somewhere like this so added it for consistency but cant find it anymore... 

Changed it back.

runtime/devnet/src/config/contracts.rs
Daanvdplas: It had to be bigger than the other value so not much thought put towards it. 

I changed the contracts configuration back to default. As we might transition to only using revive I suggest we look into this when we add revive to the runtime and take a closer look at those configurations.

Daanvdplas: https://github.com/r0gue-io/pop-node/issues/415

runtime/common/src/lib.rs
```diff
@@ -26,8 +26,11 @@ mod block_times {
 pub use block_times::*;
 
 // Time is measured by number of blocks.
+/// Time is measured by number of blocks.
```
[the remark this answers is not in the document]
Daanvdplas: Thanks!

AlexD10S (approved):
Thank you for the comment clarifying the key changes. Great work!

pallets/api/src/fungibles/tests.rs
Daanvdplas: We have decided to tackle this later. The only solution would be to add a conversion for the specific error and query whether the asset is in destroy mode. With only adding the conversion it would change the error in all cases which it shouldn't. A discussion has been started to see how to move forward here: https://forum.polkadot.network/t/discussion-treating-changes-in-dispatchable-function-errors-as-breaking-changes/11222/4

runtime/testnet/src/config/contracts.rs
```diff
@@ -64,7 +53,7 @@ impl pallet_contracts::Config for Runtime {
 	// if a too-large contract is uploaded. We noticed that it poses
 	// less friction during development when the requirement here is
 	// just more lax.
-	type MaxCodeLen = ConstU32<{ 256 * 1024 }>;
+	type MaxCodeLen = ConstU32<{ 123 * 1024 }>;
```
peterwht, line 56: where does this come from? Normally these values are aligned / power-of-two bytes. `123` is not.

runtime/testnet/src/config/contracts.rs
```diff
@@ -46,7 +35,7 @@ impl pallet_contracts::Config for Runtime {
 	type ApiVersion = ();
 	// IMPORTANT: only runtime calls through the api are allowed.
 	type CallFilter = Nothing;
-	type CallStack = [pallet_contracts::Frame<Self>; 23];
+	type CallStack = [pallet_contracts::Frame<Self>; 5];
```
peterwht, line 38: where does this come from?

runtime/testnet/src/config/contracts.rs
```diff
@@ -36,7 +25,7 @@ impl<T: pallet_contracts::Config> Randomness<T::Hash, BlockNumberFor<T>> for Dum
 parameter_types! {
 	pub const DepositPerItem: Balance = deposit(1, 0);
 	pub const DepositPerByte: Balance = deposit(0, 1);
-	pub Schedule: pallet_contracts::Schedule<Runtime> = schedule::<Runtime>();
+	pub Schedule: pallet_contracts::Schedule<Runtime> = Default::default();
```
peterwht, line 28: isn't it possible that this could break existing contracts already deployed on testnet? If the limits are now lower, contracts used for development could break.

runtime/devnet/src/config/contracts.rs
```diff
@@ -64,7 +53,7 @@ impl pallet_contracts::Config for Runtime {
 	// if a too-large contract is uploaded. We noticed that it poses
 	// less friction during development when the requirement here is
 	// just more lax.
-	type MaxCodeLen = ConstU32<{ 256 * 1024 }>;
+	type MaxCodeLen = ConstU32<{ 123 * 1024 }>;
```
peterwht, line 56: per other comment, where does this come from?

runtime/devnet/src/config/contracts.rs
```diff
@@ -46,7 +35,7 @@ impl pallet_contracts::Config for Runtime {
 	type ApiVersion = ();
 	// IMPORTANT: only runtime calls through the api are allowed.
 	type CallFilter = Nothing;
-	type CallStack = [pallet_contracts::Frame<Self>; 23];
+	type CallStack = [pallet_contracts::Frame<Self>; 5];
```
peterwht, line 38: this could also potentially break already deployed contracts on testnet. Where does this change come from?

runtime/devnet/src/config/contracts.rs
Daanvdplas: I changed it back to what we had and multiply the `runtime_validator_memory` with 2, also been done in the substrate contracts node: https://github.com/paritytech/substrate-contracts-node/blob/f209befc88cb54ff50b7483c13d19e62213d0c60/runtime/src/contracts_config.rs#L31

My understanding is that `runtime_memory` is the maximum memory available during a contract call in the runtime, while `validator_runtime_memory` needs to be larger because validators must handle additional overhead like PoV data, events, and other block-related resources.

Cargo.toml
```diff
@@ -45,14 +46,14 @@ futures = "0.3.30"
 hex = "0.4.3"
 hex-literal = "0.4.1"
 impl-trait-for-tuples = "0.2.2"
-jsonrpsee = { version = "0.23.2", features = [ "server" ] }
-log = { version = "0.4.21", default-features = false }
+jsonrpsee = { version = "0.24.3" }
```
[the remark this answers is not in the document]
Daanvdplas: Thank you, removed it back!

runtime/testnet/src/config/contracts.rs
Daanvdplas: Changed it back

runtime/testnet/src/config/contracts.rs
Daanvdplas: Was the default but changed it back

runtime/testnet/src/config/contracts.rs
Daanvdplas: You are right, I asked for feedback to maintainers of the substrate contracts node and in the meantime changed it back.

peterwht (approved):

PR #413, opened 2024-12-18
feat(nfts): `AccountBalance` deposit

chungquantin:
## Problem 

This PR addresses the deposit issue with the `AccountBalance` storage type as outlined in #387. Currently, storage for each `AccountBalance` record created does not reserve the necessary deposit. Below are the cases where `AccountBalance` records are created or destroyed:

### Related Functions

- **`do_mint`**: Creates an `AccountBalance` record for `mint_to` when a new collection item is minted.
- **`do_transfer`**: Transfers a collection item from the owner to the `dest` account. If the `dest` account does not have enough balance (including does not exist), an `AccountBalance` record is created. If the sender's `AccountBalance` becomes zero after the transfer, their `AccountBalance` record is removed.
- **`do_burn`**: Removes the `AccountBalance` of the owner, if they only have one item and that item is burned.
- **`do_mint_pre_signed`**: Calls `do_mint`.

> Description for `do_mint_pre_signed`: This function allows minting a new item using a pre-signed message. The minting process is similar to the regular minting process, but it is performed by a pre-authorized account. The `mint_to` account receives the newly minted item. The minting process is configurable through the provided `mint_data`. The attributes, metadata, and price of the item are set according to the provided `mint_data`. The `with_details_and_config` closure is called to validate the provided `collection_details` and `collection_config` before minting the item.

- **`do_buy_item`**: Calls `do_transfer` to transfer an item from the seller to the buyer.
- **`do_claim_swap`**: Calls `do_transfer` twice—once to transfer from A to B, and then from B to A.

## Solution

Refactor the `AccountBalance` storage value type to store a tuple `(u32, AccountDepositOf<T>)`, representing:  
- `u32`: Tracks the total number of items in the `collection` owned by the `account`.  
- `AccountDepositOf<T>`: A tuple of `(AccountId, DepositBalance)`
   - `AccountId`: The depositor who reserves funds for creating the `AccountBalance` record.  
   - `DepositBalance`: The amount reserved from the `depositor` for creating the `AccountBalance` record.

### Reserve Logic

- **`do_mint`**:   
  - `Depositor` = Collection owner. 
  - If minting to a new `mint_to` account, the collection owner reserves the funds to create the `AccountBalance` record.  
  - If the account already exists in `AccountBalance`, the value of `AccountBalance` is simply incremented.

- **`do_transfer`**:  
  - `Depositor` = Provided `depositor` or the transferred collection item owner. 
  - Decrements the `AccountBalance` of the sender and increments that of the recipient.  
  - If transferring to a non-existing destination account (`dest`), the `depositor` or the collection item owner (if provided) reserves the funds to create the `AccountBalance` record.  
  - If the destination account already exists, the account itself is responsible for reserving the funds for its `AccountBalance` record.

- **`do_buy_item`**:  Call `do_transfer` with the `caller` as the buyer account. 

- **`do_claim_swap`**:  
  - Executes two `do_transfer` operations:  
    1. From A to B: Call `do_transfer` with the `depositor` as the `receive_item.owner`
    2. From B to A: Call `do_transfer` with the `caller` as the `send_item.owner`

### Unreserve Logic

- **`do_transfer`**:
  - Decrements the `AccountBalance` of the sender account. 
  -  If the sender's `AccountBalance` value reaches zero, the `AccountBalance` is removed, and the deposited funds are unreserved and returned to the `depositor`.
- **`do_burn`**:  
  - Decrements the `AccountBalance` of the item owner.  
  - If the owner's `AccountBalance` value reaches zero, the `AccountBalance` is removed, and the deposited funds are unreserved and returned to the `depositor`.


pallets/nfts/src/features/transfer.rs
```diff
@@ -87,16 +89,21 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-			balance.saturating_inc();
-		});
+		let deposit_amount =
+			match collection_config.is_setting_enabled(CollectionSetting::DepositRequired) {
+				true => T::BalanceDeposit::get(),
+				false => Zero::zero(),
+			};
+		// If the destination account exists, it reserves the `BalanceDeposit` for the new
+		// `AccountBalance`. Otherwise, the caller is responsible.
+		let deposit_account = match frame_system::Pallet::<T>::account_exists(&dest) {
```
chungquantin, line 102:
```suggestion
		let deposit_account = match T::Currency::free_balance(&dest) {
```

Curious if checking the balance makes more sense than checking the account existence?

pallets/nfts/src/tests.rs
```diff
@@ -4112,169 +4129,116 @@ fn clear_collection_metadata_works() {
...
+				(account(1), 0, 43),
+				(account(2), 0, 44),
+				(account(5), 0, 45)
+			]
+		);
 	});
 }
 
 #[test]
-fn clear_collection_approvals_works() {
```
chungquantin, line 4145: Moved to the bottom.

pallets/nfts/src/tests.rs
```diff
@@ -4112,169 +4129,116 @@ fn clear_collection_metadata_works() {
 }
 
 #[test]
-fn collection_item_works() {
```
chungquantin, line 4115: Moved to the bottom.

chungquantin:
[sc-1989]

pallets/nfts/src/tests.rs
```diff
@@ -4520,6 +4484,143 @@ fn cancel_collection_approval_works() {
 	});
 }
 
+#[test]
+fn clear_collection_approvals_works() {
```
chungquantin, line 4488: Not related to this PR.

pallets/nfts/src/features/atomic_swap.rs
```diff
@@ -209,10 +209,15 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-		})?;
 		Self::do_transfer(
+			caller,
+			send_collection_id,
+			send_item_id,
+			receive_item.owner.clone(),
+			|_, _| Ok(()),
+		)?;
+		Self::do_transfer(
+			receive_item.owner.clone(),
```
evilrobot-01, line 220: Shouldnt this be caller as well? The caller would be the account concluding the swap I assume, so they should pay for the deposit if the other party potentially does not yet have an account balance of the swapped collection and does not have sufficient funds to pay for the account balance deposit.

pallets/nfts/src/weights.rs
```diff
@@ -243,7 +243,9 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	/// Storage: `Nfts::Item` (r:1 w:1)
 	/// Proof: `Nfts::Item` (`max_values`: None, `max_size`: Some(861), added: 3336, mode: `MaxEncodedLen`)
 	/// Storage: `Nfts::AccountBalance` (r:2 w:2)
-	/// Proof: `Nfts::AccountBalance` (`max_values`: None, `max_size`: Some(72), added: 2547, mode: `MaxEncodedLen`)
+	/// Proof: `Nfts::AccountBalance` (`max_values`: None, `max_size`: Some(120), added: 2595, mode: `MaxEncodedLen`)
+	/// Storage: `System::Account` (r:1 w:1)
```
evilrobot-01, line 247: Just confirming that this is due to the deposit being charged from dest when they have an account with sufficient free balance for the collection balance deposit? It would be useful to have a review comment on these changes explaining why they are happening.

pallets/nfts/src/weights.rs
```diff
@@ -732,19 +734,21 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	/// Storage: `Nfts::ItemConfigOf` (r:1 w:0)
 	/// Proof: `Nfts::ItemConfigOf` (`max_values`: None, `max_size`: Some(48), added: 2523, mode: `MaxEncodedLen`)
 	/// Storage: `Nfts::AccountBalance` (r:2 w:2)
-	/// Proof: `Nfts::AccountBalance` (`max_values`: None, `max_size`: Some(72), added: 2547, mode: `MaxEncodedLen`)
+	/// Proof: `Nfts::AccountBalance` (`max_values`: None, `max_size`: Some(120), added: 2595, mode: `MaxEncodedLen`)
+	/// Storage: `System::Account` (r:1 w:1)
```
evilrobot-01, line 738: As above.

pallets/nfts/src/weights.rs
```diff
@@ -795,19 +799,21 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	/// Storage: `Nfts::ItemConfigOf` (r:2 w:0)
 	/// Proof: `Nfts::ItemConfigOf` (`max_values`: None, `max_size`: Some(48), added: 2523, mode: `MaxEncodedLen`)
 	/// Storage: `Nfts::AccountBalance` (r:2 w:2)
-	/// Proof: `Nfts::AccountBalance` (`max_values`: None, `max_size`: Some(72), added: 2547, mode: `MaxEncodedLen`)
+	/// Proof: `Nfts::AccountBalance` (`max_values`: None, `max_size`: Some(120), added: 2595, mode: `MaxEncodedLen`)
+	/// Storage: `System::Account` (r:1 w:1)
```
evilrobot-01, line 803: As above.

pallets/nfts/src/weights.rs
```diff
@@ -818,11 +824,11 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 	/// Storage: `Nfts::Collection` (r:1 w:1)
 	/// Proof: `Nfts::Collection` (`max_values`: None, `max_size`: Some(84), added: 2559, mode: `MaxEncodedLen`)
 	/// Storage: `Nfts::AccountBalance` (r:1 w:1)
-	/// Proof: `Nfts::AccountBalance` (`max_values`: None, `max_size`: Some(72), added: 2547, mode: `MaxEncodedLen`)
-	/// Storage: `Nfts::ItemConfigOf` (r:1 w:1)
-	/// Proof: `Nfts::ItemConfigOf` (`max_values`: None, `max_size`: Some(48), added: 2523, mode: `MaxEncodedLen`)
+	/// Proof: `Nfts::AccountBalance` (`max_values`: None, `max_size`: Some(120), added: 2595, mode: `MaxEncodedLen`)
 	/// Storage: `System::Account` (r:1 w:1)
 	/// Proof: `System::Account` (`max_values`: None, `max_size`: Some(128), added: 2603, mode: `MaxEncodedLen`)
```
evilrobot-01, line 829: FYI - Another useful source for checking the sizing of deposit bytes are these benchmark results.

pallets/nfts/src/weights.rs
```diff
@@ -1016,14 +1022,14 @@ impl WeightInfo for () {
...
 		//  Estimated: `4326`
-		// Minimum execution time: 46_000_000 picoseconds.
-		Weight::from_parts(47_000_000, 4326)
+		// Minimum execution time: 40_000_000 picoseconds.
+		Weight::from_parts(44_000_000, 4326)
 			.saturating_add(RocksDbWeight::get().reads(6_u64))
 			.saturating_add(RocksDbWeight::get().writes(8_u64))
 	}
-	/// Storage: `Nfts::Collection` (r:1 w:0)
+/// Storage: `Nfts::Collection` (r:1 w:0)
```
evilrobot-01, line 1032: Weird - makes it seem like something has been copy pasted?

runtime/devnet/src/config/assets.rs
```diff
@@ -29,6 +29,8 @@ parameter_types! {
 
 parameter_types! {
 	pub NftsPalletFeatures: PalletFeatures = PalletFeatures::all_enabled();
+	// Key = 68 bytes (4+16+32+16), Value = 44 bytes (4+32+8)
+	pub const NftsBalanceDeposit: Balance = deposit(1, 112);
```
evilrobot-01, line 33: The max size in the weights file shows account balance value as 120?

runtime/testnet/src/config/assets.rs
```diff
@@ -29,6 +29,8 @@ parameter_types! {
 
 parameter_types! {
 	pub NftsPalletFeatures: PalletFeatures = PalletFeatures::all_enabled();
+	// Key = 68 bytes (4+16+32+16), Value = 44 bytes (4+32+8)
+	pub const NftsBalanceDeposit: Balance = deposit(1, 112);
```
evilrobot-01, line 33: As above.

evilrobot-01 (changes requested):
Looking nice and clean! Have made some improvement suggestions, mostly minor, along with a few general questions.

Note that I did not review the tests yet owing to time. I took a quick look at https://app.codecov.io/gh/r0gue-io/pop-node/pull/413 and it seems as though there is some stuff added which is not covered by tests. They might be false positives, but please could you double check? I will check the tests on the next pass and then we should be good to go from my perspective.

al3mart:
I'll dedicate more time later to do a proper review, but reserve and unreserve logic seems solid.

Before I was more of the opinion of centralizing the responsibilities for the existence of the assets into the minter / owner of the collection.

But I gotta say that I'm pretty convinced with the current approach. For instance moving `DepositBalance` on transfers could be pretty beneficial for the asset creators and potentially being able to absorb a cost up front that is later shared with their users might be a pretty good incentive.

pallets/nfts/src/impl_nonfungibles.rs
```diff
@@ -411,7 +411,9 @@ impl<T: Config<I>, I: 'static> Transfer<T::AccountId> for Pallet<T, I> {
 		item: &Self::ItemId,
 		destination: &T::AccountId,
 	) -> DispatchResult {
-		Self::do_transfer(*collection, *item, destination.clone(), |_, _| Ok(()))
+		Self::do_transfer(destination.clone(), *collection, *item, destination.clone(), |_, _| {
```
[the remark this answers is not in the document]
chungquantin: By refactoring helper methods, this is resolved. Thanks for the suggestion!

pallets/nfts/src/features/create_delete_item.rs
```diff
@@ -68,20 +68,20 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-			let deposit_account = match maybe_depositor {
-				None => collection_details.owner.clone(),
-				Some(depositor) => depositor,
-			};
+			let deposit_account = maybe_depositor.unwrap_or(collection_details.owner.clone());
+
+			Self::increment_account_balance(
+				collection,
+				&mint_to,
+				deposit_account.clone(),
```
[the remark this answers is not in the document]
chungquantin: Resolved by refactoring `increment_account_balance`

pallets/nfts/src/lib.rs
```diff
@@ -1086,7 +1091,7 @@ pub mod pallet {
 			let origin = ensure_signed(origin)?;
 			let dest = T::Lookup::lookup(dest)?;
 
-			Self::do_transfer(collection, item, dest, |_, details| {
+			Self::do_transfer(origin.clone(), collection, item, dest, |_, details| {
```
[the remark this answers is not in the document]
chungquantin: Resolved by refactoring helper functions.

pallets/nfts/src/features/approvals.rs
```diff
@@ -216,7 +216,10 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			Self::is_pallet_feature_enabled(PalletFeature::Approvals),
 			Error::<T, I>::MethodDisabled
 		);
-		ensure!(AccountBalance::<T, I>::get(collection, &owner) > 0, Error::<T, I>::NoItemOwned);
+		ensure!(
```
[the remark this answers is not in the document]
chungquantin: Resolved by refactoring to both check if key exist and the value is more than zero.

pallets/nfts/src/weights.rs
chungquantin: Oh yes, I only run benchmarking on related methods and replace them with the main `weight.rs` file.

pallets/nfts/src/benchmarking.rs
```diff
@@ -757,7 +757,7 @@ benchmarks_instance_pallet! {
 		let duration = T::MaxDeadlineDuration::get();
 		let target: T::AccountId = account("target", 0, SEED);
 		let target_lookup = T::Lookup::unlookup(target.clone());
-		T::Currency::make_free_balance_be(&target, T::Currency::minimum_balance());
+		T::Currency::make_free_balance_be(&target, DepositBalanceOf::<T, I>::max_value());
```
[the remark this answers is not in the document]
chungquantin: Resolved

pallets/nfts/src/tests.rs
```diff
@@ -4894,66 +4997,65 @@ fn decrement_account_balance_works() {
...
 fn ensure_account_balance_bytes() {
-	let key = Blake2_128Concat::max_len::<<Test as crate::Config>::CollectionId>()
-		.saturating_add(Blake2_128Concat::max_len::<AccountId>());
-	let value = u32::max_encoded_len()
-		.saturating_add(AccountId::max_encoded_len())
-		.saturating_add(DepositBalanceOf::<Test>::max_encoded_len());
-	assert_eq!(key + value, 112);
+	assert!(AccountBalance::<Test>::storage_info()
+		.first()
+		.filter(|i| i.max_size.unwrap_or_default() == 112)
```
chungquantin, line 5051: Benchmarking returns the max size as `120`

runtime/testnet/src/config/assets.rs
chungquantin: Even I changed the test to return the direct max size value from the storage info ([here](https://github.com/r0gue-io/pop-node/pull/413#discussion_r1894768254)), it still return `112` which is different from the benchmarking. Wonder if I am missing something, are there any extra overhead when running the actual code?

pallets/nfts/src/weights.rs
chungquantin: Hmm I checked the `ItemMetadata`'s max size and the configured bytes of `MetadataDeposit` in both testnet and devnet and it is different. Am I missing anything?

runtime/testnet/src/config/assets.rs
chungquantin: The tests was incorrect because it used the configuration of mock runtime in the pallet not devnet / testnet / mainnet. Resolved by moving the test to its specific runtimes and updated the comments.

runtime/devnet/src/config/assets.rs
chungquantin: The tests was incorrect because it used the configuration of mock runtime in the pallet not devnet / testnet / mainnet. Resolved by moving the test to its specific runtimes and updated the comments.

pallets/nfts/src/weights.rs
chungquantin: Thank you!

pallets/nfts/src/weights.rs
chungquantin, line 247: - `r: 1`: Read the balance of the `dest` account with `T::Currency::can_reserve` to identify the depositor of the call. 
- `w: 1`: Reserve from the `depositor` account. Either the `caller` or the `dest` account.

pallets/nfts/src/weights.rs
chungquantin: The `buy_item` implementation calls `do_transfer` to transfer from the buyer to the dest account. Hence, Takes reads and writes similar to the `transfer()` dispatchable. 

- `r: 1`: Read the balance of the `dest` account with `T::Currency::can_reserve` to identify the depositor of the call. 
- `w: 1`: Reserve from the `depositor` account. Either the `caller` or the `dest` account.

pallets/nfts/src/weights.rs
chungquantin: Question: Interesting to see that there is only 1 read and 1 write to the `System::Account`. From my understanding, the `do_transfer` is called twice from A to B and then B to A so it must be 2 reads and 2 writes.

pallets/nfts/src/weights.rs
chungquantin: - `r: 1`: Read the balance of the `dest` account with `T::Currency::can_reserve` to identify the depositor of the call. 
- `w: 1`: Reserve from the `depositor` account. Either the `caller` or the `dest` account.

pallets/nfts/src/benchmarking.rs
```diff
@@ -274,6 +274,7 @@ benchmarks_instance_pallet! {
 	mint {
 		let (collection, caller, caller_lookup) = create_collection::<T, I>();
 		let item = T::Helper::item(0);
+		T::Currency::make_free_balance_be(&caller, T::Currency::minimum_balance() + T::CollectionDeposit::get() + T::BalanceDeposit::get());
```
chungquantin, line 277: Note: This overrides the balance of `caller` in `create_collection`. The purpose is to make this more specific to what are reserved in the function.

pallets/nfts/src/weights.rs
```diff
@@ -2,13 +2,13 @@
...
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 40.0.0
-//! DATE: 2024-12-12, STEPS: `50`, REPEAT: `20`, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! DATE: 2024-12-22, STEPS: `50`, REPEAT: `20`, LOW RANGE: `[]`, HIGH RANGE: `[]`
 //! WORST CASE MAP SIZE: `1000000`
 //! HOSTNAME: `R0GUE`, CPU: `<UNKNOWN>`
 //! WASM-EXECUTION: `Compiled`, CHAIN: `Some("dev")`, DB CACHE: `1024`
 
 // Executed Command:
-// ./target/release/pop-node
+// ./target/production/pop-node
```
chungquantin, line 11: DQ: I build with production mode instead of release, does it matter?

evilrobot-01: No, we will need to benchmark using the production profile eventually. The only consideration here is that a production build takes longer, so release is fine for ongoing dev.

pallets/nfts/src/common_functions.rs
```diff
@@ -92,6 +92,55 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 		Self::deposit_event(Event::NextCollectionIdIncremented { next_id });
 	}
 
+	/// Increment the number of items in the `collection` owned by the `owner`. If no entry exists
+	/// for the `owner` in `AccountBalance`, create a new record and reserve `deposit_amount` from
+	/// the `deposit_account`.
+	pub(crate) fn increment_account_balance(
+		collection: T::CollectionId,
+		owner: &T::AccountId,
+		(deposit_account, deposit_amount): (
```
chungquantin, line 101: Note: Not possible to reuse the tuple type `AccountDepositOf` with `AccountId` passed by reference.

pallets/nfts/src/features/create_delete_item.rs
```diff
@@ -68,20 +68,20 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-				None => collection_details.owner.clone(),
-				Some(depositor) => depositor,
-			};
+			let deposit_account =
+				maybe_depositor.unwrap_or_else(|| collection_details.owner.clone());
+
+			Self::increment_account_balance(
+				collection,
+				&mint_to,
+				(&deposit_account, deposit_amount),
```
Daanvdplas, line 83: deposit_amount is T::ItemDeposit here? Why are we not collecting the deposit for AccountBalance as well?

pallets/nfts/src/tests.rs
```diff
@@ -273,35 +274,35 @@ fn lifecycle_should_work() {
 			account(10),
 			default_item_config()
 		));
-		assert_eq!(Balances::reserved_balance(&account(1)), 6);
+		assert_eq!(Balances::reserved_balance(&account(1)), 6 + 1); // 1 - balance deposit
```
Daanvdplas, line 277: Can we use the config `T::BalanceDeposit`, see increment_account_balance_works comment.

And everywhere else in the tests. Then we also don't need the comment as it explains itself.

pallets/nfts/src/tests.rs
```diff
@@ -761,7 +767,8 @@ fn transfer_owner_should_work() {
 		assert_ok!(Nfts::transfer(RuntimeOrigin::signed(account(1)), 0, 42, account(2)));
 		// reserved_balance of accounts 1 & 2 should be unchanged:
 		assert_eq!(Balances::reserved_balance(&account(1)), 1);
-		assert_eq!(Balances::reserved_balance(&account(2)), 0);
+		// transfer the collection item reserves account 2's balance.
```
Daanvdplas, line 770: ?

pallets/nfts/src/tests.rs
```diff
@@ -231,7 +230,8 @@ fn basic_minting_should_work() {
 		));
 		assert_eq!(collections(), vec![(account(1), 0)]);
 		assert_ok!(Nfts::mint(RuntimeOrigin::signed(account(1)), 0, 42, account(1), None));
-		assert_eq!(AccountBalance::<Test>::get(0, &account(1)), 1);
+		// Minting skips deposit as `DepositRequired` is disabled.
```
Daanvdplas, line 233: I don't understand this?

Daanvdplas (commented):
Looking good. Didn't review the tests because there were some uncertainties regarding the deposit that was not taken.

pallets/nfts/src/tests.rs
```diff
@@ -273,35 +274,35 @@ fn lifecycle_should_work() {
...
 			RuntimeOrigin::signed(account(1)),
 			0,
 			69,
 			account(20),
 			default_item_config()
 		));
-		assert_eq!(AccountBalance::<Test>::get(0, account(20)), 1);
-		assert_eq!(Balances::reserved_balance(&account(1)), 7);
+		assert_eq!(AccountBalance::get(0, account(20)), Some((1, (account(1), 1))));
+		assert_eq!(Balances::reserved_balance(&account(1)), 7 + 2); // 2 - balance deposit
```
Daanvdplas, line 286: Why is is `+ 2` here? I don't understand the comment here as well

pallets/nfts/src/tests.rs
```diff
@@ -953,17 +960,17 @@ fn set_item_metadata_should_work() {
 
 		// Successfully add metadata and take deposit
 		assert_ok!(Nfts::set_metadata(RuntimeOrigin::signed(account(1)), 0, 42, bvec![0u8; 20]));
-		assert_eq!(Balances::free_balance(&account(1)), 8);
+		assert_eq!(Balances::free_balance(&account(1)), 7); // free_balance - metadata deposit - item deposit - balance deposit
```
Daanvdplas, line 963: To minimise the changes in the PR but illustrate what is happening:
```suggestion
		assert_eq!(Balances::free_balance(&account(1)), 8 - balance_deposit);
```

pallets/nfts/src/tests.rs
```diff
@@ -761,7 +767,8 @@ fn transfer_owner_should_work() {
 		assert_ok!(Nfts::transfer(RuntimeOrigin::signed(account(1)), 0, 42, account(2)));
 		// reserved_balance of accounts 1 & 2 should be unchanged:
 		assert_eq!(Balances::reserved_balance(&account(1)), 1);
-		assert_eq!(Balances::reserved_balance(&account(2)), 0);
+		// transfer the collection item reserves account 2's balance.
+		assert_eq!(Balances::reserved_balance(&account(2)), 1); // 1 - balance deposit
```
Daanvdplas, line 771: Use config item

pallets/nfts/src/tests.rs
```diff
@@ -1048,7 +1055,7 @@ fn set_collection_owner_attributes_should_work() {
 				(Some(0), AttributeNamespace::CollectionOwner, bvec![1], bvec![0]),
 			]
 		);
-		assert_eq!(Balances::reserved_balance(account(1)), 10);
+		assert_eq!(Balances::reserved_balance(account(1)), 10 + 1); // 10 - attribute deposit, 1 - balance deposit
```
Daanvdplas, line 1058: We don't need the comments here, should be solved in a PR to sdk where we fix the test variables.

pallets/nfts/src/tests.rs
```diff
@@ -953,17 +960,17 @@ fn set_item_metadata_should_work() {
...
 		// Force origin works, too.
 		assert_ok!(Nfts::set_metadata(RuntimeOrigin::root(), 0, 42, bvec![0u8; 18]));
 
 		// Update deposit
 		assert_ok!(Nfts::set_metadata(RuntimeOrigin::signed(account(1)), 0, 42, bvec![0u8; 15]));
-		assert_eq!(Balances::free_balance(&account(1)), 13);
+		assert_eq!(Balances::free_balance(&account(1)), 12); // free_balance - metadata deposit - item deposit - balance deposit
 		assert_ok!(Nfts::set_metadata(RuntimeOrigin::signed(account(1)), 0, 42, bvec![0u8; 25]));
-		assert_eq!(Balances::free_balance(&account(1)), 3);
+		assert_eq!(Balances::free_balance(&account(1)), 2); // free_balance - metadata deposit - item deposit - balance deposit
```
Daanvdplas, line 973:
```suggestion
		assert_eq!(Balances::free_balance(&account(1)), 2 - balance_deposit);
```

pallets/nfts/src/tests.rs
```diff
@@ -953,17 +960,17 @@ fn set_item_metadata_should_work() {
...
+		assert_eq!(Balances::free_balance(&account(1)), 7); // free_balance - metadata deposit - item deposit - balance deposit
 		assert!(ItemMetadataOf::<Test>::contains_key(0, 42));
 
 		// Force origin works, too.
 		assert_ok!(Nfts::set_metadata(RuntimeOrigin::root(), 0, 42, bvec![0u8; 18]));
 
 		// Update deposit
 		assert_ok!(Nfts::set_metadata(RuntimeOrigin::signed(account(1)), 0, 42, bvec![0u8; 15]));
-		assert_eq!(Balances::free_balance(&account(1)), 13);
+		assert_eq!(Balances::free_balance(&account(1)), 12); // free_balance - metadata deposit - item deposit - balance deposit
```
Daanvdplas, line 971:
```suggestion
		assert_eq!(Balances::free_balance(&account(1)), 12 - balance deposit);
```

pallets/nfts/src/tests.rs
```diff
@@ -4729,11 +4928,121 @@ fn check_approval_with_deadline_works() {
...
 
 #[test]
-fn ensure_collection_approval_bytes() {
-	let key = Blake2_128Concat::max_len::<<Test as crate::Config>::CollectionId>() +
-		Blake2_128Concat::max_len::<AccountId>() +
-		Blake2_128Concat::max_len::<AccountId>();
-	let value = Option::<BlockNumberFor<Test>>::max_encoded_len()
-		.saturating_add(BalanceOf::<Test>::max_encoded_len());
-	assert_eq!(key + value, 133);
+fn collection_item_works() {
```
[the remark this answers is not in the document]
chungquantin: This is not clean because you have to move shared functions like `account` and type aliases used in the `test.rs` around.

pallets/nfts/src/tests.rs
```diff
@@ -4729,11 +4928,121 @@ fn check_approval_with_deadline_works() {
...
+			assert_eq!(
+				AccountBalance::get(collection_id, &owner),
+				Some((i + 1, (deposit_account.clone(), 1)))
+			);
+		});
+	});
+}
+
+#[test]
+fn decrement_account_balance_works() {
```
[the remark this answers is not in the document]
chungquantin: This is not clean because you have to move shared functions like `account` and type aliases used in the `test.rs` around.

pallets/nfts/src/impl_nonfungibles.rs
```diff
@@ -411,7 +411,7 @@ impl<T: Config<I>, I: 'static> Transfer<T::AccountId> for Pallet<T, I> {
 		item: &Self::ItemId,
 		destination: &T::AccountId,
 	) -> DispatchResult {
-		Self::do_transfer(*collection, *item, destination.clone(), |_, _| Ok(()))
+		Self::do_transfer(destination, *collection, *item, destination.clone(), |_, _| Ok(()))
```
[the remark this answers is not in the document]
chungquantin: Yes that function is used by other pallets. I put `destination` because we don't have any `origin` provided and we can't change the trait interface.

pallets/nfts/src/tests.rs
chungquantin: Resolved by changing the comment

pallets/nfts/src/tests.rs
```diff
@@ -759,9 +775,10 @@ fn transfer_owner_should_work() {
 		assert_eq!(Balances::reserved_balance(&account(3)), 44);
 
 		assert_ok!(Nfts::transfer(RuntimeOrigin::signed(account(1)), 0, 42, account(2)));
-		// reserved_balance of accounts 1 & 2 should be unchanged:
-		assert_eq!(Balances::reserved_balance(&account(1)), 1);
-		assert_eq!(Balances::reserved_balance(&account(2)), 0);
+		// Reserve `balance_deposit` from account 2 and reserved balance of account 1 remains the
+		// same.
+		assert_eq!(Balances::reserved_balance(&account(1)), balance_deposit);
```
chungquantin, line 780:
```suggestion
		assert_eq!(Balances::reserved_balance(&account(1)),1);
```

pallets/nfts/src/impl_nonfungibles.rs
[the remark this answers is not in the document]
Daanvdplas: This is a concern but like you say, we have no other origin to put in here and in the end the destination account should be the one paying for it. Perhaps in the future we could make the change to the trait function to allow for the `caller` parameter but I'd say this is something to reassess later.

PS: don't resolve so other reviewers see this conversation too

pallets/nfts/src/tests.rs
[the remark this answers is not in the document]
Daanvdplas: Don't see a problem with that but alright. Lets leave this open and see what other reviewers think.

pallets/nfts/src/tests.rs
```diff
@@ -759,9 +775,10 @@ fn transfer_owner_should_work() {
 		assert_eq!(Balances::reserved_balance(&account(3)), 44);
 
 		assert_ok!(Nfts::transfer(RuntimeOrigin::signed(account(1)), 0, 42, account(2)));
-		// reserved_balance of accounts 1 & 2 should be unchanged:
+		// Reserve `balance_deposit` from account 2 and reserved balance of account 1 remains the
+		// same.
```
Daanvdplas, line 779: Didn't look into why but the comment indicated that checks should be the case. You changed it to what was happening. The comment didn't provide any value but it is not the PR to improve unfortunately.
```suggestion
		// The reserved balance of account 1 should remain the same. For account 2 the `balance_deposit`  is 
                // reserved because it is the new item owner. 
```

pallets/nfts/src/tests.rs
```diff
@@ -1177,6 +1198,8 @@ fn set_collection_system_attributes_should_work() {
 #[test]
 fn set_item_owner_attributes_should_work() {
 	new_test_ext().execute_with(|| {
+		let balance_deposit = balance_deposit();
```
Daanvdplas, line 1201: No need for a variable if it is only used once.

pallets/nfts/src/tests.rs
```diff
@@ -1536,6 +1559,8 @@ fn validate_deposit_required_setting() {
 #[test]
 fn set_attribute_should_respect_lock() {
 	new_test_ext().execute_with(|| {
+		let balance_deposit = balance_deposit();
```
Daanvdplas, line 1562: same

pallets/nfts/src/tests.rs
```diff
@@ -1828,6 +1857,7 @@ fn burn_works() {
 			Some(account(4)),
 		));
 
+		// Throws error `Error::UnknownItem` if burning an unknown collection item.
```
Daanvdplas, line 1860: Can remove as nothing has changed here so this only creates unnecessary noise

pallets/nfts/src/tests.rs
```diff
@@ -1847,16 +1877,32 @@ fn burn_works() {
...
+		// Throws error `Error::NoItemOwned` if the account does not have an entry in
+		// `AccountBalance`.
+		let account_balance = AccountBalance::take(0, account(5));
+		assert_noop!(
+			Nfts::burn(RuntimeOrigin::signed(account(5)), 0, 42),
+			Error::<Test>::NoItemOwned
+		);
+		AccountBalance::set(0, account(5), account_balance);
+
+		// Throws error `Error::NoPermission`.
```
Daanvdplas, line 1890: Can remove as nothing has changed here so this only creates unnecessary noise

pallets/nfts/src/tests.rs
```diff
@@ -1847,16 +1877,32 @@ fn burn_works() {
...
+		assert_eq!(Balances::reserved_balance(account(1)), 2 + balance_deposit);
 		assert_ok!(Nfts::burn(RuntimeOrigin::signed(account(5)), 0, 42));
+		// Burn an item unreserving the item deposit while retaining the reservation for `balance
+		// deposit`.
+		assert_eq!(Balances::reserved_balance(account(1)), 1 + balance_deposit);
 		assert_ok!(Nfts::burn(RuntimeOrigin::signed(account(5)), 0, 69));
-		assert!(!AccountBalance::<Test>::contains_key(0, &account(5)));
+		// Burn an item unreserving the balance deposit when the value of `AccountBalance` reaches
+		// zero. Remove the `AccountBalance` record.
+		assert_eq!(Balances::reserved_balance(account(1)), 0);
```
Daanvdplas, line 1904: Same line two lines under

pallets/nfts/src/tests.rs
```diff
@@ -1847,16 +1877,32 @@ fn burn_works() {
...
-		assert_eq!(Balances::reserved_balance(account(1)), 2);
 
+		// Throws error `Error::NoItemOwned` if the account does not have an entry in
+		// `AccountBalance`.
+		let account_balance = AccountBalance::take(0, account(5));
+		assert_noop!(
+			Nfts::burn(RuntimeOrigin::signed(account(5)), 0, 42),
+			Error::<Test>::NoItemOwned
+		);
+		AccountBalance::set(0, account(5), account_balance);
```
Daanvdplas, line 1888: It is good that you are testing the error case of logic that we have added yet this error can never happen as `NoPermission` is protecting against it. If you want to add it I would prefer to see a separate test. This also clearly demonstrates to the reviewer that the added logic is tested.

pallets/nfts/src/tests.rs
```diff
@@ -3614,7 +3666,7 @@ fn pre_signed_mints_should_work() {
 		assert_eq!(deposit.amount, 3);
 
 		assert_eq!(Balances::free_balance(&user_0), 100 - 2 + 10); // 2 - collection deposit, 10 - mint price
-		assert_eq!(Balances::free_balance(&user_2), 100 - 1 - 3 - 6 - 10); // 1 - item deposit, 3 - metadata, 6 - attributes, 10 - mint price
+		assert_eq!(Balances::free_balance(&user_2), 100 - balance_deposit - 1 - 3 - 6 - 10); // 1 - balance deposit, 1 - item deposit, 3 - metadata, 6 - attributes, 10 - mint price
```
Daanvdplas, line 3669: Not necessary in my opinion as the variable name explains enough.
```suggestion
		assert_eq!(Balances::free_balance(&user_2), 100 - balance_deposit - 1 - 3 - 6 - 10); // 1 - item deposit, 3 - metadata, 6 - attributes, 10 - mint price
```

pallets/nfts/src/tests.rs
```diff
@@ -3789,7 +3841,7 @@ fn pre_signed_attributes_should_work() {
 		assert_eq!(deposit.account, Some(user_2.clone()));
 		assert_eq!(deposit.amount, 3);
 
-		assert_eq!(Balances::free_balance(&user_1), 100 - 2 - 1); // 2 - collection deposit, 1 - item deposit
+		assert_eq!(Balances::free_balance(&user_1), 100 - 2 - 1 - 1); // 2 - collection deposit, 1 - item deposit, 1 - balance deposit
```
Daanvdplas, line 3844:
```suggestion
		assert_eq!(Balances::free_balance(&user_1), 100 - 2 - 1 - balance_deposit()); // 2 - collection deposit, 1 - item deposit
```

pallets/nfts/src/tests.rs
```diff
@@ -3587,6 +3635,10 @@ fn pre_signed_mints_should_work() {
 			signature.clone(),
 			user_1.clone(),
 		));
+		assert_eq!(
```
Daanvdplas, line 3638: Rather see this in a separate test.

pallets/nfts/src/tests.rs
Daanvdplas: This creates a lot of noise in the PR. Can we please redo this

pallets/nfts/src/tests.rs
```diff
@@ -4112,168 +4164,212 @@ fn clear_collection_metadata_works() {
 }
 
 #[test]
-fn collection_item_works() {
+fn mint_requires_deposit_works() {
```
Daanvdplas, line 4167: Yes!! This test makes reviewing the code so much better!! Because we know exactly what is tested and why (by looking at the changes in the main PR), it is much clearer and improves the quality of the review tremendously. 

Like we discussed in the call two weeks ago, we would like to see the tests that already existed with the least amount of changes possible. Then, all the added logic to dispatchables is tested in a separate test function.

pallets/nfts/src/tests.rs
Daanvdplas: Please revert for this PR as it creates noise and it is not important for this PR

pallets/nfts/src/tests.rs
Daanvdplas: I take my comment back, this should be reverted.

Daanvdplas (commented):
I think the PR is as good as finished but due to how the changes have been made the PR is simply too complex and time consuming to review. The pallet is super complicated and in order to review the checks you add to those big tests we have to understand the test which takes too much time. 

This is also the case for the main PR. It would make our lives a lot easier as the tests for the added logic is separated from the already existing tests. I guess this means that you first have to make changes in the main PR and rebase this PR.

pallets/nfts/src/tests.rs
```diff
@@ -214,6 +213,10 @@ fn cancel_collection_approval(
 	}
 }
 
+fn balance_deposit() -> DepositBalanceOf<Test> {
```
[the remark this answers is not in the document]
chungquantin: Sorry I don't understand this. The method uses parameter provided for the `Test` runtime, I don't think user might want to use that configuration.

pallets/nfts/src/tests.rs
chungquantin: I don't understand

pallets/nfts/src/tests.rs
```diff
@@ -273,35 +280,35 @@ fn lifecycle_should_work() {
 			account(10),
 			default_item_config()
 		));
-		assert_eq!(Balances::reserved_balance(&account(1)), 6);
+		assert_eq!(Balances::reserved_balance(&account(1)), 6 + balance_deposit);
```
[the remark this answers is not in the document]
chungquantin: Added below.

pallets/nfts/src/tests.rs
```diff
@@ -734,9 +748,10 @@ fn transfer_owner_should_work() {
 		assert_eq!(Balances::reserved_balance(&account(3)), 44);
 
 		assert_ok!(Nfts::transfer(RuntimeOrigin::signed(account(1)), 0, 42, account(2)));
-		// reserved_balance of accounts 1 & 2 should be unchanged:
+		// The reserved balance of account 1 should remain the same. For account 2 the
+		// `balance_deposit`  is reserved because it is the new item owner.
```
chungquantin, line 752:
```suggestion
		// `balance_deposit` is reserved because it is the new item owner.
```

pallets/nfts/src/tests.rs
```diff
@@ -5028,3 +5068,336 @@ fn destroy_with_collection_approvals_returns_error() {
 		assert_eq!(ItemConfigOf::<Test>::iter_prefix(collection_id).count() as u32, collection_id);
 	});
 }
+
+#[test]
+fn mint_requires_deposit_works() {
```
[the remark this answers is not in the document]
chungquantin: Resolved by adding the `burn_removes_deposit_works` below.

pallets/nfts/src/tests.rs
```diff
@@ -5028,3 +5068,336 @@ fn destroy_with_collection_approvals_returns_error() {
...
+		// Throws error `BalancesError::InsufficientBalance`.
+		assert_noop!(
+			Nfts::mint(
+				RuntimeOrigin::signed(collection_owner.clone()),
+				collection_id,
+				item_id,
+				item_owner.clone(),
+				None
+			),
+			BalancesError::<Test>::InsufficientBalance
```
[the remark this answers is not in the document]
chungquantin: Removed.

pallets/nfts/src/features/transfer.rs
```diff
@@ -87,16 +89,22 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-			balance.saturating_inc();
-		});
+		let deposit_amount =
+			match collection_config.is_setting_enabled(CollectionSetting::DepositRequired) {
+				true => T::BalanceDeposit::get(),
+				false => Zero::zero(),
+			};
+		// The destination account covers the `BalanceDeposit` if it has sufficient balance.
+		// Otherwise, the caller is accountable for it.
+		let deposit_account = match T::Currency::can_reserve(&dest, T::BalanceDeposit::get()) {
```
peterwht, line 102: `can_reserve` is not using the already calculated`deposit_amount`.

pallets/nfts/src/features/transfer.rs
```diff
@@ -87,16 +89,22 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
...
-		AccountBalance::<T, I>::mutate(collection, &dest, |balance| {
-			balance.saturating_inc();
-		});
+		let deposit_amount =
+			match collection_config.is_setting_enabled(CollectionSetting::DepositRequired) {
+				true => T::BalanceDeposit::get(),
+				false => Zero::zero(),
+			};
+		// The destination account covers the `BalanceDeposit` if it has sufficient balance.
+		// Otherwise, the caller is accountable for it.
```
peterwht, line 101: I don't think the caller should become accountable for the deposit. The `caller` can be either the current owner or the delegate. Let's explore the two cases:

1) I am the owner and am transferring
I am transferring the NFT and would like to be "free" of it. However, the account I am transferring to can't cover the deposit balance.. I now have my deposit stuck in an NFT that someone else owns. I will not be able to get this deposit back.
2) I am a delegate to transfer the NFT. For example, an NFT marketplace
I am brokering the transaction to transfer the NFT, however, the person receiving it can't cover the balance. Hence, I as a marketplace also now have a deposit stuck in an NFT that I don't have control over. 

In either case, it seems unfair to the caller to become accountable for a deposit when it is impossible (unlikely) that they can recover the deposit -- because they don't have control of the NFT.

**Secondly, this actually introduces a security flaw.**

Transferring an NFT does NOT require approval from the destination account. But we are now reserving a deposit for an unapproved NFT into some arbitrary account. A malicious actor could cause accounts to have unknown deposits. Maybe causing them to dip below a balance threshold.

IMO, the person transferring the NFT should have to bear the deposit (transferred to the new owner). Recommendation:
- Repatriate the deposit from the current depositor's account: https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/nfts/src/features/transfer.rs#L143

peterwht (changes requested):
Overall pretty clean! Was nice diving into the NFT pallets some. However, there is a security flaw introduced with `transfer`. Please review comments.

Also, can we please rename the PR title to `feat(nfts): AccountBalance deposit`. We are introducing new functionality, so it should be a feature.

pallets/nfts/src/features/transfer.rs
peterwht: A concrete flaw example: an account creates a large collection of NFTs. They pay the deposits upfront. They then transfer these NFTs to one or many other accounts. They free up their deposits, but the receiving accounts now have large chunks of funds locked with these NFT deposits.

al3mart: Great points!

Making the transfer fail if the beneficiary doesn't have enough funds to cover the deposit might hurt the UX of some scenarios without other mechanisms to help it :thinking:

Seems like it might be difficult to move away from approvals if we really want to stick with this

pallets/nfts/src/lib.rs
```diff
@@ -178,12 +178,6 @@ pub mod pallet {
 		#[pallet::constant]
 		type CollectionDeposit: Get<DepositBalanceOf<Self, I>>;
 
-		/// The basic amount of funds that must be reserved for a collection approval.
-		// Key: `sizeof((CollectionId, AccountId, AccountId))` bytes.
-		// Value: `sizeof((Option<BlockNumber>, Balance))` bytes.
-		#[pallet::constant]
-		type CollectionApprovalDeposit: Get<DepositBalanceOf<Self, I>>;
```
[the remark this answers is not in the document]
chungquantin: This is a storage item we added in the main pallet-nfts optimization PR. Order changed to keep it consistent and easy to keep track of changes.

pallets/nfts/src/features/transfer.rs
chungquantin: Implementation of the `can_reserve` checks the `account.free` which is deducted when you call `reserve()`: https://paritytech.github.io/polkadot-sdk/master/src/pallet_balances/impl_currency.rs.html#522

Not very clear to me when you say: "`can_reserve` is not using the already calculated `deposit_amount.`". Am I missing anything?

Daanvdplas: `let deposit_account = match T::Currency::can_reserve(&dest, deposit_amount)`

pallets/nfts/src/features/transfer.rs
chungquantin: Resolved in the PR: https://github.com/r0gue-io/pop-node/pull/418

pallets/nfts/src/features/transfer.rs
chungquantin: Outdated.

Daanvdplas (approved):
Great work! LGTM!

pallets/nfts/src/impl_nonfungibles.rs
```diff
@@ -411,7 +411,11 @@ impl<T: Config<I>, I: 'static> Transfer<T::AccountId> for Pallet<T, I> {
 		item: &Self::ItemId,
 		destination: &T::AccountId,
 	) -> DispatchResult {
-		Self::do_transfer(*collection, *item, destination.clone(), |_, _| Ok(()))
+		// The item's owner pays for the deposit of `AccountBalance` if the `dest` holds no items
+		// in `collection`. A malicious actor could have a deposit reserved from `dest` without
+		// them knowing about the transfer. The deposit amount can be accounted for in the off chain
+		// price of the NFT.
```
peterwht, line 417:
```suggestion
		// The item's owner pays for the deposit of `AccountBalance` if the `dest` holds no items
		// in `collection`. If `dest` paid the deposit, a malicious actor could transfer NFTs to reserve involuntary deposits for `dest` . The deposit amount can be accounted for in the off chain
		// price of the NFT.
```

peterwht (commented):
Nice! Wasn't able to find much to change ;) Left two tiny comments. After those, I am good to approve.

pallets/nfts/src/features/transfer.rs
```diff
@@ -52,6 +54,7 @@ impl<T: Config<I>, I: 'static> Pallet<T, I> {
 			&CollectionDetailsFor<T, I>,
 			&mut ItemDetailsFor<T, I>,
 		) -> DispatchResult,
+		depositor: Option<&T::AccountId>,
```
peterwht, line 57: Sorry, I should have been more clear!! Could we move this inbetween `dest` and `with_details`. `with_details` is also an optional, and since it's a closure it should actually be the very last (makes reading difficult when there are parameters after a closure). After this, it's good to merge.

chungquantin: Resolved.

peterwht (approved):

PR #414, opened 2024-12-18
chore: comment out ismp

Daanvdplas:
In order to uplift pop to stable 2412 the ismp related code will be commented out. This is because the ismp dependencies are currently at 2407 and there is no eta when they will do those upgrades. In the meantime we can proceed with the upgrades and relevant work depending on this.

As for the messaging feature that is in the pipeline, I created a branch from main: `feat/messaging`. This branch is on v1.14 with ismp still in the devnet runtime.

Before reviewing, see latest commit for something that was required to make `zepter` happy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,15 +44,15 @@ jobs:
         run: |
           cargo check --release --locked --features=runtime-benchmarks,try-runtime
 
-  check-ismp:
-    needs: lint
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: "./.github/actions/init"
-      - name: Check Build with ISMP
-        run: |
-          cargo check --release --locked --features=ismp,runtime-benchmarks,try-runtime
+#  check-ismp:
+#    needs: lint
+#    runs-on: ubuntu-latest
+#    steps:
+#      - uses: actions/checkout@v4
+#      - uses: "./.github/actions/init"
+#      - name: Check Build with ISMP
+#        run: |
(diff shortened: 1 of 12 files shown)


peterwht:
before approving this, I think it's worthwhile to see if we can revert back to our ISMP fork and update that repo as well. In the past it's been smooth with only PSVM being required.

See this commit: https://github.com/r0gue-io/ismp/commit/46862a00e2e227e8c77a6b9c1a14b278f6b64c06

# IMPORTANT NOTE:
The ISMP repo will NOT build standalone. You will verify its compilation through building pop-node with the ISMP modules pointing to your new ISMP branch. This is because we are only using the ISMP pallet.

.github/workflows/ci.yml
```diff
@@ -44,15 +44,15 @@ jobs:
...
-  check-ismp:
-    needs: lint
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-      - uses: "./.github/actions/init"
-      - name: Check Build with ISMP
-        run: |
-          cargo check --release --locked --features=ismp,runtime-benchmarks,try-runtime
+#  check-ismp:
```
chungquantin, line 47: Since this has been merged into `main`, it's worth leaving clear TODOs for the commented-out code.

chungquantin (commented):
If everything related to ISMP removed, why don't we fork `main` branch to `ismp`, create a PR and then remove everything related to ismp in `main`. This avoid leaving all the commented-out code in everywhere in the file. 

Edit: Saw you removed review requests, I'll leave my comments here anyway.

PR #234, opened 2024-08-23
chore: upgrade to stable2407

chungquantin:
**Notable changes in terms of pallet contracts:**
- [Add support for transient storage in contracts host functions](https://github.com/paritytech/polkadot-sdk/pull/4566): Introduce the `type MaxTransientStorageSize = ConstU32<{ 1024 * 1024 }>` in the runtimes.

IMPORTANT NOTE: this doesn't compile due to westend runtimes. Please compare this PR and once approved we can merge https://github.com/r0gue-io/pop-node/pull/412 and run the CI.

PR #417, opened 2025-01-03
chore: upgrade to stable2412

peterwht:
- [x] 2407 https://github.com/r0gue-io/pop-node/pull/234
- [x] 2409 https://github.com/r0gue-io/pop-node/pulls?q=is%3Apr+is%3Aclosed
- [x] 2412 https://github.com/r0gue-io/pop-node/pull/425


chungquantin:
Not important but when uplifting the `contracts-parachain`, there is a different setting for `MaxCodeLen` for pallet-revive that we need to pay attention: https://github.com/r0gue-io/contracts-parachain/pull/12#discussion_r1905688864

Newer versions remove this check so we should use `stable2412` version of `pallet-revive`.

Daanvdplas (approved):
LGTM!

al3mart (approved):
Looks good!

Things like default weight usage can, and should, be addressed outside of this PR with the changes coming with the mainnet runtime review:

PR #421, opened 2025-01-08
refactor: account balance

Daanvdplas:
- Main thing is that I combined the two tests relevant to the account balance (account balance update and deposit). 
- The deposit tests were not checking the reserved balance so added that. 
- Refactored the tests to `common_functions` as it much cleaner (separating unit tests to their own file). 
- Changed the `depositor` for the atomic swap to the destination accounts.
- Left some comments to clarify how something works and why

### DIFF
--- a/pallets/nfts/src/tests.rs
+++ b/pallets/nfts/src/tests.rs
@@ -18,7 +18,7 @@
 //! Tests for Nfts pallet.
 
 use enumflags2::BitFlags;
-use frame_support::{
+pub(crate) use frame_support::{
 	assert_noop, assert_ok,
 	dispatch::WithPostDispatchInfo,
 	traits::{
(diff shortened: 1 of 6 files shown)


pallets/nfts/src/tests.rs
```diff
@@ -38,19 +38,15 @@ use sp_runtime::{
...
 type CollectionApprovals = crate::CollectionApprovals<Test>;
 type CollectionApprovalDeposit = <Test as Config>::CollectionApprovalDeposit;
 type WeightOf = <Test as Config>::WeightInfo;
 
-fn account(id: u8) -> AccountIdOf<Test> {
+pub(crate) fn account(id: u8) -> AccountIdOf<Test> {
 	[id; 32].into()
 }
 
-fn root() -> RuntimeOrigin {
```
Daanvdplas, line 50: Removed as it was used very inconsistently in the added code.

chungquantin (approved):
LGTM!

PR #422, opened 2025-01-09
Psp2 Dao_example0

ndkazu:
### Description

Issue #337 
This contract implements a Decentralized Autonomous Organization using Psp22. 
The key functionalities include:

- **Membership Management:** It maintains a registry of DAO members.
- **Proposal Lifecycle:** The contract manages the creation, voting, and execution of proposals. Each proposal includes details like description, voting period, vote tallies, execution status, beneficiary, and amount to be awarded.
- **Voting Mechanism:** It implements a voting system where members can cast votes with their balance on proposals. The contract tracks voting periods and maintains vote counts for each proposal.
- **Token Integration:** The DAO is associated with a specific Psp22 token_id. 
- **Governance Parameters:** governance parameters such as voting periods are customizable.
- **Vote Tracking:** The contract keeps track of when members last voted.
- **Proposal Execution:** Once a proposal's voting period ends and it passes, the contract handles the execution of the proposal: Transfer funds to the chosen beneficiary.]

## ToDo
- **Tests** 
- [x]  Create a new Dao works
- [x]   A new member joins the Dao works
- [x]   Member creates a proposal works
- [x]   Members vote system works
- [x]   Proposal enactment works
- [x]   Voting twice for the same proposal fails
- [x]   Vote fails if not a member
- [x]   Vote fails if voting period ended
- [x]   Proposal enactment fails if proposal was rejected
- [x]   Same proposal consecutive claim fails
- **Others**
- [x] Events
- [x] in-code documentation

## Remark

Considering the fact that  [another issue](https://github.com/r0gue-io/pop-node/issues/378) is asking for psp22 examples using cross contract call, I decided to use `api::...` instead of the macro `contract_ref!`.


Daanvdplas (commented):
Amazing work!

Right now the example doesn't make sense to me or it is not "fun" enough. This dao funds projects with its own governance token, I think it should be a token holding value, like DOT / USDT / USDC (much cooler!). Moreover, people buy voting power with the governance token, why is their balance of the governance token simply not checked using the api? Then we don't need all the storage in the contract. Doing some research to token based daos, this is also the main implementation of it (e.g. uniswap dao). I think doing some research to how a token based dao works would be good and see what makes sense with e.g. USDT funding proposals.

Moreover, the goal of this contract example is to show how easy it is to interact with the fungibles api and build your use case around it. Right now we are only using the `create` and `transfer_from` methods, can we think of more (e.g. `mint` or `burn`, minting when buying the governance token, burning when voting or creating a proposal). This should only be added if it makes sense with the overall implementation of the dao though. Alternative is interacting with multiple tokens (e.g. usdt). Also, leave out the word psp22, this is not applicable here. A contract is compliant to this standard only if it contains all functions and events specified in the standard.

As we had a conversation somewhere else I will stop my review now. Next steps are to define the story of the dao, then finalise the contract and tests.

PR #423, opened 2025-01-09
refactor(xcm)!: restrict reserves to AssetHub, remove relay chain support

Daanvdplas:
**BREAKING CHANGE**

Changes the xcm config to only accept reserve transfers of the dot token from AH. As for the mainnet runtime this will function as the only accepted reserve for now.

A small refactor of the structure of the features in the integration tests have been applied to simplify things, supported with a small readme. 

The CI has received some optimisations to exclude the integration tests. Only the testnet integration tests are added to the CI for now. This can also be changed to all pop's runtimes with westend / paseo runtime. Adding them all could also be done but it feels like an overkill to run for each PR. Open to suggestions.

[sc-2066]


runtime/devnet/src/config/xcm.rs
```diff
@@ -112,18 +110,37 @@ pub type Barrier = TrailingSetTopicAsId<(
...
-impl<T: Get<Location>> ContainsPair<Asset, Location> for NativeAssetFrom<T> {
+/// Accepts the relay asset from Asset Hub.
+pub struct RelayAssetFromAssetHub;
+impl ContainsPair<Asset, Location> for RelayAssetFromAssetHub {
 	fn contains(asset: &Asset, origin: &Location) -> bool {
-		let loc = T::get();
+		let loc = AssetHub::get();
 		&loc == origin &&
 			matches!(asset, Asset { id: AssetId(asset_loc), fun: Fungible(_a) }
 			if *asset_loc == Location::from(Parent))
```
Daanvdplas, line 120: Didn't see value in having this generic as it always looks for the asset location of the relay and we won't use it for a different chain.

integration-tests/src/lib.rs
```diff
@@ -270,7 +270,9 @@ fn fund_pop_from_system_para(
 }
 
 /// Reserve Transfers of native asset from Relay to Parachain should work
+#[cfg(not(feature = "testnet"))]
```
Daanvdplas, line 273: Can be removed in the release where we set the reserve of dot to asset hub on testnet as well.

integration-tests/src/lib.rs
```diff
@@ -317,7 +319,9 @@ fn reserve_transfer_native_asset_from_relay_to_para() {
 }
 
 /// Reserve Transfers of native asset from Parachain to Relay should work
+#[cfg(not(feature = "testnet"))]
 #[test]
+#[should_panic]
 fn reserve_transfer_native_asset_from_para_to_relay() {
```
Daanvdplas, line 325: This test can be removed in the next release as well as it should not be possible to do a reserve transfer to the relay with only AH as reserve.

runtime/mainnet/src/config/xcm.rs
```diff
@@ -103,18 +101,37 @@ pub type Barrier = TrailingSetTopicAsId<(
...
+		// Exclude the relay location.
+		if matches!(origin, Location { parents: 1, interior: Here }) {
+			return false;
+		}
+		matches!(asset.id, AssetId(ref id) if id == origin)
+	}
+}
+
+/// Combinations of (Asset, Location) pairs which we trust as reserves.
+pub type TrustedReserves = (NativeAssetExceptRelay, RelayAssetFromAssetHub);
```
al3mart, line 134: While I don't think this presents a problem, I'm have some hesitations around including more reserves on mainnet than relay native's for now.

Channels are still required for any transfer to occur. But maybe for that same reason we expand the reserves when that moment comes ?

I'd like to hear your considerations as I can think of opinions for both, keeping it and removing it. :)

runtime/devnet/src/config/xcm.rs
al3mart: Ok given that having multiple reserves for the same asset is not a good practice. And we most likely won't want to have many, for instance, system chains acting as the reserve of DOT.

But I liked better `NativeAssetFrom<AssetHub>` than `RelayAssetFromAssetHub` :)

al3mart (changes requested):
Left a few nits, but the changes are good and tests run green!

Some easy unit testing for `NativeAssetExceptRelay` and `RelayAssetFromAssetHub` showing than the results are the expected given a list of locations would have been nice!
Or, a very small integration test checking the reserves if the above is not really something we want given we have 3x runtimes.

runtime/mainnet/src/config/xcm.rs
Daanvdplas: I see no reason to remove it as like you said, we need channels open first with a chain. Perhaps we want chains' native token to always have AH as reserve but that might not be desired by the chain itself. I don't see a reason for removing it because I can't think of a chain that we want to open channels with and don't want to accept their token. 

@peterwht

Daanvdplas:
> But I liked better NativeAssetFrom<AssetHub> than RelayAssetFromAssetHub :)

I don't get this comment, code should be as easy as possible so if you agree we should remove the generic. As for the change from native asset to relay asset, native asset is unclear to me as it sounds as the native asset of pop, which is correct, however I don't find it self explanatory that the relay asset is the native asset of pop.

runtime/devnet/src/config/xcm.rs
Daanvdplas: I don't get this comment, code should be as easy as possible so if you agree we should remove the generic. As for the change from native asset to relay asset, native asset is unclear to me as it sounds as the native asset of pop, which is correct, however I don't find it self explanatory that the relay asset is the native asset of pop.

The integration tests would also need this change of terminology to my opinion but I wanted to first hear other's opinion and this should be done in a different PR.

integration-tests/src/lib.rs
al3mart: This can be resolved.

integration-tests/src/lib.rs
al3mart: This can be resolved.

al3mart (commented):
Last changes look great!
Happy to approve once we get the above messages resolved, although they are not really blockers here.

runtime/devnet/src/config/xcm.rs
```diff
@@ -112,18 +110,37 @@ pub type Barrier = TrailingSetTopicAsId<(
 	>,
 )>;
 
-/// Asset filter that allows native/relay asset if coming from a certain location.
-// Borrowed from https://github.com/paritytech/polkadot-sdk/blob/ea458d0b95d819d31683a8a09ca7973ae10b49be/cumulus/parachains/runtimes/testing/penpal/src/xcm_config.rs#L239 for now
-pub struct NativeAssetFrom<T>(PhantomData<T>);
-impl<T: Get<Location>> ContainsPair<Asset, Location> for NativeAssetFrom<T> {
+/// Accepts the relay asset from Asset Hub.
+pub struct RelayAssetFromAssetHub;
+impl ContainsPair<Asset, Location> for RelayAssetFromAssetHub {
```
peterwht, line 115: I feel this change unnecessary as well. The simplest change would be to change to `TrustedReserves = NativeAssetFrom<AssetHub>`. To me, this is quite clear that we are using the native asset from AssetHub. If this is not self-explanatory to be using DOT, I would prefer a simple comment above our `TrustedReserves`

Additional reasons for reverting this change:
- Less noise for reviewer
- `NativeAssetFrom` is more commonly used 
- If we ever decide we want to accept only some native assets, it's as simple as adding `NativeAssetFrom<SomeChain>`.

.github/workflows/ci.yml
```diff
@@ -90,7 +90,7 @@ jobs:
         uses: actions-rs/clippy-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          args: --release --locked --features=runtime-benchmarks,ismp
+          args: --release --locked --workspace --exclude integration-tests --features=runtime-benchmarks,ismp
```
peterwht, line 93: I'm not sure we should do this change.

I realize all of the 2412 PRs are currently failing. However, IMO, it's a conscious decision to let them fail. Integration tests failing is a temporary thing until the 2412 release is fixed. So, in order to keep it in our mind, I personally prefer to see the integration tests CI failing so that it reminds me to run it manually on my local machine.

By excluding from the CI, it's easier to let things slip through the cracks and we also lose the reminder to fix once 2412 is fixed.

peterwht (changes requested):
A few changes please! IMO, this PR has scope creep introducing some unnecessary noise for what could have been simpler.

One additional change:
- Please update the PR title to be something like: "refactor(xcm)!: restrict reserves to AssetHub, remove relay chain support". Also note that this marks this PR as a BREAKING CHANGE. Please update the PR description to include "BREAKING CHANGE".
    - Considering that relay chain reserves no longer works, this is a breaking change for testnet. Just to be mindful, this could potentially break existing contracts on testnet.

.github/workflows/ci.yml
Daanvdplas: I don't get your point, the CI runs the integration tests (currently not all pop runtimes though) so the CI will still show you that the integration tests don't work

.github/workflows/ci.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
       - name: Check Build
         run: |
-          cargo check --release --locked --features=runtime-benchmarks,try-runtime
+          cargo check --release --locked --workspace --exclude integration-tests --features=runtime-benchmarks,try-runtime
```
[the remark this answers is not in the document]
Daanvdplas: Will remember for next time! Keeping it in for now.

integration-tests/src/lib.rs
```diff
@@ -124,24 +123,6 @@ fn para_to_system_para_sender_assertions(t: ParaToSystemParaTest) {
 	);
 }
 
-fn para_to_relay_sender_assertions(t: ParaToRelayTest) {
```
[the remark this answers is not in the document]
Daanvdplas: I don't get this comment

integration-tests/src/lib.rs
```diff
@@ -271,6 +200,7 @@ fn fund_pop_from_system_para(
 
 /// Reserve Transfers of native asset from Relay to Parachain should work
 #[test]
+#[should_panic]
```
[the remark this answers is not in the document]
Daanvdplas: This is to my opinion scope creep

runtime/devnet/src/config/xcm.rs
Daanvdplas: Changed back in [8d122db](https://github.com/r0gue-io/pop-node/pull/423/commits/8d122dbbf0382ebee2a99816c6503ff67e2d1271)

runtime/devnet/src/config/xcm.rs
Daanvdplas: Changed back [29baf90](https://github.com/r0gue-io/pop-node/pull/423/commits/29baf90446a0f78335f7f7459dc778957b86677a)

integration-tests/src/lib.rs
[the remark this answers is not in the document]
peterwht: I wouldn't consider adding tests to ensure changes in functionality is correct as scope creep. However, happy to address at a later time!

runtime/mainnet/src/config/xcm.rs
peterwht: Needs addressing before a re-review

.github/workflows/ci.yml
peterwht: fair enough. I did misunderstand some there. I do see that the tests are still run, it's just not included in the initial checks. I am good with that then.

peterwht:
Sorry for being a stickler, but I still don't consider the PR title accurate.

Breaking changes should include an `!`. This is also not a chore and should be marked as a `refactor!`. This is because we are changing existing code without introducing new functionality nor fixing a bug.

Daanvdplas:
Didn't know about the style of using an explanation mark for breaking changes. My view is that a refactor is less of a impactful change than what pop's accepted reserves are on the protocol level. 

Changed the name of the PR

runtime/mainnet/src/config/xcm.rs
Daanvdplas: Happened to have squeezed it in [29baf90](https://github.com/r0gue-io/pop-node/pull/423/commits/29baf90446a0f78335f7f7459dc778957b86677a), apologies!

peterwht:
> Didn't know about the style of using an explanation mark for breaking changes. My view is that a refactor is less of a impactful change than what pop's accepted reserves are on the protocol level.
> 
> Changed the name of the PR

Very much appreciated! I think it's something we all need to learn more. I've been basing this largely off of this: https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#-commit-message-guidelines

But we can always make our own guidelines. For example, I could see this PR as a `feat` or a `fix`. But based on the angular guidelines a refactor is best suited.

peterwht (approved):
Thanks for sticking with my nitpicks! Good to go now

integration-tests/src/lib.rs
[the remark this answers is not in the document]
Daanvdplas: As discussed privately, this will be handled in a separate PR.

runtime/devnet/src/config/xcm.rs
```diff
@@ -123,7 +123,27 @@ impl<T: Get<Location>> ContainsPair<Asset, Location> for NativeAssetFrom<T> {
 			if *asset_loc == Location::from(Parent))
 	}
 }
-pub type TrustedReserves = (NativeAsset, NativeAssetFrom<AssetHub>);
```
chungquantin, line 126: `NativeAsset` is no longer used anywhere.

chungquantin (approved):
It looks good to me. The tests and new logic changes make sense. Not much value can be added from my end but I notice a few minor clippy warnings that should be resolved. (e.g. `NativeAsset`)

Integration tests for runtimes works completely fine on my local.

runtime/devnet/src/config/xcm.rs
Daanvdplas: Thank you!

runtime/mainnet/src/config/xcm.rs
```diff
@@ -211,3 +213,54 @@ impl cumulus_pallet_xcm::Config for Runtime {
...
+	}
+
+	#[test]
+	fn relay_as_relay_asset_reserve_fails() {
+		let relay_asset = Asset::from((AssetId::from(Parent), Fungibility::from(100u128)));
+		assert!(!TrustedReserves::contains(&relay_asset, &Parent.into()));
+	}
+
+	#[test]
+	fn decline_native_assets() {
```
al3mart, line 247: Could use a comment to point that this is every other native asset but the system one.

al3mart (approved):
Looking pretty good. Thanks!

Left a change for a test name as it seemed more correct.

runtime/mainnet/src/config/xcm.rs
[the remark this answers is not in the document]
Daanvdplas: [b7342fa](https://github.com/r0gue-io/pop-node/pull/423/commits/b7342fa48ac444188a1ee9160b5a28e227d3fa9c)

al3mart (approved):

PR #424, opened 2025-01-09
fix: remove old dependencies and include Cargo.lock in VCS

Daanvdplas:
Removes the `Cargo.lock` file in the `.gitignore` so that the same is used as locally, this solves the CI errors.

Note: we could consider to add all the modules to the workspace which would reduce the amount of required `Cargo.lock` files, but this takes a little more time and might not be desired (mostly because for the example contracts we want crates versions so that it can help devs to use the contract).


.gitignore
```diff
@@ -13,6 +13,3 @@
 
 # Binaries
 **/bin/
-
-# Cargo.lock
-**/Cargo.lock
```
Daanvdplas, line 18: Removed the `Cargo.lock` file so that the same Cargo.lock is used everywhere as locally.

pop-api/.gitignore
```diff
@@ -4,10 +4,6 @@
 debug/
 target/
 
-# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
-# More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
-Cargo.lock
```
Daanvdplas, line 9: Same

pop-api/examples/fungibles/Cargo.toml
```diff
@@ -15,12 +15,6 @@ drink = { package = "pop-drink", git = "https://github.com/r0gue-io/pop-drink" }
 env_logger = { version = "0.11.3" }
 serde_json = "1.0.114"
 
-# TODO: due to compilation issues caused by `sp-runtime`, `frame-support-procedural` and `staging-xcm` this dependency
```
Daanvdplas, line 18: Removed as it isn't needed anymore

pop-api/integration-tests/Cargo.toml
```diff
@@ -26,9 +25,7 @@ scale = { package = "parity-scale-codec", version = "3.6.12", default-features =
 	"derive",
 ] }
 sp-io = { version = "37.0.0", default-features = false }
-sp-runtime = { version = "=38.0.0", default-features = false }
-# TODO: Requires to resolve a dependency version issue. See more at: https://github.com/r0gue-io/pop-node/issues/313.
```
Daanvdplas, line 30: Same as above, these dependencies were only added to make it compile.

Daanvdplas (commented):
The `TODO` was added in this PR back at the time: https://github.com/r0gue-io/pop-node/pull/317/files

chungquantin (approved):
Looks great!

al3mart (approved):
Awesome, thanks!

PR #426, opened 2025-01-12
feat(mainnet): add XCM runtime APIs, config tests, and refactor to API mod [DON'T MERGE] 

peterwht:
**DO NOT SQUASH MERGE -- REMOVE THIS MESSAGE BEFORE MERGE**
Prereq: https://github.com/r0gue-io/pop-node/pull/425

This PR has various improvements to the mainnet runtime. Please review the individual commits:

[test(mainnet): more configuration tests](https://github.com/r0gue-io/pop-node/pull/426/commits/26592f9203772273d4e66162f3581ed437bdaa21)
[refactor(mainnet): runtime APIs to apis module](https://github.com/r0gue-io/pop-node/pull/426/commits/b04efa2ce93de552c17df4d8dc121bd75756d44c)
[feat(mainnet): add XCM runtime APIs](https://github.com/r0gue-io/pop-node/pull/426/commits/4d18a39a3ff312889229d6e3fe4f6fc4274009af)

PR #425, opened 2025-01-09
chore: uplift to polkadot 2412

peterwht:
Uplifts to [Stable 2412](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-stable2412).

Due to dependency issues with `asset-hub-westend-runtime` with the 2412 release, `integration-tests` have been included from the workspace. The integration tests have been updated and confirmed to pass. 

### Relevant PRs causing node changes

- https://github.com/paritytech/polkadot-sdk/pull/6156
- https://github.com/paritytech/polkadot-sdk/pull/4639
- https://github.com/paritytech/polkadot-sdk/pull/5919

### Relevant PRs causing runtime changes
- https://github.com/paritytech/polkadot-sdk/pull/5693
- https://github.com/paritytech/polkadot-sdk/pull/3685
    - this includes a change to `dispatch_info` changing to hold `call_weight` and `extension_weight`

### TODOs (in follow up PRs, non-blocking for this PR)
- [x] Add pallet-revive https://github.com/r0gue-io/pop-node/pull/427
- [x] Add xcm_runtime_apis: https://github.com/r0gue-io/pop-node/pull/426
- [ ] Add SystemExtensionsBench | https://github.com/paritytech/polkadot-sdk/pull/3685 
    - [ ] https://github.com/r0gue-io/pop-node/issues/435

PR #427, opened 2025-01-12
feat(mainnet): add revive [DON'T MERGE]

peterwht:
Prereq: https://github.com/r0gue-io/pop-node/pull/426 & https://github.com/r0gue-io/pop-node/pull/425

This PR adds pallet-revive to the mainnet runtime. This was added directly to mainnet as do not have a running mainnet, so we can start preparing the mainnet runtime for the launch. 

The revive configuration and pallet name (in our runtime config) will be contracts.rs and `Contracts`, respectively. 

Please review the commits and commit messages for individual components:

[feat(mainnet): add pallet-revive](https://github.com/r0gue-io/pop-node/commit/285ea11864ea3224a96e4ea5d702ea5ef767d18a)

[test(mainnet): revive configuration tests](https://github.com/r0gue-io/pop-node/commit/184ec1d71214c4959e1da6d51b1cf19ab4ab46fc)

[feat(mainnet): add EthExtraImpl for UncheckedExtrinsic](https://github.com/r0gue-io/pop-node/commit/16ca362e7f73cd0db3053976e36f907c7b0e4e82)

[feat(mainnet): add revive runtime api](https://github.com/r0gue-io/pop-node/commit/e096ef929940e61b905093f448aa54f8b4aae1cd)

### DIFF
--- a/runtime/mainnet/src/config/contracts.rs
+++ b/runtime/mainnet/src/config/contracts.rs
@@ -6,13 +6,18 @@ use frame_system::EnsureSigned;
 
 use crate::{
 	deposit, Balance, Balances, Perbill, Runtime, RuntimeCall, RuntimeEvent, RuntimeHoldReason,
-	Timestamp,
+	Timestamp, UNIT,
 };
 
+const ETH: u128 = 1_000_000_000_000_000_000;
+
 parameter_types! {
+	// TODO: review implications of this
 	pub const DepositPerItem: Balance = deposit(1, 0);
+	// TODO: review implications of this
 	pub const DepositPerByte: Balance = deposit(0, 1);
 	pub CodeHashLockupDepositPercent: Perbill = Perbill::from_percent(30);
+	pub const NativeToEthRatio: u32 = (ETH/UNIT) as u32;
 }
 
 impl pallet_revive::Config for Runtime {
@@ -26,7 +31,7 @@ impl pallet_revive::Config for Runtime {
 	type DepositPerByte = DepositPerByte;
 	type DepositPerItem = DepositPerItem;
 	type InstantiateOrigin = EnsureSigned<Self::AccountId>;
-	type NativeToEthRatio = ConstU32<1>;
+	type NativeToEthRatio = NativeToEthRatio;
 	// TODO: review implications of this
 	type PVFMemory = ConstU32<{ 512 * 1024 * 1024 }>;
 	type RuntimeCall = RuntimeCall;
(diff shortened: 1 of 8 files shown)


peterwht:
[sc-2219]

runtime/mainnet/src/lib.rs
```diff
@@ -667,6 +691,10 @@ mod runtime {
 	#[runtime::pallet_index(33)]
 	pub type MessageQueue = pallet_message_queue::Pallet<Runtime>;
 
+	// Contracts (using pallet-revive)
+	#[runtime::pallet_index(40)]
+	pub type Contracts = pallet_revive::Pallet<Runtime>;
```
evilrobot-01, line 696: Whilst this would be my preference, there is a risk that tooling does not support this (e.g. Polkadot JS which loads the existing contracts UI but expects wasm contracts). Determining support for this approach might be important before any launch.

runtime/mainnet/src/config/contracts.rs
```diff
@@ -0,0 +1,227 @@
...
+
+	use super::*;
+
+	// 18 decimals
+	const ONE_ETH: u128 = 1_000_000_000_000_000_000;
+
+	#[test]
+	fn address_mapper_is_account_id32_mapper() {
+		assert_eq!(
+			TypeId::of::<<Runtime as pallet_revive::Config>::AddressMapper>(),
```
evilrobot-01, line 75: Based on scope, probably easiest to import pallet_revive::Config to avoid the pallet_revive:: prefix repetition. The frame_system can be prefixed by exception if required.

runtime/mainnet/src/lib.rs
peterwht: really good point. Sounds like it would just be more headache than it's worth.

runtime/mainnet/src/config/contracts.rs
```diff
@@ -0,0 +1,227 @@
...
+	fn chain_extension_is_unset() {
+		assert_eq!(
+			TypeId::of::<<Runtime as pallet_revive::Config>::ChainExtension>(),
+			TypeId::of::<()>(),
+		);
+	}
+
+	#[test]
+	fn chain_id_is_correct() {
+		assert_eq!(<<Runtime as pallet_revive::Config>::ChainId as Get<u64>>::get(), 3395);
```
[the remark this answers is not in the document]
peterwht: You are correct. Technically, can be added manually to overwrite what is in Metamask, but it will give an error / warning: "network ID taken already".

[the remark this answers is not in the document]
al3mart: This might be useful: https://chainlist.org/?testnets=true

runtime/mainnet/src/lib.rs
al3mart: Hopefully new tooling will do a better job with this sort of things. I believe some version of the ledger app also suffered from something similar to this with the balances pallet.

al3mart (commented):
I'd agree with the change of name from Contracts to Revive for the new pallet.
The memory limits for revive config seem alright. The pallet integrity check implementation provides some more context around them, but it looks like we should be OK with the default values for now.

Aside from that all the changes look great!

One for the aesthetics.
In the pallet config and its tests we are using both, the type for other pallets present in the runtime, as `Balances` or `Timestamp` and the full qualified name for others, like `pallet_transaction_payment` or `pallet_xcm`. I'd argue that it will look better if we stick with only one of the options across the config.

The above examples:

- [`Balances`](https://github.com/r0gue-io/pop-node/blob/c137ddc5293c3e0cff469e2797c38f065961fcdb/runtime/mainnet/src/config/contracts.rs#L29) used to provide `Currency`.
- [`Timestamp`](https://github.com/r0gue-io/pop-node/blob/c137ddc5293c3e0cff469e2797c38f065961fcdb/runtime/mainnet/src/config/contracts.rs#L42) used to provide `Time`.
- [`pallet_transaction_payment::Pallet<Self>`](https://github.com/r0gue-io/pop-node/blob/c137ddc5293c3e0cff469e2797c38f065961fcdb/runtime/mainnet/src/config/contracts.rs#L46C21-L46C62) used to provide `WeightPrice`.
- [`pallet_xcm::Pallet<Self>`](https://github.com/r0gue-io/pop-node/blob/c137ddc5293c3e0cff469e2797c38f065961fcdb/runtime/mainnet/src/config/contracts.rs#L46C21-L46C62) used to provide `Xcm`.

runtime/mainnet/src/config/contracts.rs
```diff
@@ -0,0 +1,54 @@
...
+	type ChainExtension = ();
+	type ChainId = ConstU64<3395>;
+	type CodeHashLockupDepositPercent = CodeHashLockupDepositPercent;
+	type Currency = Balances;
+	type Debug = ();
+	type DepositPerByte = DepositPerByte;
+	type DepositPerItem = DepositPerItem;
+	type InstantiateOrigin = EnsureSigned<Self::AccountId>;
+	type NativeToEthRatio = ConstU32<1>;
+	// TODO: review implications of this
```
Daanvdplas, line 35: When do you want to review this? Seems to be a perfect time to do it now if we add it.

Daanvdplas:
Not sure why `final review` label is used and the CI is failing + `TODO` comments are added in mainnet code. We should have clear guidelines here or better communicated in the PR description.

runtime/mainnet/src/lib.rs
peterwht: Probably worthwhile to open a forum post

runtime/mainnet/src/config/contracts.rs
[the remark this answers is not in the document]
peterwht: Thanks Ale! Checking those two resources, in addition to Metamask, 3395 is still available for use.

runtime/mainnet/src/config/contracts.rs
```diff
@@ -0,0 +1,54 @@
...
+	pub const DepositPerItem: Balance = deposit(1, 0);
+	pub const DepositPerByte: Balance = deposit(0, 1);
+	pub CodeHashLockupDepositPercent: Perbill = Perbill::from_percent(30);
+}
+
+impl pallet_revive::Config for Runtime {
+	type AddressMapper = pallet_revive::AccountId32Mapper<Self>;
+	type CallFilter = Nothing;
+	type ChainExtension = ();
+	type ChainId = ConstU64<3395>;
```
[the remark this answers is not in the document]
peterwht: Definitely would like to see this done. I created a story here to track: https://app.shortcut.com/r0gue/epic/2251?group_by=none&vc_group_by=day&ct_workflow=all&cf_workflow=500000005

runtime/mainnet/src/config/contracts.rs
peterwht: https://github.com/r0gue-io/pop-node/pull/427/commits/aad400171cdaa27cb6365b811439ad534241053e

runtime/mainnet/src/lib.rs
peterwht: https://github.com/r0gue-io/pop-node/pull/427/commits/5971736c7d3eec99ecf9421872ad4394e0b433e9

runtime/mainnet/src/config/contracts.rs
```diff
@@ -0,0 +1,227 @@
...
+	type WeightPrice = pallet_transaction_payment::Pallet<Self>;
+	type Xcm = pallet_xcm::Pallet<Self>;
+}
+
+impl TryFrom<RuntimeCall> for pallet_revive::Call<Runtime> {
+	type Error = ();
+
+	fn try_from(value: RuntimeCall) -> Result<Self, Self::Error> {
+		match value {
+			RuntimeCall::Contracts(call) => Ok(call),
```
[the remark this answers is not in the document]
peterwht: https://github.com/r0gue-io/pop-node/pull/427/commits/5971736c7d3eec99ecf9421872ad4394e0b433e9

runtime/mainnet/src/config/contracts.rs
Daanvdplas: Reminder

peterwht:
superseded by https://github.com/r0gue-io/pop-node/pull/427

PR #429, opened 2025-01-14
fix: typo and param name in pallet-api `Read` trait 

cuteolaf:


pallets/api/src/lib.rs
```diff
@@ -12,15 +12,15 @@ mod mock;
...
+	/// The type of result returned.
 	type Result;
 
 	/// Determines the weight of the requested read, used to charge the appropriate weight before
 	/// the read is performed.
 	///
 	/// # Parameters
 	/// - `request` - The read request.
-	fn weight(read: &Self::Read) -> Weight;
+	fn weight(request: &Self::Read) -> Weight;
```
chungquantin, line 23: This should not be changed.

chungquantin (changes requested):

pallets/api/src/lib.rs
Daanvdplas: Then the comment above needs fixing

Daanvdplas (approved):
Thank you for going through our code and fixing where you find mistakes! Much appreciated!

PR #430, opened 2025-01-14
fix: wrong comment

cuteolaf:


chungquantin:
Thanks for creating this PR, but the comment is accurate. The Pop API is currently designed to accept only the `origin` as the contract. Therefore, the `origin` must pre-approve the contract for the `transfer()` function to work.

Daanvdplas (approved):
Thank you again for pushing this fix! Much appreciated!

Daanvdplas:
> Thanks for creating this PR, but the comment is accurate. The Pop API is currently designed to accept only the origin as the contract. Therefore, the origin must pre-approve the contract for the transfer() function to work.

This is incorrect, this is the case when you call `transfer_from`. In addition, the `from` parameter doesn't even exist so the comment does indeed not make sense.

chungquantin (approved):

PR #432, opened 2025-01-16
feat: add `pallet-api/nonfungibles`

chungquantin:
The non-fungibles pallet offers a streamlined interface for interacting with non-fungible assets. The goal is to provide a simplified, consistent API that adheres to standards in the smart contract space. Built on top of the optimized version of the `pallet-nfts` (See #387).

[sc-1242]


pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,613 @@
...
+		/// - `to` - The recipient account.
+		#[pallet::call_index(3)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::transfer() + T::DbWeight::get().reads_writes(1, 0))]
+		pub fn transfer(
+			origin: OriginFor<T>,
+			collection: CollectionIdOf<T>,
+			item: ItemIdOf<T>,
+			to: AccountIdOf<T>,
+		) -> DispatchResult {
+			ensure_signed(origin.clone())?;
```
Daanvdplas, line 237: Not sure whether this is necessary

Daanvdplas (commented):
Looking very good! Left a few comments based on a high level review. Ready to finalise everything as I didn't find anything concerning!

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,613 @@
...
+	use sp_std::vec::Vec;
+
+	use super::*;
+
+	/// State reads for the non-fungibles API with required input.
+	#[derive(Encode, Decode, Debug, MaxEncodedLen)]
+	#[cfg_attr(feature = "std", derive(PartialEq, Clone))]
+	#[repr(u8)]
+	#[allow(clippy::unnecessary_cast)]
+	pub enum Read<T: Config> {
```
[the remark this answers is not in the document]
Daanvdplas: Also the tests can be moved after the dispatchables.

Daanvdplas:
CI also failing which would be good to resolve for final review.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,613 @@
...
+		#[pallet::call_index(8)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::destroy(
+    		witness.item_metadatas,
+    		witness.item_configs,
+    		witness.attributes,
+		))]
+		pub fn destroy(
+			origin: OriginFor<T>,
+			collection: CollectionIdOf<T>,
+			witness: DestroyWitness,
```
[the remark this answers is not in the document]
chungquantin: I thought an issue to this mainly because we don't any front running situations happen? Like in the case of `mint`. But agree that only the authority can destroy the collection so this might not be a case.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,613 @@
...
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Transfers the collection item from the `origin` to account `to`.
+		///
+		/// # Parameters
+		/// - `collection` - The collection of the item to be transferred.
+		/// - `item` - The item to transfer.
+		/// - `to` - The recipient account.
+		#[pallet::call_index(3)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::transfer() + T::DbWeight::get().reads_writes(1, 0))]
```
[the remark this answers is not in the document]
chungquantin: Resolved by returning the post dispatch weight when error thrown for getting `owner`.

pallets/api/src/nonfungibles/mod.rs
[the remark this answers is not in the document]
chungquantin: PR for fungibles pallet: https://github.com/r0gue-io/pop-node/pull/444

pallets/api/src/nonfungibles/mod.rs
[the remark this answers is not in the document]
chungquantin: PR: https://github.com/r0gue-io/pop-node/pull/445/

Without the witness, I see the DX is improved significantly but the only concern is the pre-dispatch weight.

pallets/api/src/nonfungibles/weights.rs
```diff
@@ -0,0 +1,267 @@
...
+	/// Storage: `System::Account` (r:1 w:1)
+	/// Proof: `System::Account` (`max_values`: None, `max_size`: Some(128), added: 2603, mode: `MaxEncodedLen`)
+	/// The range of component `a` is `[0, 1]`.
+	/// The range of component `i` is `[0, 1]`.
+	fn approve(a: u32, i: u32, ) -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `606`
+		//  Estimated: `68341938741594600 + a * (3168 ±44_956_152_011_428) + i * (2663 ±729_819_634_434_010)`
+		// Minimum execution time: 13_000_000 picoseconds.
+		Weight::from_parts(25_055_102, 68341938741594600)
```
chungquantin, line 66: Wonder why this is unexpectedly high

pallets/api/src/lib.rs
```diff
@@ -7,6 +7,7 @@ pub mod extension;
 pub mod fungibles;
 #[cfg(test)]
 mod mock;
+pub mod nonfungibles;
```
Daanvdplas, line 10: Missing docs.

pallets/api/src/mock.rs
```diff
@@ -110,6 +115,90 @@ impl crate::fungibles::Config for Test {
...
+		&self,
+		_msg: L,
+		_signer: &<Self::Signer as IdentifyAccount>::AccountId,
+	) -> bool {
+		false
+	}
+}
+
+#[cfg(feature = "runtime-benchmarks")]
+impl pallet_nfts::pallet::BenchmarkHelper<u32, u32, Noop, u64, Noop> for () {
```
Daanvdplas, line 146: Just checking if this is necessary as I don't see it in pallet-nfts mock

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
+//! The non-fungibles pallet offers a streamlined interface for interacting with non-fungible
+//! assets. The goal is to provide a simplified, consistent API that adheres to standards in the
```
Daanvdplas, line 2:
```suggestion
//! tokens. The goal is to provide a simplified, consistent API that adheres to standards in the
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+};
+use sp_runtime::traits::StaticLookup;
+pub(crate) use weights::WeightInfo;
+
+#[cfg(feature = "runtime-benchmarks")]
+mod benchmarking;
+mod impls;
+#[cfg(test)]
+mod tests;
+pub mod weights;
```
Daanvdplas, line 25: Missing docs

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
+type CollectionIdOf<T> =
+	<NftsOf<T> as Inspect<<T as frame_system::Config>::AccountId>>::CollectionId;
+type ItemIdOf<T> = <NftsOf<T> as Inspect<<T as frame_system::Config>::AccountId>>::ItemId;
+type ItemPriceOf<T> = BalanceOf<T>;
+type CollectionDetailsOf<T> = CollectionDetails<AccountIdOf<T>, BalanceOf<T>>;
+type AttributeNamespaceOf<T> = AttributeNamespace<AccountIdOf<T>>;
+type CollectionConfigFor<T> =
+	CollectionConfig<ItemPriceOf<T>, BlockNumberFor<T>, CollectionIdOf<T>>;
+
```
Daanvdplas, line 36:
```suggestion
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+pub mod weights;
+
+type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
+type CollectionIdOf<T> =
+	<NftsOf<T> as Inspect<<T as frame_system::Config>::AccountId>>::CollectionId;
+type ItemIdOf<T> = <NftsOf<T> as Inspect<<T as frame_system::Config>::AccountId>>::ItemId;
+type ItemPriceOf<T> = BalanceOf<T>;
+type CollectionDetailsOf<T> = CollectionDetails<AccountIdOf<T>, BalanceOf<T>>;
+type AttributeNamespaceOf<T> = AttributeNamespace<AccountIdOf<T>>;
+type CollectionConfigFor<T> =
```
Daanvdplas, line 34:
```suggestion
type CollectionConfigOf<T> =
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+
+type NftsErrorOf<T> = pallet_nfts::Error<T, NftsInstanceOf<T>>;
+type NftsWeightInfoOf<T> = <T as pallet_nfts::Config<NftsInstanceOf<T>>>::WeightInfo;
+pub(super) type BalanceOf<T> =
+	<<T as pallet_nfts::Config<NftsInstanceOf<T>>>::Currency as Currency<
+		<T as frame_system::Config>::AccountId,
+	>>::Balance;
+pub(super) type NftsOf<T> = pallet_nfts::Pallet<T, NftsInstanceOf<T>>;
+pub(super) type NftsInstanceOf<T> = <T as Config>::NftsInstance;
+// Type aliases for pallet-nfts storage items.
```
Daanvdplas, line 45: Oke with removing but up to you if you see value to it!

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+//! smart contract space.
+
+use frame_support::{
+	dispatch::WithPostDispatchInfo,
+	traits::{nonfungibles_v2::Inspect, Currency},
+};
+use frame_system::pallet_prelude::BlockNumberFor;
+pub use pallet::*;
+use pallet_nfts::WeightInfo as NftsWeightInfoTrait;
+pub use pallet_nfts::{
```
Daanvdplas, line 12: Not sure whether this `pub` is necessary:
```suggestion
pub use pallet_nfts::{
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	/// The events that can be emitted.
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// Event emitted when allowance by `owner` to `operator` changes.
+		Approval {
```
Daanvdplas, line 84: Would have been nice to add a comment and explain how the event has slightly deviated from the spec. The `collection` has been added (which is good) but in stead of having to find out why myself for 10 min it could have easily be prevented by adding a small comment in this PR and add context. Same for the comment for this event, it is also different and would be nice to explain why, or add in the spec before asking for final review here. A spec is the source of truth and should almost be a literal copy paste, any changes or improvements made by the implementer should ideally be explained to prevent confusion.

Both changes are good though :)

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+	}
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	/// The events that can be emitted.
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// Event emitted when allowance by `owner` to `operator` changes.
```
Daanvdplas, line 83:
```suggestion
		/// Event emitted when allowance by `owner` to `operator` changes.
		// Differing style: event name abides by the PSP34 standard.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+			item: Option<ItemIdOf<T>>,
+			/// The owner providing the allowance.
+			owner: AccountIdOf<T>,
+			/// The beneficiary of the allowance.
+			operator: AccountIdOf<T>,
+			/// Whether allowance is set or removed.
+			approved: bool,
+		},
+		/// Event emitted when a token transfer occurs.
+		// Differing style: event name abides by the PSP22 standard.
```
Daanvdplas, line 98:
```suggestion
		// Differing style: event name abides by the PSP34 standard.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+		Transfer {
+			/// The collection ID.
```
Daanvdplas, line 100: Inconsistent with `collection` in `Approval` and `Created`. I would suggest `The collection identifier.`

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+			collection: CollectionIdOf<T>,
+			/// The collection item ID.
```
Daanvdplas, line 102: Also inconsistent
```suggestion
			/// The item which is transferred.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+			to: Option<AccountIdOf<T>>,
+		},
+		/// Event emitted when a collection is created.
+		Created {
+			/// The collection identifier.
+			id: CollectionIdOf<T>,
+			/// The creator of the collection.
+			creator: AccountIdOf<T>,
+			/// The administrator of the collection.
+			admin: AccountIdOf<T>,
```
Daanvdplas, line 116: Any reason for deviating from the `owner` parameter in `pallet-nfts` or is this forgotten in the copy paste of the fungibles event?

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,595 @@
...
+		type WeightInfo: WeightInfo;
+	}
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	/// The events that can be emitted.
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
+	pub enum Event<T: Config> {
```
Daanvdplas, line 82: Missing `AttributeSet` event from spec.

Daanvdplas (commented):
Stopped reviewing after the events in `mod.rs` because I have found too many deviations from the spec without any explanation. Could you please give the PR a final review yourself while looking at the spec (especially found different comments) and explain any changes made to prevent confusion :)

pallets/api/src/lib.rs
chungquantin: There is no doc for the exported extension and fungibles as well.

pallets/api/src/nonfungibles/mod.rs
chungquantin: no comment for fungibles & nfts pallet as well.

pallets/api/src/nonfungibles/mod.rs
chungquantin: Those types will be used to implement the pop-api for non-fungibles and integration tests. Even though they can be imported from pallet_nfts, making this public to avoid adding `pallet-nfts` crate.

pallets/api/src/lib.rs
Daanvdplas: We should always write docs for `pub` from now on. Any missing comments have to be resolved when time allows in separate PRs

pallets/api/src/nonfungibles/mod.rs
chungquantin: I made it on purpose. I don't think the `owner` field in the `pallet-nfts` makes sense because it is not the owner but the admin. And this also makes it similar to the one we have in `pallet-api/fungibles`

pallets/api/src/mock.rs
chungquantin: It is required. Otherwise you will get this

```rs
 type Helper = ();
    |                   ^^ the trait `pallet_nfts::BenchmarkHelper<u32, u32, mock::Noop, u64, mock::Noop>` is not implemented for `()`
```

pallets/api/src/nonfungibles/mod.rs
chungquantin: Resolved by adding the `AttributeSet`

pallets/api/src/nonfungibles/mod.rs
chungquantin: https://github.com/r0gue-io/pop-node/issues/446

pallets/api/src/lib.rs
chungquantin: Issue: https://github.com/r0gue-io/pop-node/issues/446

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+			item: Option<ItemIdOf<T>>,
+			/// The key for the attribute.
+			key: Vec<u8>,
+			/// The data for the attribute.
+			data: Vec<u8>,
+		},
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
```
chungquantin, line 131: Similar to the events, to support multiple tokens, we also need to accept `collection` as parameters to dispatchables related to a non-fungible token.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+			/// The collection item identifier.
+			item: ItemIdOf<T>,
+			/// The namespace of the attribute.
+			namespace: AttributeNamespaceOf<T>,
+			/// The key of the attribute.
+			key: BoundedVec<u8, T::KeyLimit>,
+		},
+		/// Details of a specified collection.
+		#[codec(index = 9)]
+		Collection(CollectionIdOf<T>),
```
chungquantin, line 506: The following `Read` operations, along with the others listed below, are additional reads not specified in the standard or spec. However, I believe they are essential to the design of the Pop API's nonfungible feature. Another potential Read to consider is `Item(CollectionIdOf<T>, ItemIdOf<T>)`, which queries the details of a collection item.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+		}
+
+		/// Cancel all the approvals of a specific item.
+		///
+		/// # Parameters
+		/// - `collection` - The collection of the item of whose approvals will be cleared.
+		/// - `item` - The item of the collection of whose approvals will be cleared.
+		#[pallet::call_index(5)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::clear_all_transfer_approvals())]
+		pub fn clear_all_transfer_approvals(
```
chungquantin, line 197: Additional dispatchables added that are not mentioned in the spec.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+		}
+
+		/// Cancel approvals to transfer all owner's collection items.
+		///
+		/// # Parameters
+		/// - `collection` - The collection whose approvals will be cleared.
+		/// - `limit` - The amount of collection approvals that will be cleared.
+		#[pallet::call_index(6)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::clear_collection_approvals(*limit))]
+		pub fn clear_collection_approvals(
```
chungquantin, line 212: Additional dispatchables added that are not mentioned in the spec.

pallets/api/src/nonfungibles/tests.rs
```diff
@@ -0,0 +1,1259 @@
...
+const ITEM: u32 = 1;
+
+type CollectionApprovals = pallet_nfts::CollectionApprovals<Test, NftsInstanceOf<Test>>;
+type Event = crate::nonfungibles::Event<Test>;
+type NftsError = pallet_nfts::Error<Test, NftsInstanceOf<Test>>;
+type WeightInfo = <Test as Config>::WeightInfo;
+
+/// Information about a collection.
+#[derive(Clone, Encode, Decode, Eq, PartialEq, TypeInfo, MaxEncodedLen)]
+struct CollectionDetails {
```
chungquantin, line 33: Required because `CollectionDetails`'s fields in `pallet-nfts` are not public.

pallets/api/src/mock.rs
Daanvdplas: I see, the third and last are probably for the `OffchainPublic` and `OffchainSignature`

pallets/api/src/nonfungibles/tests.rs
```diff
@@ -0,0 +1,1259 @@
...
+		assert_eq!(to_balance_after_transfer, to_balance_before_transfer + 1);
+		assert_eq!(from_balance_after_transfer, from_balance_before_transfer - 1);
+		System::assert_last_event(
+			Event::Transfer { collection, item, from: Some(owner), to: Some(dest) }.into(),
+		);
+	});
+}
+
+#[test]
+fn mint_works() {
```
Daanvdplas, line 99: Based on the dispatchable order I would prefer to have it followed by `approve` tests.

pallets/api/src/nonfungibles/tests.rs
```diff
@@ -0,0 +1,1259 @@
...
+				attribute,
+				value
+			),
+			NftsError::NoPermission
+		);
+	});
+}
+
+#[test]
+fn create_works() {
```
Daanvdplas, line 570: Missing test for the CollectionId read but I don't know how easy it is to add... If not we can leave it

Daanvdplas: No test that the error for nfts::create work

pallets/api/src/nonfungibles/tests.rs
```diff
@@ -0,0 +1,1259 @@
...
+				AttributeNamespace::CollectionOwner,
+				attribute
+			),
+			Some(value.into())
+		);
+	});
+}
+
+#[test]
+fn clear_attribute_works() {
```
Daanvdplas, line 471: Missing check whether error works

pallets/api/src/nonfungibles/tests.rs
```diff
@@ -0,0 +1,1259 @@
...
+		let attribute = BoundedVec::truncate_from("some attribute".as_bytes().to_vec());
+		let collection = COLLECTION;
+		let delegate = BOB;
+		let item = ITEM;
+		let owner = ALICE;
+		let value = BoundedVec::truncate_from("some value".as_bytes().to_vec());
+
+		nfts::create_collection_mint(owner, owner, item);
+		// Successfully approve delegate to set attributes.
+		assert_ok!(NonFungibles::approve_item_attributes(
```
Daanvdplas, line 518: Missing check that error works

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+		pub fn mint(
+			origin: OriginFor<T>,
+			to: AccountIdOf<T>,
+			collection: CollectionIdOf<T>,
+			item: ItemIdOf<T>,
+			witness: MintWitness<ItemIdOf<T>, ItemPriceOf<T>>,
+		) -> DispatchResult {
+			let owner = ensure_signed(origin.clone())?;
+			NftsOf::<T>::mint(origin, collection, item, T::Lookup::unlookup(to), Some(witness))?;
+			Self::deposit_event(Event::Transfer { collection, item, from: None, to: Some(owner) });
```
Daanvdplas, line 431: Dont understand why this is not `to`:
```suggestion
			Self::deposit_event(Event::Transfer { collection, item, from: None, to });
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+		#[pallet::call_index(20)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::burn())]
+		pub fn burn(
+			origin: OriginFor<T>,
+			collection: CollectionIdOf<T>,
+			item: ItemIdOf<T>,
+		) -> DispatchResult {
+			let owner = ensure_signed(origin.clone())?;
+			NftsOf::<T>::burn(origin, collection, item)?;
+			Self::deposit_event(Event::Transfer { collection, item, from: Some(owner), to: None });
```
Daanvdplas, line 449: Can `from` not be someone else than the signer? It must be the owner of the item however this requires a storage read.

runtime/devnet/src/config/api/mod.rs
```diff
@@ -130,8 +146,8 @@ pub struct Filter<T>(PhantomData<T>);
 
 impl<T: frame_system::Config<RuntimeCall = RuntimeCall>> Contains<RuntimeCall> for Filter<T> {
 	fn contains(c: &RuntimeCall) -> bool {
-		use fungibles::Call::*;
-		T::BaseCallFilter::contains(c) &&
+		let contain_fungibles: bool = {
```
Daanvdplas, line 149:
```suggestion
		fungibles: bool = {
```

runtime/devnet/src/config/assets.rs
```diff
@@ -41,7 +42,19 @@ parameter_types! {
...
+pub struct KeyLimit<const N: u32>;
+
+impl<const N: u32> Get<u32> for KeyLimit<N> {
+	fn get() -> u32 {
+		N
+	}
+}
+
+pub(crate) type TrustBackedNftsInstance = pallet_nfts::Instance1;
+pub type TrustBackedNftsCall = pallet_nfts::Call<Runtime, TrustBackedNftsInstance>;
```
Daanvdplas, line 56: docs

runtime/devnet/src/config/assets.rs
```diff
@@ -62,7 +75,7 @@ impl pallet_nfts::Config for Runtime {
 	// TODO: source from primitives
 	type ItemId = ItemId;
 	// TODO: source from primitives
-	type KeyLimit = ConstU32<64>;
+	type KeyLimit = KeyLimit<64>;
```
Daanvdplas, line 78: What is the value add here?

runtime/devnet/src/lib.rs
```diff
@@ -652,13 +652,16 @@ mod runtime {
...
+	#[runtime::pallet_index(151)]
+	pub type NonFungibles = nonfungibles::Pallet<Runtime>;
 }
 
 #[cfg(feature = "runtime-benchmarks")]
 mod benches {
 	frame_benchmarking::define_benchmarks!(
 		[frame_system, SystemBench::<Runtime>]
 		[fungibles, Fungibles]
+		[nonfungibles, NonFungibles]
```
Daanvdplas, line 664: alpabetical

Daanvdplas (commented):
Solid PR! Most of the functionality is correct so well done! 

Most of the comments are related to differences with the spec which caused confusion as deviating from it means that it needs to be reviewed again which takes time. In addition, having the order of functions in the same order makes the review process streamlined, same counts for the order of tests with the code (i.e. dispatachables order, etc). 

Moreover, the additional functions that have been added create review overload (because it is already a very big PR) because this requires much more than reviewing the logic, tests and benchmarks. This is also why I haven't reviewed those functions.

runtime/devnet/src/config/api/mod.rs
chungquantin: `fungibles` is the crate name

runtime/devnet/src/config/assets.rs
chungquantin: Same for `TrustedBackedAssetsCall` above so I suggest resolving those in this issue https://github.com/r0gue-io/pop-node/issues/446 to keep it consistent.

runtime/devnet/src/lib.rs
chungquantin: f < n < p. And these are not ordered alphabetically, see `cumulus_` below.

pallets/api/src/nonfungibles/mod.rs
chungquantin: I personally see that field is incorrect. Maybe it could be changed in `pallet-nfts`?

pallets/api/src/nonfungibles/tests.rs
chungquantin: Ok, I'll create a PR. Will keep this for now then rebase when the PR is merged.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+type AttributeNamespaceOf<T> = AttributeNamespace<AccountIdOf<T>>;
+type CollectionConfigOf<T> = CollectionConfig<ItemPriceOf<T>, BlockNumberFor<T>, CollectionIdOf<T>>;
+type CollectionDetailsOf<T> = CollectionDetails<AccountIdOf<T>, BalanceOf<T>>;
+type CollectionIdOf<T> =
+	<NftsOf<T> as Inspect<<T as frame_system::Config>::AccountId>>::CollectionId;
+type ItemIdOf<T> = <NftsOf<T> as Inspect<<T as frame_system::Config>::AccountId>>::ItemId;
+type ItemPriceOf<T> = BalanceOf<T>;
+type NftsErrorOf<T> = pallet_nfts::Error<T, NftsInstanceOf<T>>;
+type NftsWeightInfoOf<T> = <T as pallet_nfts::Config<NftsInstanceOf<T>>>::WeightInfo;
+pub(super) type BalanceOf<T> =
```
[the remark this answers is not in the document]
chungquantin: These are used in `tests.rs` so yes. But I just noticed that `pub(crate)` is enough. Do you agree?

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+
+#[frame_support::pallet]
+pub mod pallet {
+	use frame_support::{
+		dispatch::{DispatchResult, DispatchResultWithPostInfo},
+		pallet_prelude::*,
+		traits::Incrementable,
+	};
+	use frame_system::pallet_prelude::*;
+	use pallet_nfts::{CancelAttributesApprovalWitness, DestroyWitness, MintWitness};
```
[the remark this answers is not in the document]
chungquantin: You're right. Resolved by removing those.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+
+	/// State reads for the non-fungibles API with required input.
+	#[derive(Encode, Decode, Debug, MaxEncodedLen)]
+	#[cfg_attr(feature = "std", derive(PartialEq, Clone))]
+	#[repr(u8)]
+	#[allow(clippy::unnecessary_cast)]
+	pub enum Read<T: Config> {
+		/// Total item supply of a specified `collection`.
+		#[codec(index = 0)]
+		TotalSupply(CollectionIdOf<T>),
```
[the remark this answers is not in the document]
chungquantin: Maybe when I wrote the spec we did not consider the order of functions, [please see PSP34](https://github.com/Cardinal-Cryptography/PSP34/blob/main/traits.rs), `total_supply` is before `balance_of`

runtime/devnet/src/config/assets.rs
chungquantin: The maximum length of an attribute key.

pallets/api/src/nonfungibles/benchmarking.rs
```diff
@@ -0,0 +1,222 @@
...
+)]
+mod benchmarks {
+	use super::*;
+
+	// Parameter:
+	// - 'a': whether `approved` is true or false.
+	// - 'i': whether `item` is provided.
+	#[benchmark]
+	fn approve(a: Linear<0, 1>, i: Linear<0, 1>) -> Result<(), BenchmarkError> {
+		let collection_id = CollectionIdOf::<T>::zero();
```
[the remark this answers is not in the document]
chungquantin: There is a weight function `collection()` which is used for weighting `Read::Collection` query. Hence, can't name it as `collection`

pallets/api/src/nonfungibles/tests.rs
chungquantin: Resolved by reordering all tests based on dispatchable order

pallets/api/src/nonfungibles/tests.rs
```diff
@@ -0,0 +1,1259 @@
...
+		owner: AccountId,
+		mint_to: AccountId,
+		item: ItemIdOf<Test>,
+		operator: AccountId,
+	) {
+		let (collection, item) = create_collection_mint(owner, mint_to, item);
+		assert_ok!(Nfts::approve_transfer(signed(owner), collection, item, operator, None));
+	}
+
+	pub(super) fn create_collection_mint(
```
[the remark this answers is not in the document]
chungquantin: Resolved by changing it to `create_collection_and_mint`

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+		#[pallet::weight(NftsWeightInfoOf::<T>::transfer() + T::DbWeight::get().reads(1))]
+		pub fn transfer(
+			origin: OriginFor<T>,
+			collection: CollectionIdOf<T>,
+			item: ItemIdOf<T>,
+			to: AccountIdOf<T>,
+		) -> DispatchResultWithPostInfo {
+			let owner = NftsOf::<T>::owner(collection, item)
+				.ok_or(NftsErrorOf::<T>::UnknownItem)
+				.map_err(|e| e.with_weight(T::DbWeight::get().reads(1)))?;
```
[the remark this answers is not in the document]
chungquantin: Resolved

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+			Self::deposit_event(Event::Transfer {
+				collection,
+				item,
+				from: Some(owner),
+				to: Some(to),
+			});
+			Ok(().into())
+		}
+
+		/// Either approve or cancel approval for an `operator` to perform transfers of a specific
```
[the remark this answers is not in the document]
chungquantin: Resolved.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,621 @@
...
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Transfers the collection item from the `origin` to account `to`.
+		///
+		/// # Parameters
+		/// - `collection` - The collection of the item to be transferred.
+		/// - `item` - The item to transfer.
+		/// - `to` - The recipient account.
```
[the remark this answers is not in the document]
chungquantin: Resolved.

pallets/api/src/nonfungibles/mod.rs
chungquantin: These are implemented for the pallet while the spec mainly stick to the PSP34 standard. And we did agree that we should not limit the capabilities of the pallet. Hence, as we already provided this in the fork pallet nfts, implement it for the nonfungibles pallet will enrich the pallet features

pallets/api/src/nonfungibles/tests.rs
chungquantin: Resolved in this PR: https://github.com/r0gue-io/pop-node/pull/449

pallets/api/src/nonfungibles/mod.rs
Daanvdplas: Up to you, we can also create an issue and adapt later as it delays finishing this PR

pallets/api/src/nonfungibles/mod.rs
Daanvdplas: It does however make the PR bigger, more to review, etc.

pallets/api/src/nonfungibles/mod.rs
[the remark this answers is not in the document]
Daanvdplas: I was able to remove them and it still worked (if I not made a mistake)

pallets/api/src/nonfungibles/mod.rs
Daanvdplas: So if it is the `ForceOrigin` the event would be incorrect

pallets/api/src/nonfungibles/impls.rs
```diff
@@ -1,72 +1,50 @@
...
 
-use super::{pallet::*, AccountIdOf, CollectionIdOf, ItemIdOf, NftsOf, WeightInfo};
+use super::{pallet::*, AccountIdLookupOf, CollectionIdOf, ItemIdOf, NftsOf};
 
 impl<T: Config> Pallet<T> {
-	/// Approves the transfer of a specific item or all collection items owned by the `owner` to an
-	/// `operator`.
+	// Approves the transfer of a specific item or all collection items owned by the `owner` to an
+	// `operator`.
 	///
```
Daanvdplas: You forgot this one

Daanvdplas: You forgot this one

pallets/api/src/nonfungibles/impls.rs
```diff
@@ -1,72 +1,50 @@
...
-	/// # Parameters
-	/// - `owner` - The owner of the specified collection item(s).
-	/// - `collection` - The identifier of the collection.
-	/// - `maybe_item` - The optional item of the collection that the operator has an approval to
-	///   transfer. If not provided, an approval to transfer all `owner`'s collection items will be
-	///   cancelled.
-	/// - `operator` - The account that had permission to transfer a specified collection item or
-	///   all collection items owned by the `owner`.
+	// # Parameters
```
Daanvdplas, line 31: Same

pallets/api/src/nonfungibles/impls.rs
```diff
@@ -1,72 +1,50 @@
...
+	// `operator`.
 	///
-	/// # Parameters
-	/// - `owner` - The owner of the specified collection item(s).
-	/// - `collection` - The identifier of the collection.
-	/// - `maybe_item` - The optional item of the collection to be approved for delegated transfer.
-	///   If `None`, the approval applies to all `owner`'s collection items.
-	/// - `operator` - The account to delegate permission to transfer a specified collection item or
-	///   all collection items owned by the `owner`.
+	// # Parameters
```
Daanvdplas, line 10: Please remove, only adds more review work

pallets/api/src/nonfungibles/impls.rs
```diff
@@ -1,72 +1,50 @@
-use frame_support::dispatch::{DispatchResultWithPostInfo, WithPostDispatchInfo};
```
Daanvdplas: I find this separate file unnecessary now.

Daanvdplas (commented):
Just skimmed over and found some improvements in the `impl.rs`. I will do a proper review monday.

pallets/api/src/nonfungibles/mod.rs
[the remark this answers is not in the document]
chungquantin: You're right. But I don't understand why `pub` is not required when tests.rs use those types.

pallets/api/src/nonfungibles/mod.rs
chungquantin: Resolved.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,664 @@
...
+				item,
+				namespace,
+				key.clone(),
+				value.clone(),
+			)?;
+			Self::deposit_event(Event::AttributeSet {
+				collection,
+				item,
+				key: key.to_vec(),
+				data: value.to_vec(),
```
peterwht, line 313: We are currently cloning each of these twice. Once when you do `key.clone()`, and once when you do `key.to_vec()`

I recommend defining two vars at the top of the fn: `key_vec = key.to_vec();` and `data_vec = data.to_vec()`. This then allows you to do only one clone.

peterwht (changes requested):
Nice to see how clean this and simple the mapping is to the pallet.

I haven't reviewed everything, but there are several comments I left.

One change is that we shouldn't need to charge extra for a DB read in the weight macro. Reason being is that the item is pulled into the memory overlay and pallet-nfts already charges for that read.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,664 @@
...
+		/// - `collection` - The collection to destroy.
+		/// - `witness` - Information on the items minted in the `collection`. This must be
+		/// correct.
+		#[pallet::call_index(13)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::destroy(
+    		witness.item_metadatas,
+    		witness.item_configs,
+    		witness.attributes,
+		))]
+		pub fn destroy(
```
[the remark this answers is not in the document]
peterwht: see here for more context: https://github.com/r0gue-io/pop-node/pull/445#issuecomment-2613368602

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,664 @@
...
+		/// - `config` - The configuration of the collection.
+		#[pallet::call_index(12)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::create() + T::DbWeight::get().reads(1))]
+		pub fn create(
+			origin: OriginFor<T>,
+			admin: AccountIdOf<T>,
+			config: CollectionConfigOf<T>,
+		) -> DispatchResultWithPostInfo {
+			let creator = ensure_signed(origin.clone())?;
+			let id = NextCollectionIdOf::<T>::get()
```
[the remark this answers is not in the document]
chungquantin: `CollectionId` does not have `Decrementable` trait which is not possible to substract one from the ID.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,664 @@
...
+		///
+		/// Origin must be either the item's owner or an account approved by the owner to
+		/// transfer the item.
+		///
+		/// # Parameters
+		/// - `collection` - The collection identifier.
+		/// - `item` - The item.
+		/// - `to` - The recipient account.
+		#[pallet::call_index(3)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::transfer() + T::DbWeight::get().reads(1))]
```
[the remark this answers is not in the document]
chungquantin: Resolved by removing the double charged weight

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,664 @@
...
+			Ok(())
+		}
+
+		/// Destroy a single collection item.
+		///
+		/// # Parameters
+		/// - `collection` - The collection identifier.
+		/// - `item` - The item to burn.
+		#[pallet::call_index(8)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::burn() + T::DbWeight::get().reads(1))]
```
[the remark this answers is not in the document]
chungquantin: Resolved by removing the double charged weight

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,664 @@
...
+			Ok(().into())
+		}
+
+		/// Issue a new collection of non-fungible items.
+		///
+		/// # Parameters
+		/// - `admin` - The admin of this collection.
+		/// - `config` - The configuration of the collection.
+		#[pallet::call_index(12)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::create() + T::DbWeight::get().reads(1))]
```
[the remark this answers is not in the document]
chungquantin: Resolved by removing the double charged weight

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,664 @@
...
+		///   from that collection needs to be provided within the witness data object. If the mint
+		///   price is set, then it should be additionally confirmed in the `witness`.
+		#[pallet::call_index(7)]
+		#[pallet::weight(NftsWeightInfoOf::<T>::mint())]
+		pub fn mint(
+			origin: OriginFor<T>,
+			to: AccountIdOf<T>,
+			collection: CollectionIdOf<T>,
+			item: ItemIdOf<T>,
+			witness: MintWitness<ItemIdOf<T>, ItemPriceOf<T>>,
```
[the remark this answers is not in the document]
chungquantin: Great suggestion!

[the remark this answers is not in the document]
chungquantin: Resolved by refactoring based on your suggestion. I keep the field name as `price` instead of `maybe_price` to remain the consistency with other dispatchables' parameters.

pallets/api/src/nonfungibles/mod.rs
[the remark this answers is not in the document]
chungquantin: I don't see much value by adding those two methods. I prefer having only the `destroy_with_witness` and then handle the query of destroy witness on the contract side, inside the pop-api method implementation. In that way, it still cost one extra read but not charging the max-weight for post-dispatch.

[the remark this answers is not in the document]
Daanvdplas: This is out of scope and should be tackled in a different PR

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+			Self::deposit_event(Event::Transfer {
+				collection,
+				item,
+				from: Some(owner),
+				to: Some(to),
+			});
+			Ok(().into())
+		}
+
+		/// Approves operator to transfer item(s) from the `owner`'s account.
```
Daanvdplas, line 162: No `owner` parameter:
```suggestion
		/// Approves operator to transfer item(s) from the owner's account.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+		/// - `item` - Optional item. `None` means all collection items owned by the `owner`.
+		/// - `operator` - The account that is allowed to transfer the item.
+		/// - `approved` - Whether the operator is given or removed the right to transfer the
+		///   item(s).
+		#[pallet::call_index(4)]
+		#[pallet::weight(WeightOf::<T>::approve(*approved as u32, item.is_some() as u32))]
+		pub fn approve(
+			origin: OriginFor<T>,
+			collection: CollectionIdOf<T>,
+			item: Option<ItemIdOf<T>>,
```
Daanvdplas, line 175: The order of the fields in the events and functions are not in the same order as defined in the spec which I think means it is not inline with the standard. Whether this is a problem is questionable because by adding the `collection` field / parameter it is already not standard compliant. However the order of these fields and parameters will be different in the contract so it might be a bit odd. Curious to hear your opinion. Btw, I agree that this order you have now makes the most sense.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+				Self::do_cancel_approval(origin, collection, item, operator_lookup)
+			};
+			Self::deposit_event(Event::Approval { collection, item, operator, owner, approved });
+			result
+		}
+
+		/// Mints an item to the specified recipient account.
+		///
+		/// # Parameters
+		/// - `collection` - The collection of the item to mint.
```
Daanvdplas, line 193: paramters are not in the same order as the function definition

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+			item: ItemIdOf<T>,
+			price: Option<ItemPriceOf<T>>,
+		) -> DispatchResult {
+			NftsOf::<T>::mint(
+				origin,
+				collection,
+				item,
+				T::Lookup::unlookup(to.clone()),
+				Some(MintWitness { owned_item: Some(item), mint_price: price }),
+			)?;
```
Daanvdplas, line 212: The `owned_item` in the witness data is not the same item as is minted

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+			Self::deposit_event(Event::Transfer { collection, item, from: None, to: Some(to) });
+			Ok(())
+		}
+
+		/// Destroy a single collection item.
```
Daanvdplas, line 217: As per spec
```suggestion
		/// Destroys the specified item. Clearing the corresponding approvals.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+			item: ItemIdOf<T>,
+		) -> DispatchResultWithPostInfo {
+			let owner =
+				NftsOf::<T>::owner(collection, item).ok_or(NftsErrorOf::<T>::UnknownItem)?;
+			NftsOf::<T>::burn(origin, collection, item)?;
+			Self::deposit_event(Event::Transfer { collection, item, from: Some(owner), to: None });
+			Ok(().into())
+		}
+
+		/// Issue a new collection of non-fungible items.
```
Daanvdplas, line 236: As per spec:
```suggestion
		/// Creates an NFT collection. 
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+		///
+		/// # Parameters
+		/// - `admin` - The admin of this collection.
```
Daanvdplas, line 239:
```suggestion
		/// - `admin` - The admin account of the collection.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+		/// - `config` - The configuration of the collection.
```
Daanvdplas, line 240: As per spec:
```suggestion
		/// - `config` - Settings and config to be set for the new collection.

```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+			let creator = ensure_signed(origin.clone())?;
+			let id = NextCollectionIdOf::<T>::get()
+				.or(T::CollectionId::initial_value())
+				.ok_or(NftsErrorOf::<T>::UnknownCollection)?;
+			NftsOf::<T>::create(origin, T::Lookup::unlookup(admin.clone()), config)?;
+			Self::deposit_event(Event::Created { id, creator, admin });
+			Ok(())
+		}
+
+		/// Destroy a collection of items.
```
Daanvdplas, line 257:
```suggestion
		/// Destroy an NFT collection.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+			NftsOf::<T>::destroy(origin, collection, witness)
+		}
+
+		/// Set an attribute for a collection or item.
+		///
+		/// # Parameters
+		/// - `collection` - The collection identifier.
+		/// - `item` - Optional item whose attribute to set. If not provided, the `collection`'s
+		///   attribute is set.
+		/// - `namespace` - Attribute's namespace.
```
Daanvdplas, line 283:
```suggestion
		/// - `namespace` - The attribute's namespace.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+				data: value_vec,
+			});
+			Ok(())
+		}
+
+		/// Clear an attribute for the collection or item.
+		///
+		/// # Parameters
+		/// - `collection` - The collection identifier.
+		/// - `item` - Optional item whose metadata to clear. If `None`, metadata of the
```
Daanvdplas, line 312:
```suggestion
		/// - `item` - The optional item whose metadata to clear. If `None`, metadata of the
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+		///   `collection` will be cleared.
+		/// - `namespace` - Attribute's namespace.
```
Daanvdplas, line 314:
```suggestion
		/// - `namespace` - The attribute's namespace.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+			witness: DestroyWitness,
+		) -> DispatchResultWithPostInfo {
+			NftsOf::<T>::destroy(origin, collection, witness)
+		}
+
+		/// Set an attribute for a collection or item.
+		///
+		/// # Parameters
+		/// - `collection` - The collection identifier.
+		/// - `item` - Optional item whose attribute to set. If not provided, the `collection`'s
```
Daanvdplas, line 281:
```suggestion
		/// - `item` - The optional item whose attribute to set. If not provided, the `collection`'s
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+			key: BoundedVec<u8, T::KeyLimit>,
+		) -> DispatchResult {
+			NftsOf::<T>::clear_attribute(origin, collection, item, namespace, key)
+		}
+
+		/// Set the metadata for an item.
+		///
+		/// # Parameters
+		/// - `collection` - The collection identifier.
+		/// - `item` - The item whose metadata to set.
```
Daanvdplas, line 332:
```suggestion
		/// - `item` - The item. If `None`, set metadata for the collection.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+		/// - `data` - The general information of this item. Limited in length by `StringLimit`.
```
Daanvdplas, line 333:
```suggestion
		/// - `data` - The metadata. Limited in length by `StringLimit`.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+			data: BoundedVec<u8, T::StringLimit>,
+		) -> DispatchResult {
+			NftsOf::<T>::set_metadata(origin, collection, item, data)
+		}
+
+		/// Clear the metadata for an item.
+		///
+		/// # Parameters
+		/// - `collection` - The collection identifier.
+		/// - `item` - The item whose metadata to clear.
```
Daanvdplas, line 349: As per spec:
```suggestion
		/// Clear the metadata for an item or collection.
                ///
                /// Caller must be the admin of the collection.
                ///
                /// # Parameters
                /// - `collection` - The collection.
                /// - `item` - The item.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+			origin: OriginFor<T>,
+			collection: CollectionIdOf<T>,
+			item: Option<ItemIdOf<T>>,
+			namespace: AttributeNamespaceOf<T>,
+			key: BoundedVec<u8, T::KeyLimit>,
+		) -> DispatchResult {
+			NftsOf::<T>::clear_attribute(origin, collection, item, namespace, key)
+		}
+
+		/// Set the metadata for an item.
```
Daanvdplas, line 328:
```suggestion
		/// Set the metadata for an item.
		///
		/// Caller must be the admin of the collection.
```

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,654 @@
...
+		pub fn clear_metadata(
+			origin: OriginFor<T>,
+			collection: CollectionIdOf<T>,
+			item: ItemIdOf<T>,
+		) -> DispatchResult {
+			NftsOf::<T>::clear_metadata(origin, collection, item)
+		}
+
+		/// Approve item's attributes to be changed by a delegated third-party account.
+		///
```
Daanvdplas, line 361:
```suggestion
		/// 
                /// Caller must be the owner of the item.
                ///
```

Daanvdplas (commented):
I stopped as there were so many deviations from the spec without knowing the reason behind it.

pallets/api/src/nonfungibles/mod.rs
chungquantin: Resolved by syncing with the spec. Following the standard compliant is more important so refactoring the parameter order.

pallets/api/src/nonfungibles/mod.rs
chungquantin: Reverted the changes

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+			origin: OriginFor<T>,
+			collection: CollectionIdOf<T>,
+			witness: DestroyWitness,
+		) -> DispatchResultWithPostInfo {
+			NftsOf::<T>::destroy(origin, collection, witness)
+		}
+
+		/// Set an attribute for a collection or item.
+		///
+		/// Origin must be Signed and must conform to the namespace ruleset:
```
chungquantin, line 278: Copied from pallet-nfts. Not added in the spec.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+		/// - `CollectionOwner` namespace could be modified by the `collection` Admin only;
+		/// - `ItemOwner` namespace could be modified by the `item` owner only. `item` should be set
+		///   in that case;
+		/// - `Account(AccountId)` namespace could be modified only when the `origin` was given a
+		///   permission to do so;
```
Daanvdplas, line 283: Perhaps also good to add. This validates the point made earlier to do a final check with pallet nfts if there is any docs we should add to the dispatchable docs here and forgotten in the spec.
```suggestion
		///   permission to do so;
		///
		/// The funds of `origin` are reserved according to the formula:
		/// `AttributeDepositBase + DepositPerByte * (key.len + value.len)` taking into
		/// account any already reserved funds.
```

pallets/api/src/nonfungibles/weights.rs
Daanvdplas: Have you tried re running it?

pallets/api/src/nonfungibles/weights.rs
```diff
@@ -0,0 +1,267 @@
...
+	/// Proof: `Nfts::CollectionConfigOf` (`max_values`: None, `max_size`: Some(73), added: 2548, mode: `MaxEncodedLen`)
+	/// Storage: `Nfts::AccountBalance` (r:1 w:0)
+	/// Proof: `Nfts::AccountBalance` (`max_values`: None, `max_size`: Some(120), added: 2595, mode: `MaxEncodedLen`)
+	/// Storage: `Nfts::CollectionApprovals` (r:1 w:1)
+	/// Proof: `Nfts::CollectionApprovals` (`max_values`: None, `max_size`: Some(137), added: 2612, mode: `MaxEncodedLen`)
+	/// Storage: `System::Account` (r:1 w:1)
+	/// Proof: `System::Account` (`max_values`: None, `max_size`: Some(128), added: 2603, mode: `MaxEncodedLen`)
+	/// The range of component `a` is `[0, 1]`.
+	/// The range of component `i` is `[0, 1]`.
+	fn approve(a: u32, i: u32, ) -> Weight {
```
Daanvdplas, line 61: These weight results are incorrect. It seems to always add all the reads and writes it has found for all 4 different dispatchables. I think we need to specify the post dispatch weight either way... 

I would first try to rerun the benchmarks if you have done that already.

pallets/api/src/nonfungibles/weights.rs
```diff
@@ -0,0 +1,267 @@
...
+		// Proof Size summary in bytes:
+		//  Measured:  `606`
+		//  Estimated: `68341938741594600 + a * (3168 ±44_956_152_011_428) + i * (2663 ±729_819_634_434_010)`
+		// Minimum execution time: 13_000_000 picoseconds.
+		Weight::from_parts(25_055_102, 68341938741594600)
+			// Standard Error: 195_002
+			.saturating_add(Weight::from_parts(324_489, 0).saturating_mul(a.into()))
+			.saturating_add(T::DbWeight::get().reads(4_u64))
```
Daanvdplas, line 69: Here you can see it always adds all the reads.

runtime/devnet/src/config/assets.rs
```diff
@@ -41,7 +42,19 @@ parameter_types! {
 	pub const NftsMaxDeadlineDuration: BlockNumber = 12 * 30 * DAYS;
 }
 
-impl pallet_nfts::Config for Runtime {
+#[derive(Debug)]
+#[cfg_attr(feature = "std", derive(PartialEq, Clone))]
+pub struct KeyLimit<const N: u32>;
```
Daanvdplas, line 47: Missing docs

Daanvdplas (approved):
PR looking really great! The weight of `fn approve` need to be rerun because right now they are incorrect. It also seems to always take all the reads found for benchmarking all possible cases which is also incorrect.

Other than that everything looks great and happy to approve when the weights are fixed!

EDIT: By accident gave an approve, should have been a comment.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+			Self::deposit_event(Event::Approval { collection, item, operator, owner, approved });
+			result
+		}
+
+		/// Mints an item to the specified recipient account.
+		///
+		/// # Parameters
+		/// - `collection` - The collection.
+		/// - `to` - The recipient account.
+		/// - `item` - An identifier of the new item.
```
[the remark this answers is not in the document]
chungquantin: ok with the suggestion but fyi, pallet-nfts has the same docs for `item`.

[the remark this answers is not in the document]
Daanvdplas: Yeah I saw it in the spec, just an improvement.

pallets/api/src/nonfungibles/tests.rs
```diff
@@ -0,0 +1,1350 @@
...
+			let collection = COLLECTION;
+			let dest = BOB;
+			let item = ITEM;
+			let owner = ALICE;
+
+			// Throw `NftsError::UnknownItem` if no item found.
+			assert_noop!(
+				NonFungibles::transfer(signed(dest), collection, dest, item),
+				NftsError::UnknownItem
+			);
```
peterwht, line 44: not sure we should be testing these errors like this. pallet-nfts does the error checking for us. So, even though we query the storage item at the top of the transfer dispatchable the result would still be same if we didn't return the error (as pallet-nfts would).

I would just recommend that we have one test that makes sure every pallet-nft error variant properly maps to something in the Pop API (this may exist somewhere already)

peterwht (changes requested):
It's definitely getting there!

I think the testing is a little too much though. The tests feel like they are testing pallet-nfts. But pallet-nfts is already tested.

The scope of testing for this API pallet is:
- Does our API properly call into pallet-nfts and have the desired outcome
- Does our API properly handle errors that might be returned from pallet-nfts?
    - To be clear, I am not saying: "are the proper errors returned from pallet-nfts" -- that's not our job. Our job is just to make sure we handle the errors
- Does our Read functionality properly cover pallet-nfts.
- (Anything else I missed?)

To put it concretely, I recommend the following changes to the testing:
- Remove any cases where you check if a specific error is returned from calling a dispatchable -- not this pallet's responsibility
- Remove any base-case asserts. For example: "Alice does not have an NFT", or "the collection does not exist". Once again, those are handled by pallet-nfts. Just check the outcome: "Does Alice now have an NFT after using our API?"
- Make sure that every error returned from pallet-nfts maps properly to an error that our API handles (this test may exist somewhere).
- Make sure all the reads via the API also work, the weights are accessible, etc.

cc @Daanvdplas

pallets/api/src/nonfungibles/tests.rs
chungquantin: `nonfungibles::transfer()` does return this error `UnknownItem` so we need to include it in the test and it is required in this case. When we query the `owner` in the `transfer` dispatchable, it returns `None` and the best flow is to throw the error in the dispatchable. 

However, based on the internal discussion, I get your point here.

Daanvdplas: I disagree, we might have put another error here. Any code that is written by us should be tested.

pallets/api/src/nonfungibles/weights.rs
chungquantin: Tried re-running the benchmarks and this remain the same. One thing I learn is if `i` is hard-coded to `0` or `1`, the generated weight is normal. Quite weird to see the surge of generated weight.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+		/// - `approved` - Whether the operator is given or removed the right to transfer the
+		///   item(s).
+		#[pallet::call_index(4)]
+		#[pallet::weight(WeightOf::<T>::approve(*approved as u32, item.is_some() as u32))]
+		pub fn approve(
+			origin: OriginFor<T>,
+			collection: CollectionIdOf<T>,
+			operator: AccountIdOf<T>,
+			item: Option<ItemIdOf<T>>,
+			approved: bool,
```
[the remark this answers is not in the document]
chungquantin: Resolved by adding the deadline field

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+
+		/// Mints an item to the specified recipient account.
+		///
+		/// # Parameters
+		/// - `collection` - The collection.
+		/// - `to` - The recipient account.
+		/// - `item` - An identifier of the new item.
+		/// - `witness` - When the mint type is `HolderOf(collection_id)`, then the owned item_id
+		///   from that collection needs to be provided within the witness data object. If the mint
+		///   price is set, then it should be additionally confirmed in the `witness`.
```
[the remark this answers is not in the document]
chungquantin: Resolved by adding more docs from `pallet-nfts` to `create`, `destroy`, `clear_attribute`, `set_metadata` and `clear_metadata`. Mainly related to the reserved funds. See latest commit. https://github.com/r0gue-io/pop-node/pull/432/commits/5955fcfc1e617393c6702aded4cb99913dd2ca51

runtime/devnet/src/config/assets.rs
```diff
@@ -44,6 +44,7 @@ parameter_types! {
 
 #[derive(Debug)]
 #[cfg_attr(feature = "std", derive(PartialEq, Clone))]
+/// Public because `pallet_nfts::Config` is a public trait.
```
Daanvdplas, line 47: This is how it will show in the docs if we publish this crate. Not really suitable tmo. Can we explain what this type is instead, you can simply take the docs of pallet nfts' config type you are setting it with.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,688 @@
...
+			/// The collection identifier.
+			id: CollectionIdOf<T>,
+			/// The creator of the collection.
+			creator: AccountIdOf<T>,
+			/// The administrator of the collection.
+			admin: AccountIdOf<T>,
+		},
+	}
+
+	#[pallet::call]
```
Daanvdplas, line 128:
```suggestion
	/// The non-fungibles dispatchables. For more information about a dispatchable refer to `pallet-nfts`.
	#[pallet::call]
```

runtime/devnet/src/config/assets.rs
Daanvdplas:
```suggestion
/// The maximum length of an attribute key.
```

Daanvdplas (approved):
LGTM!

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,703 @@
...
+		/// Returns whether the operator is approved by the owner to withdraw item. If item is not
+		/// provided, it returns whether the operator is approved to withdraw all owner's items for
+		/// the given collection.
+		Allowance(bool),
+		/// Returns the total supply of a collection.
+		TotalSupply(u128),
+		/// Returns the attribute value of item for a given key, if any.
+		GetAttribute(Option<Vec<u8>>),
+		/// Returns the details of a specified collection, if any.
+		Collection(Option<CollectionDetailsOf<T>>),
```
chungquantin, line 618: This query might not be a good idea as it requires `CollectionDetails` type on the contract size which increases the size of the contract and field `owner` and `items` can be queried using existing Reads `OwnerOf` and `TotalSupply`. My suggestion is to replace this with three additional Reads: 

- `CollectionItemConfigs`
- `CollectionItemMetadatas`
- `CollectionAttributes`

peterwht: I think it's good to have as long as we can make it optional on the contract side. Basically, let's see if we can provide developers the option (good devex), while not bloating contract size for those that don't use it.

If you don't think it's useful for a contract to have, let's exclude it and see if we get some feedback to include it.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -0,0 +1,703 @@
...
+		///   item(s).
+		/// - `deadline`: The optional deadline (in block numbers) specifying the time limit for the
+		///   approval, only required if `approved` is true.
+		#[pallet::call_index(3)]
+		#[pallet::weight(
+			NftsWeightInfoOf::<T>::approve_transfer()
+			.max(NftsWeightInfoOf::<T>::approve_collection_transfer())
+			.max(NftsWeightInfoOf::<T>::cancel_approval())
+			.max(NftsWeightInfoOf::<T>::cancel_collection_approval()))
+		]
```
peterwht, line 152: I think this is a good solution for now. But could we create an issue to make this more accurate (as the original code tried to do). Then let's link that issue in a code comment (non-doc).

peterwht (approved):
LGTM! Left a single comment

pallets/api/src/nonfungibles/mod.rs
chungquantin: Issue: https://github.com/r0gue-io/pop-node/issues/463

pallets/api/src/nonfungibles/mod.rs
chungquantin: Resolved by removing the Collection read as I don't see it is needed at the moment and brings more complexity to the implementation of the Pop API library (found an issue with that returned huge CollectionDetails can't be decoded properly in the integration test PR). 

Hence, excluding it now and implement the feature later in the future if there is a request for it.

chungquantin: Issue: https://github.com/r0gue-io/pop-node/issues/464

Daanvdplas (approved):

PR #433, opened 2025-01-16
chore: add release conventions document and refactor CI to run node release separately

Daanvdplas:
This PR does two things (apologies for not separating them, will do next time).
1. Lines out the release conventions. View here: https://github.com/r0gue-io/pop-node/blob/daan/docs-release_conventions/docs/release.md
2. Refactors the CI so that it is clear how to manually run the job to build the release for a node without a runtime build ([works](https://github.com/r0gue-io/pop-node/actions/runs/12810004638)). As well as having the CI pickup automatically when a release title is created with `node` ([works](https://github.com/r0gue-io/pop-node/actions/runs/12816167595)).

[sc-2432]


al3mart (approved):

.github/workflows/release.yml
```diff
@@ -16,11 +16,11 @@ on:
         description: Build node
         default: true
       runtime:
-        description: 'Runtime to build (devnet, testnet, mainnet)'
-        default: ''
+        description: 'Runtime to build (none, devnet, testnet, mainnet)'
```
chungquantin, line 19: curious why `none`?

Daanvdplas: So that in the UI you can choose from the option `none` previously this was not clear to me as it was an empty bar. See here `run workflow` on the top right: https://github.com/r0gue-io/pop-node/actions/workflows/release.yml

chungquantin (approved):

PR #434, opened 2025-01-16
refactor(deps): include pop api integration tests in workspace

peterwht:
The `pop-api` crate is excluded from the workspace as it should function as a standalone crate. However, `pop-api/integration-tests` is linked to the pop runtimes yet it is still excluded from the workspace.

This exclusion results in two main issues:
1. If the pop runtimes are on a different polkadot-sdk version, the pop-api/integration-tests will fail
2. Running `cargo test` does not include the pop-api/integration-tests resulting in missing necessary changes when doing runtime upgrades.

This PR includes `pop-api/integration-tests` solving these two issues. The crate was renamed to `pop-api-integration-tests` to avoid conflicts with `integration-tests`.


pop-api/integration-tests/contracts/create_token_in_constructor/Cargo.toml
```diff
@@ -3,6 +3,8 @@ edition = "2021"
 name = "create_token_in_constructor"
 version = "0.1.0"
 
+[workspace]
```
peterwht, line 6: Excludes contract from workspace.

pop-api/integration-tests/contracts/fungibles/Cargo.toml
```diff
@@ -3,6 +3,8 @@ edition = "2021"
 name = "fungibles"
 version = "0.1.0"
 
+[workspace]
```
peterwht, line 6: Excludes contract from workspace.

pop-api/integration-tests/Cargo.toml
```diff
@@ -1,36 +1,39 @@
...
-frame-support = { version = "36.0.0", default-features = false }
-frame-system = { version = "36.1.0", default-features = false }
-log = "0.4.22"
-pallet-assets = { version = "37.0.0", default-features = false }
-pallet-balances = { version = "37.0.0", default-features = false }
-pallet-contracts = { version = "35.0.0", default-features = false }
+env_logger.workspace = true
+log.workspace = true
+
+codec = { workspace = true, default-features = false, features = [
```
peterwht, line 14: renamed from scale -> codec as the workspace deps has the name as "codec"

.github/workflows/ci.yml
```diff
@@ -131,28 +131,6 @@ jobs:
         working-directory: pop-api
         run: cargo test --release --locked --all-features
 
-  api-integration-tests-devnet:
```
peterwht, line 134: no longer necessary as it's included in the regular `cargo test`

pop-api/integration-tests/Cargo.toml
Daanvdplas, line 14: As for the rest of the dependencies, because in the root Cargo.toml file they are declared already with default features false, you don't have to do it again here.
```suggestion
codec = { workspace = true, features = [ "derive" ] }
```

.github/workflows/ci.yml
```diff
@@ -120,38 +120,32 @@ jobs:
...
       - uses: "./.github/actions/init"
 
-      - name: Run integration tests
-        working-directory: pop-api/integration-tests
-        run: cargo test --release --locked --features devnet
+      - name: Run pop api integration tests
+        run: cargo test -p pop-api-integration-tests --release --locked --features "${{ matrix.features }}"
 
-  api-integration-tests-testnet:
+  api-tests:
```
Daanvdplas, line 139: Conflicts with your comment here: https://github.com/r0gue-io/pop-node/pull/434/files#r1919344165

.github/workflows/ci.yml
```diff
@@ -101,7 +101,7 @@ jobs:
       - uses: "./.github/actions/init"
 
       - name: Run tests
-        run: cargo test --release --locked --workspace --exclude integration-tests --features=runtime-benchmarks
+        run: cargo test --release --locked --workspace --exclude integration-tests,pop-api-integration-tests --features=runtime-benchmarks
```
Daanvdplas, line 104:
```suggestion
        run: cargo test --release --locked --workspace --exclude integration-tests --exclude pop-api-integration-tests --features=runtime-benchmarks
```

Daanvdplas (approved):
Looking great, thanks for doing this! Left a few comments to improve but will approve as I don't want to block here.

al3mart (commented):
Did review and I haven't seen anything I'd change.

Happy to approve once the above comments are resolved and CI fixed :)

.github/workflows/ci.yml
evilrobot-01: Prefer the comma separated version personally, less characters.

evilrobot-01:
> This exclusion results in two main issues:

A third is disk usage of build artifacts.

.github/workflows/ci.yml
```diff
@@ -120,38 +120,32 @@ jobs:
...
-        working-directory: pop-api
-        run: cargo test --release --locked --all-features
-
-  api-integration-tests-devnet:
+  pop-api-integration-tests:
+    needs: lint
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        features:
```
evilrobot-01, line 128: Nice clean up. Might be worth adding a comment as to why these have to be run separately. Will just help anyone needing to improve in the future by being explicit.

.github/workflows/ci.yml
```diff
@@ -120,38 +120,32 @@ jobs:
...
-      - uses: actions/checkout@v4
-
-      - uses: "./.github/actions/init"
-
-      - name: Run tests
-        working-directory: pop-api
-        run: cargo test --release --locked --all-features
-
-  api-integration-tests-devnet:
+  pop-api-integration-tests:
```
evilrobot-01, line 123: Minor: not sure why needs to be prefixed with pop when its obvious by the repo/context. Crate name makes sense though, so feel free to ignore.

.github/workflows/ci.yml
```diff
@@ -120,38 +120,32 @@ jobs:
...
+          - "testnet"
     steps:
       - uses: actions/checkout@v4
 
       - uses: "./.github/actions/init"
 
-      - name: Run integration tests
-        working-directory: pop-api/integration-tests
-        run: cargo test --release --locked --features devnet
+      - name: Run pop api integration tests
```
evilrobot-01, line 136:
```suggestion
      - name: Run API integration tests
```
Feel free to ignore.

Cargo.lock
```diff
@@ -157,9 +157,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.7"
+version = "1.0.10"
```
evilrobot-01, line 160: Strange that there are version bumps when only a simple amend to the workspace manifest. Was this intentional?

pop-api/integration-tests/contracts/create_token_in_constructor/Cargo.toml
evilrobot-01: Nice workaround, but might it be better to set in the workspace root via so that anything within contracts is automatically excluded:

```
[workspace]
exclude = ["pop-api/integration-tests/contracts/*"]
```

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-members-and-exclude-fields

Note: I dont know that this works, just seems a more robust way as its defined in a single place and automatically excludes any future contracts added.

evilrobot-01 (commented):
Looks good, only minor comments.

pop-api/integration-tests/contracts/create_token_in_constructor/Cargo.toml
peterwht: adding this to INCLUDE does work: `"pop-api/integration-tests/contracts/*",
`

peterwht: I tried adding to the exclude and the include. Exclude the same problem arises (crate thinks it's in a workspace), while including it results in the build hanging.

peterwht: Nevermind, tried again and it worked by including. Weird.

.github/workflows/ci.yml
peterwht: this comment is outdated. We introduced a new task that does api-integration-tests but only requires one test for devenet & testnet.

.github/workflows/ci.yml
peterwht: unfortunately the comma separated doesn't work and needs the --exclude per package.

Cargo.lock
peterwht: Not intentional . Tried to reset the cargo.lock to main and do a rebuild, but there are still version bumps. Likely due to the pop-api/integration-tests now using later versions. https://github.com/r0gue-io/pop-node/pull/434/commits/4fdfe782a71b3d7ae66e024db26d974555dba0c2

.github/workflows/ci.yml
peterwht: changed in https://github.com/r0gue-io/pop-node/pull/434/commits/b4ba54746511858e051486698eaa26dc07a106f1

.github/workflows/ci.yml
peterwht: https://github.com/r0gue-io/pop-node/pull/434/commits/b4ba54746511858e051486698eaa26dc07a106f1

.github/workflows/ci.yml
peterwht: https://github.com/r0gue-io/pop-node/pull/434/commits/d44ac5618e3b98190c0382fc5efceef7f705c53e

.github/workflows/ci.yml
peterwht: https://github.com/r0gue-io/pop-node/pull/434/commits/b4ba54746511858e051486698eaa26dc07a106f1

al3mart (approved):

PR #438, opened 2025-01-21
refactor(mainnet): utility config

al3mart:
Creates **utility** config module including:
- `pallet_multisig`
- `pallet_preimage`
- `pallet_scheduler`
- `pallet_utility`

Notable configuration items are:

- `pallet_multisig`

```rust
// One storage item; key size is 32 + 32; value is size 4+4+16+32 bytes = 120 bytes.
type DepositBase = deposit(1, 120);
// Additional storage item size of 32 bytes.
type DepositFactor = deposit(0, 32);
type MaxSignatories = 100;
```

---

- `pallet_preimage`

```rust
pub const PreimageHoldReason: RuntimeHoldReason = RuntimeHoldReason::Preimage(pallet_preimage::HoldReason::Preimage);
pub const PreimageBaseDeposit: Balance = deposit(2, 64);
pub const PreimageByteDeposit: Balance = deposit(0, 1);
type ManagerOrigin = EnsureRoot<AccountId>;
```

---

- `pallet_scheduler`

Maximum of 50 calls scheduled per block.
No more than 60% of the block weight can be scheduled per block.
Only `root` can schedule, and only the same origin can cancel scheduled tasks.
```rust
type MaxScheduledPerBlock = ConstU32<50>;
type MaximumWeight = Perbill::from_percent(60) *
		RuntimeBlockWeights::get().max_block;
type OriginPrivilegeCmp = EqualPrivilegeOnly;		
type ScheduleOrigin = EnsureRoot<AccountId>;
```

[sc-2205]

### DIFF
--- a/runtime/mainnet/src/config/utility.rs
+++ b/runtime/mainnet/src/config/utility.rs
@@ -104,4 +104,53 @@ mod tests {
 			TypeId::of::<()>(),
 		);
     }
+
+    #[test]
+    fn preimage_uses_balances_as_currency() {
+        assert_eq!(
+			TypeId::of::<<Runtime as pallet_preimage::Config>::Currency>(),
+			TypeId::of::<Balances>(),
+		);
+    }
+
+    #[test]
+    fn preimage_manage_origin_is_root() {
+        assert_eq!(
+			TypeId::of::<<Runtime as pallet_preimage::Config>::ManagerOrigin>(),
+			TypeId::of::<EnsureRoot<AccountId>>(),
+		);
+    }
+
+    #[test]
+    fn preimage_does_not_use_default_weights() {
+        assert_ne!(
+            TypeId::of::<<Runtime as pallet_preimage::Config>::WeightInfo>(),
+            TypeId::of::<()>()
+        );
+    }
+
+    #[test]
+    fn preimage_hold_reason_uses_linear_price() {
+        assert_eq!(
+            TypeId::of::<<Runtime as pallet_preimage::Config>::Consideration>(),
+            TypeId::of::<
+                HoldConsideration<
+                    AccountId,
+                    Balances,
+                    PreimageHoldReason,
+                    LinearStoragePrice<PreimageBaseDeposit, PreimageByteDeposit, Balance>,
+                >,
+            >()
+        );
+    }
+
+    #[test]
+    fn preimage_base_deposit() {
+        assert_eq!(PreimageBaseDeposit::get(), deposit(2, 64));
+    }
+
+    #[test]
+    fn preimage_byte_deposit() {
+        assert_eq!(PreimageByteDeposit::get(), deposit(0, 1));
+    }
 }
--- a/runtime/mainnet/src/config/utility.rs
+++ b/runtime/mainnet/src/config/utility.rs
@@ -1,0 +1,21 @@
+use frame_support::parameter_types;
+
+use crate::{deposit, Balance, Balances, Runtime, RuntimeCall, RuntimeEvent};
+
+parameter_types! {
+	// One storage item; key size is 32 + 32; value is size 4+4+16+32 bytes = 120 bytes.
+	pub const DepositBase: Balance = deposit(1, 120);
+	// Additional storage item size of 32 bytes.
+	pub const DepositFactor: Balance = deposit(0, 32);
+	pub const MaxSignatories: u32 = 100;
+}
+
+impl pallet_multisig::Config for Runtime {
+	type Currency = Balances;
+	type DepositBase = DepositBase;
+	type DepositFactor = DepositFactor;
+	type MaxSignatories = MaxSignatories;
+	type RuntimeCall = RuntimeCall;
+	type RuntimeEvent = RuntimeEvent;
+	type WeightInfo = pallet_multisig::weights::SubstrateWeight<Runtime>;
+}
--- a/runtime/mainnet/src/config/utility.rs
+++ b/runtime/mainnet/src/config/utility.rs
@@ -1,6 +1,4 @@
-use frame_support::parameter_types;
-
-use crate::{deposit, Balance, Balances, OriginCaller, Runtime, RuntimeCall, RuntimeEvent};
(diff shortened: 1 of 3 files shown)


runtime/mainnet/src/config/utility.rs
```diff
@@ -0,0 +1,247 @@
...
+		RuntimeBlockWeights::get().max_block;
+}
+
+impl pallet_scheduler::Config for Runtime {
+	#[cfg(feature = "runtime-benchmarks")]
+	type MaxScheduledPerBlock = ConstU32<512>;
+	#[cfg(not(feature = "runtime-benchmarks"))]
+	type MaxScheduledPerBlock = ConstU32<50>;
+	type MaximumWeight = MaximumSchedulerWeight;
+	type OriginPrivilegeCmp = EqualPrivilegeOnly;
```
al3mart, line 57: Only root can schedule.

runtime/mainnet/src/config/utility.rs
```diff
@@ -0,0 +1,247 @@
...
+
+use crate::{
+	deposit, parameter_types, AccountId, Balance, Balances, EnsureRoot, HoldConsideration,
+	LinearStoragePrice, OriginCaller, Perbill, Preimage, Runtime, RuntimeBlockWeights, RuntimeCall,
+	RuntimeEvent, RuntimeHoldReason, RuntimeOrigin, Weight,
+};
+
+parameter_types! {
+	// One storage item; key size is 32 + 32; value is size 4+4+16+32 bytes = 120 bytes.
+	pub const DepositBase: Balance = deposit(1, 120);
```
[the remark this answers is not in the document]
peterwht: just to note as well: this is a conservative estimate that makes our multisig more expensive than most in the ecosystem. Not saying it's wrong, just noting.

Have we checked how much DOT is actually reserved in this case?

peterwht (approved):
Also looks good. Have one comment, but going to approve as it's not a blocker!

runtime/mainnet/src/config/utility.rs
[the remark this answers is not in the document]
al3mart: > Have we checked how much DOT is actually reserved in this case?

Not yet. I wanted to push the new value here and have some discussion around it, as you point out we might want to move from that 120 because of reasons.

Will figure the actual amount in DOT. Should give us enough info to decide :+1:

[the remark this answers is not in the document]
Daanvdplas: Would be great if we can verify the 120 bytes by a test.

runtime/mainnet/src/config/utility.rs
```diff
@@ -0,0 +1,247 @@
...
+
+impl pallet_utility::Config for Runtime {
+	type PalletsOrigin = OriginCaller;
+	type RuntimeCall = RuntimeCall;
+	type RuntimeEvent = RuntimeEvent;
+	type WeightInfo = pallet_utility::weights::SubstrateWeight<Runtime>;
+}
+
+#[cfg(test)]
+mod tests {
```
Daanvdplas, line 75: Nitpick but could we put the tests in submodules per pallet, in the order they are implemented in this file.

runtime/mainnet/src/config/utility.rs
```diff
@@ -0,0 +1,247 @@
...
+			TypeId::of::<<Runtime as pallet_multisig::Config>::Currency>(),
+			TypeId::of::<Balances>(),
+		);
+	}
+
+	#[test]
+	fn multisig_call_deposit_has_base_amount() {
+		assert_eq!(
+			<<Runtime as pallet_multisig::Config>::DepositBase as Get<Balance>>::get(),
+			deposit(1, 120)
```
Daanvdplas, line 110: Would be good to see where this 120 comes from by using the KeyLen and the MaxEncodedLen of the value.

Same for the others.

runtime/mainnet/src/config/utility.rs
```diff
@@ -0,0 +1,247 @@
...
+	fn scheduler_call_queue_per_block_is_limited() {
+		assert_eq!(
+			TypeId::of::<<Runtime as pallet_scheduler::Config>::MaxScheduledPerBlock>(),
+			TypeId::of::<ConstU32<50>>(),
+		);
+	}
+
+	#[test]
+	#[cfg(feature = "runtime-benchmarks")]
+	fn scheduler_call_queue_per_block_is_limited() {
```
Daanvdplas, line 146: Can we swap this test with the above to keep them in the order of the config types :):)

runtime/mainnet/src/config/utility.rs
```diff
@@ -0,0 +1,247 @@
...
+	#[test]
+	fn scheduler_does_not_use_default_weights() {
+		assert_ne!(
+			TypeId::of::<<Runtime as pallet_scheduler::Config>::WeightInfo>(),
+			TypeId::of::<()>()
+		);
+	}
+
+	#[test]
+	fn scheduler_uses_preimage_to_look_up_calls() {
```
Daanvdplas, line 192: Can we keep these tests in the order of the config types. Can we also add a test for the PalletsOrigin

runtime/mainnet/src/config/utility.rs
```diff
@@ -0,0 +1,247 @@
...
+	#[test]
+	fn preimage_uses_balances_as_currency() {
+		assert_eq!(
+			TypeId::of::<<Runtime as pallet_preimage::Config>::Currency>(),
+			TypeId::of::<Balances>(),
+		);
+	}
+
+	#[test]
+	fn preimage_manage_origin_is_root() {
```
Daanvdplas, line 208: Can we also enforce the tests in the order of the config types :)

Daanvdplas (commented):
Nicely done! Improvement would be to add tests (where possible) where we verify the calculation done in the comments for e.g. the `DepositBase`.

In addition, I would like us to add submodules for the pallet for tests, as well as writing the tests in the order of the config types. As these types get ordered by the +nightly fmt it is constant and makes reviewing and searching those tests very easy :)

PR #439, opened 2025-01-21
refactor(mainnet): collation config

al3mart:


peterwht (commented):
Left a few nitpicks!

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,279 @@
...
+	pub const Period: u32 = 6 * HOURS;
+}
+
+#[docify::export(aura_config)]
+impl pallet_aura::Config for Runtime {
+	type AllowMultipleBlocksPerSlot = ConstBool<true>;
+	type AuthorityId = AuraId;
+	type DisabledValidators = ();
+	// Number of blocks we can produce per session.
+	type MaxAuthorities = Period;
```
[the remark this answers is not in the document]
al3mart: If configured for such, there could be that many authorities, or more. It wouldn't be a problem.
But with the current block time and session lenght, aura wouldn't be able to rotate the whole authority set before the end of the session.

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,279 @@
...
+		// Disabled to start until sufficient distribution/value to allow candidates to provide
+		// candidacy bond
+		assert_eq!(
+			<<Runtime as pallet_collator_selection::Config>::MaxCandidates as Get<u32>>::get(),
+			0
+		);
+	}
+
+	#[test]
+	fn collator_selection_requires_at_least_five_collators() {
```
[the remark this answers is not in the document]
al3mart: thanks for spotting this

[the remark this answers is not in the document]
al3mart: [e716881](https://github.com/r0gue-io/pop-node/pull/439/commits/e716881363b874560eba213eff55b719d90dbb7e)

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,279 @@
...
+	#[test]
+	fn collator_selection_uses_native_asset() {
+		assert_eq!(
+			TypeId::of::<<Runtime as pallet_collator_selection::Config>::Currency>(),
+			TypeId::of::<Balances>(),
+		);
+	}
+
+	#[test]
+	fn collator_selection_update_origin_limited_to_root() {
```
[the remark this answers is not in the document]
al3mart: [e716881](https://github.com/r0gue-io/pop-node/pull/439/commits/e716881363b874560eba213eff55b719d90dbb7e)

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,279 @@
...
+	#[test]
+	fn authorship_notes_block_author_via_collator_selection() {
+		assert_eq!(
+			TypeId::of::<<Runtime as pallet_authorship::Config>::EventHandler>(),
+			TypeId::of::<(CollatorSelection,)>(),
+		);
+	}
+
+	#[test]
+	fn collator_selection_uses_native_asset() {
```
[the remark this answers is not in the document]
al3mart: [e716881](https://github.com/r0gue-io/pop-node/pull/439/commits/e716881363b874560eba213eff55b719d90dbb7e)

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+/// We allow root to execute privileged collator selection operations.
+pub type CollatorSelectionUpdateOrigin = EnsureRoot<AccountId>;
+
+impl pallet_collator_selection::Config for Runtime {
+	type Currency = Balances;
+	// should be a multiple of session or things will get inconsistent
+	type KickThreshold = Period;
+	type MaxCandidates = ConstU32<0>;
+	type MaxInvulnerables = ConstU32<20>;
+	type MinEligibleCollators = ConstU32<1>;
```
peterwht, line 41: I know we talked about this a bit in DM. For this current mainnet we are preparing, we want the ideal amount for every config as we will have to do one runtime upgrade to get here.

I personally think we should have a minimum of 3 collators (what you originally had). I personally think we should start with 3 collators only as well.

peterwht (commented):
Ready to approve after just one comment is addressed.

Regarding the min collators, do you think it's better to align to our current genesis state or should we have our desired runtime with ideal parameters from the start?

runtime/mainnet/src/config/collation.rs
al3mart: I do think 3 would be perfect too!

Let's get the changes into the genesis state first, and then alight this config.

al3mart: I have reverted this change

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+	type Currency = Balances;
+	// should be a multiple of session or things will get inconsistent
+	type KickThreshold = Period;
+	type MaxCandidates = ConstU32<0>;
+	type MaxInvulnerables = ConstU32<20>;
+	type MinEligibleCollators = ConstU32<3>;
+	type PotId = PotId;
+	type RuntimeEvent = RuntimeEvent;
+	type UpdateOrigin = CollatorSelectionUpdateOrigin;
+	type ValidatorId = <Self as frame_system::Config>::AccountId;
```
peterwht, line 45:
```suggestion
	type ValidatorId = AccountId;
```

any reason not to do this? It compiles.

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+
+impl pallet_session::Config for Runtime {
+	type Keys = SessionKeys;
+	type NextSessionRotation = pallet_session::PeriodicSessions<Period, Offset>;
+	type RuntimeEvent = RuntimeEvent;
+	// Essentially just Aura, but let's be pedantic.
+	type SessionHandler = <SessionKeys as sp_runtime::traits::OpaqueKeys>::KeyTypeIdProviders;
+	type SessionManager = CollatorSelection;
+	type ShouldEndSession = pallet_session::PeriodicSessions<Period, Offset>;
+	type ValidatorId = <Self as frame_system::Config>::AccountId;
```
peterwht, line 61:
```suggestion
	type ValidatorId = AccountId;
```

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+
+	#[test]
+	fn session_keys_provided_by_aura() {
+		assert_eq!(
+			TypeId::of::<<Runtime as pallet_session::Config>::Keys>(),
+			TypeId::of::<SessionKeys>(),
+		);
+		// Session keys implementation uses aura-defined authority identifier type
+		SessionKeys {
+			aura: <Runtime as pallet_aura::Config>::AuthorityId::from_slice(&[0u8; 32]).unwrap(),
```
peterwht, line 233: what's the purpose of this? No assert or usage. Just a compile test?

peterwht (commented):
Just left a few nitpitcks to tidy up! Super close though

runtime/mainnet/src/config/collation.rs
al3mart: Good one! Thanks

runtime/mainnet/src/config/collation.rs
al3mart: Awesome

al3mart: [be40c5f](https://github.com/r0gue-io/pop-node/pull/439/commits/be40c5fe8a7efc1bde595971b08a6a61e78ab9ad)

runtime/mainnet/src/config/collation.rs
al3mart: [be40c5f](https://github.com/r0gue-io/pop-node/pull/439/commits/be40c5fe8a7efc1bde595971b08a6a61e78ab9ad)

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+	#[test]
+	fn aura_disabled_validators_not_used() {
+		assert_eq!(
+			TypeId::of::<<Runtime as pallet_aura::Config>::DisabledValidators>(),
+			TypeId::of::<()>(),
+		);
+	}
+
+	#[test]
+	fn aura_max_authorities() {
```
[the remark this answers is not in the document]
al3mart: [1b3823a](https://github.com/r0gue-io/pop-node/pull/439/commits/1b3823a15fe0d5609d1e39fe6e91966f213032b5)

runtime/mainnet/src/config/collation.rs
al3mart: We configure `pallet_session` with the following config for `Keys`:
```rust
type Keys = SessionKeys;
```

We define `SessionKeys` as part of the opaque types in our runtime as:
```rust
impl_opaque_keys! {
	pub struct SessionKeys {
		pub aura: Aura,
	}
}
```

That line initializes the struct so that it can be used in the assert in that test.
Happy to change the test if there's a better way to do that :)

peterwht: the restructure makes it more clear, thank you!

peterwht (approved):

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+			TypeId::of::<<Runtime as pallet_aura::Config>::AllowMultipleBlocksPerSlot>(),
+			TypeId::of::<ConstBool<true>>(),
+		);
+	}
+
+	#[test]
+	fn aura_uses_sr25519_for_authority_id() {
+		assert_eq!(
+			TypeId::of::<<Runtime as pallet_aura::Config>::AuthorityId>(),
+			TypeId::of::<sp_consensus_aura::sr25519::AuthorityId>(),
```
Daanvdplas, line 104: 👌

runtime/mainnet/src/config/collation.rs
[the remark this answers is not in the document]
Daanvdplas: Could we add a comment above `Period` for this :)

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+	type SessionManager = CollatorSelection;
+	type ShouldEndSession = pallet_session::PeriodicSessions<Period, Offset>;
+	type ValidatorId = AccountId;
+	// we don't have stash and controller, thus we don't need the convert as well.
+	type ValidatorIdOf = pallet_collator_selection::IdentityCollator;
+	type WeightInfo = pallet_session::weights::SubstrateWeight<Runtime>;
+}
+
+#[cfg(test)]
+mod tests {
```
Daanvdplas, line 68: Nitpick; I would love to have the tests in separate modules per pallet. Not separate files, just modules within this module (like done in the treasury)

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+	pub const PotId: PalletId = PalletId(*b"PotStake");
+	pub const Offset: u32 = 0;
+}
+
+/// We allow root to execute privileged collator selection operations.
+pub type CollatorSelectionUpdateOrigin = EnsureRoot<AccountId>;
+
+impl pallet_collator_selection::Config for Runtime {
+	type Currency = Balances;
+	// should be a multiple of session or things will get inconsistent
```
Daanvdplas, line 37:
```suggestion
	// Should be a multiple of session or things will get inconsistent.
```

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+	type KickThreshold = Period;
+	type MaxCandidates = ConstU32<0>;
+	type MaxInvulnerables = ConstU32<20>;
+	type MinEligibleCollators = ConstU32<3>;
```
Daanvdplas, line 41: Description of the PR: "Max of 20 invulnerable authorities, of which at least 1 needs to be elected." 

Should this thus not be 1?

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+		);
+	}
+
+	#[test]
+	fn collator_selection_distributes_block_rewards_via_pot() {
+		// Context: block author receives rewards from 'pot', less ED. A keyless account
+		// 'pot' is generated from the `PotId` value configured for the pallet.
+		assert_eq!(
+			TypeId::of::<<Runtime as pallet_collator_selection::Config>::PotId>(),
+			TypeId::of::<PotId>(),
```
Daanvdplas, line 176:
```suggestion
			PalletId(*b"PotStake"),
```

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+}
+
+impl cumulus_pallet_aura_ext::Config for Runtime {}
+
+impl pallet_session::Config for Runtime {
+	type Keys = SessionKeys;
+	type NextSessionRotation = pallet_session::PeriodicSessions<Period, Offset>;
+	type RuntimeEvent = RuntimeEvent;
+	// Essentially just Aura, but let's be pedantic.
+	type SessionHandler = <SessionKeys as sp_runtime::traits::OpaqueKeys>::KeyTypeIdProviders;
```
Daanvdplas, line 58: Can you explain this?

runtime/mainnet/src/config/collation.rs
```diff
@@ -0,0 +1,288 @@
...
+	type SessionManager = CollatorSelection;
+	type ShouldEndSession = pallet_session::PeriodicSessions<Period, Offset>;
+	type ValidatorId = AccountId;
+	// we don't have stash and controller, thus we don't need the convert as well.
```
Daanvdplas, line 62: I don't understand this:
```suggestion
	// We don't have a stash and controller account, thus we don't need the convert as well.
```

Daanvdplas (commented):
Neat PR, left a few comments for clarification.

runtime/mainnet/src/config/collation.rs
al3mart: The type needs to be configured with an implementation that provides (from the pallet doc):

> A conversion from account ID to validator ID.
> Its cost must be at most one storage read.

Which was useful when authorities were using controller accounts. This is no longer the case so an identity function is used.

I'll remove the comment as it might be more confusing than helpful. The config itself can provide enough clarification.

runtime/mainnet/src/config/collation.rs
al3mart: [This macro](https://github.com/r0gue-io/pop-node/blob/975bf974c97ddafae884d63dd541169e59c20d3c/runtime/mainnet/src/lib.rs#L230-L234) will resolve `KeyTypeIdProviders` to use Aura.

runtime/mainnet/src/config/collation.rs
al3mart: `collator_selection_pot_account_is_valid` test already locks the configuration of PotId.

I don't think this is necessary.

runtime/mainnet/src/config/collation.rs
al3mart: [ee0f3d2](https://github.com/r0gue-io/pop-node/pull/439/commits/ee0f3d2c8a1347f03f8e9a4750a72e06d8baede0)

runtime/mainnet/src/config/collation.rs
Daanvdplas: Now you are explaining the previous comment 😅

runtime/mainnet/src/config/collation.rs
Daanvdplas: The hash* (or whatever it is), but alright

runtime/mainnet/src/config/collation.rs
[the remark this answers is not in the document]
Daanvdplas: Sorry but the current comment if very messy, I would prefer something along the lines of what you wrote [here](https://github.com/r0gue-io/pop-node/pull/439/files#r1929331824) as it is much clearer and provides much more useful information of why `6* HOURS`

al3mart:
> Only the comment above `Period` and then it is good to go!

How does this not solve the request ?
https://github.com/r0gue-io/pop-node/pull/439/commits/16e5a46f85aaa5587bf8421c44c0647a81179521

al3mart:
rebased to [al3mart/refactor-mainnet-config](https://github.com/r0gue-io/pop-node/tree/al3mart/refactor-mainnet-config) after merging proxy config into the base branch.

Daanvdplas (approved):
LGTM!

PR #441, opened 2025-01-21
refactor(mainnet): pallet_proxy config & tests

al3mart:
Extends **proxy** config module for the pallet:
- `pallet_proxy`

Notable configuration items are:

- `pallet_proxy`

Updated deposits for mainnet runtime:

```rust
// One storage item; key size 32 + 8.
pub const ProxyDepositBase: Balance = deposit(1, 40);
// Additional storage item size of AccountId 32 bytes + ProxyType 1 byte + BlockNum 4 bytes.
pub const ProxyDepositFactor: Balance = deposit(0, 37);
// One storage item; key size 32, value size 16.
pub const AnnouncementDepositBase: Balance = deposit(1, 48);
// Additional storage item 32 bytes AccountId + 32 bytes Hash + 4 bytes BlockNum.
pub const AnnouncementDepositFactor: Balance = deposit(0, 68);
```

```rust
type MaxPending = 32;
type MaxProxies = 32;
```

[sc-2205]

### DIFF
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -116,6 +116,12 @@ pub mod proxy {
 		NonTransfer,
 		/// Proxy with the ability to reject time-delay proxy announcements.
 		CancelProxy,
+		/// Assets proxy. Can execute any call from `assets`, **including asset transfers**.
+		Assets,
+		/// Owner proxy. Can execute calls related to asset ownership.
+		AssetOwner,
+		/// Asset manager. Can execute calls related to asset management.
+		AssetManager,
 		/// Collator selection proxy. Can execute calls related to collator selection mechanism.
 		Collator,
 	}
@@ -131,6 +137,8 @@ pub mod proxy {
 				(x, y) if x == y => true,
 				(ProxyType::Any, _) => true,
 				(_, ProxyType::Any) => false,
+				(ProxyType::Assets, ProxyType::AssetOwner) => true,
+				(ProxyType::Assets, ProxyType::AssetManager) => true,
 				(ProxyType::NonTransfer, ProxyType::Collator) => true,
 				_ => false,
 			}
--- a/runtime/common/src/lib.rs
+++ b/runtime/common/src/lib.rs
@@ -116,12 +116,6 @@ pub mod proxy {
 		NonTransfer,
 		/// Proxy with the ability to reject time-delay proxy announcements.
 		CancelProxy,
-		/// Assets proxy. Can execute any call from `assets`, **including asset transfers**.
-		Assets,
-		/// Owner proxy. Can execute calls related to asset ownership.
-		AssetOwner,
-		/// Asset manager. Can execute calls related to asset management.
-		AssetManager,
 		/// Collator selection proxy. Can execute calls related to collator selection mechanism.
 		Collator,
 	}
@@ -137,8 +131,6 @@ pub mod proxy {
 				(x, y) if x == y => true,
 				(ProxyType::Any, _) => true,
 				(_, ProxyType::Any) => false,
-				(ProxyType::Assets, ProxyType::AssetOwner) => true,
-				(ProxyType::Assets, ProxyType::AssetManager) => true,
 				(ProxyType::NonTransfer, ProxyType::Collator) => true,
 				_ => false,
 			}
--- a/runtime/mainnet/src/config/proxy.rs
+++ b/runtime/mainnet/src/config/proxy.rs
@@ -31,8 +31,8 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
 }
 
 parameter_types! {
-	// One storage item; key size 32, value size 16.
-	pub const ProxyDepositBase: Balance = deposit(1, 48);
+	// One storage item; key size 32 + hash size 8.
+	pub const ProxyDepositBase: Balance = deposit(1, 40);
 	// Additional storage item size of AccountId 32 bytes + ProxyType 1 byte + BlockNum 4 bytes.
 	pub const ProxyDepositFactor: Balance = deposit(0, 37);
 	// One storage item; key size 32, value size 16.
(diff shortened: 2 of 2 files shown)


runtime/mainnet/src/config/proxy.rs
```diff
@@ -1,11 +1,9 @@
...
-};
-use sp_runtime::traits::BlakeTwo256;
+use pop_runtime_common::proxy::{MaxPending, MaxProxies, ProxyType};
 
-use crate::{Balances, Runtime, RuntimeCall, RuntimeEvent};
+use crate::{
+	deposit, parameter_types, Balance, Balances, BlakeTwo256, Runtime, RuntimeCall, RuntimeEvent,
+};
 
 impl InstanceFilter<RuntimeCall> for ProxyType {
```
peterwht, line 8: this is the closest I can leave a comment.

For `ProxyType::Assets`, we don't have the assets pallet but note how in AH it uses a RuntimeCall::Assets

https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L494

Anything we need to do about this?

runtime/mainnet/src/config/proxy.rs
```diff
@@ -41,6 +39,17 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
 	}
 }
 
+parameter_types! {
+	// One storage item; key size 32, value size 16.
+	pub const ProxyDepositBase: Balance = deposit(1, 48);
```
peterwht, line 44: we have 6 times the amount of bytes Polkadot has. Any reason not to just align to the Polkadot relay chain on this: https://github.com/polkadot-fellows/runtimes/blob/main/relay/polkadot/src/lib.rs#L977

peterwht (commented):
Only concern is around the ProxyDepositBase and ProxyDepositFactor.

1. Wasn't able to calculate the value size myself (so not sure where the 16 comes from)
2. Our deposits are several times more expensive than the polkadot relay.

runtime/mainnet/src/config/proxy.rs
al3mart: Certainly. Double checking after your comment I'm also seeing that I did the following miscalculation:

the number we have for `ProxyDepositBase` was the result of adding `AccountId` + `Balance`, but I realize that we should be accounting for 8 bytes, which is the correct length of the key of the map.

I also added the bytes for `Balance` because the deposited amount is always stored. Although, as it is the size corresponding to the deposit itself, I could see why it is not being included on Polkadot's configuration.

`ProxyDepositFactor` reasoning in the comment above its declaration, what's the concern ?

runtime/mainnet/src/config/proxy.rs
al3mart: The calls for the assets pallets could be extended to the type later on as the pallets are included in the runtime.

Although I'm happy to remove these types for now as they don't make much sense until the pallets are there. Even if it is just a naming detail.

al3mart: [d3bfc6c](https://github.com/r0gue-io/pop-node/pull/441/commits/d3bfc6c748aa46e487980649bfaaa9455ffc70b0)

runtime/common/src/lib.rs
```diff
@@ -116,12 +116,6 @@ pub mod proxy {
 		NonTransfer,
 		/// Proxy with the ability to reject time-delay proxy announcements.
 		CancelProxy,
-		/// Assets proxy. Can execute any call from `assets`, **including asset transfers**.
-		Assets,
-		/// Owner proxy. Can execute calls related to asset ownership.
-		AssetOwner,
-		/// Asset manager. Can execute calls related to asset management.
-		AssetManager,
```
al3mart, line 124: No proxies of these types exist on the running testnet. Nor collator which will change it's index within the enum.

runtime/mainnet/src/config/proxy.rs
al3mart: [d4c936e](https://github.com/r0gue-io/pop-node/pull/441/commits/d4c936e5bca5659e6ed1324b4156ab6f82e1866e)

As discussed off GH this change updates ProxyDepositBase from  `deposit(1, 48)` to `deposit(1, 40)`.

peterwht (approved):
Nice! Thanks for resolving the deposit issue. Makes much more sense now.

runtime/mainnet/src/config/proxy.rs
```diff
@@ -55,3 +99,116 @@ impl pallet_proxy::Config for Runtime {
...
+		}
+		assert!(ProxyType::is_superset(&ProxyType::NonTransfer, &ProxyType::Collator));
+		assert!(!ProxyType::is_superset(&ProxyType::Collator, &ProxyType::NonTransfer));
+	}
+
+	#[test]
+	fn proxy_has_announcement_deposit_base() {
+		assert_eq!(
+			<<Runtime as pallet_proxy::Config>::AnnouncementDepositBase as Get<Balance>>::get(),
+			deposit(1, 48),
```
Daanvdplas, line 146: Would be good if we can back up this `48` value with the actual key size and value size.

Daanvdplas (commented):
Missing the explanation why are we not using the values from `pop-common` anymore. Other than that, looking good!

runtime/mainnet/src/config/proxy.rs
```diff
@@ -8,6 +8,10 @@ use crate::{
 };
 
 /// The type used to represent the kinds of proxying allowed.
+// Mainnet will use this definition of ProxyType instead of the ones in
+// `pop-common` crates until `pallet-assets` is integrated in the runtime.
+// `ProxyType` in `pop-common` include Assets specific proxies which won't
+// make much sense in this runtime.
```
Daanvdplas: Are we not adding pallet assets to mainnet runtime?

Daanvdplas (approved):
Nice! Thanks for the updated tests, very good!

runtime/mainnet/src/config/proxy.rs
al3mart: What's the benefit if smart contracts cannot access it yet ?

al3mart: Resolved via DMs

PR #443, opened 2025-01-21
refactor(mainnet): add governance & pallet_sudo tests

al3mart:
Creates **governance** config module including:
- `pallet_sudo`

[sc-2205]


peterwht (approved):
Looks good! To document here, we discussed that Sudo should stay in the runtime for at least one runtime upgrade to ensure the Council works. Once that is ensured, we can, via an extrinsic, kill the sudo.

PR #444, opened 2025-01-21
refactor: reorder `Read` impl in fungibles pallet

chungquantin:
Quick PR to refactor the order of `Read` implementation in fungibles pallet to keep the structure of pallet consistent with others.

### DIFF
--- a/pallets/api/src/fungibles/tests.rs
+++ b/pallets/api/src/fungibles/tests.rs
@@ -24,56 +24,6 @@ type AssetsWeightInfo = AssetsWeightInfoOf<Test>;
 type Event = crate::fungibles::Event<Test>;
 type WeightInfo = <Test as Config>::WeightInfo;
 
-mod encoding_read_result {
-	use super::*;
-
-	#[test]
-	fn total_supply() {
-		let total_supply = 1_000_000 * UNIT;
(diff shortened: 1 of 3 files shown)


Daanvdplas (commented):
Could you also refactor the tests so that the dispatchables are first, then the read functions etc.

pallets/api/src/fungibles/tests.rs
```diff
@@ -543,18 +493,6 @@ fn token_exists_works() {
 	});
 }
 
-fn signed(account: AccountId) -> RuntimeOrigin {
```
chungquantin, line 546: These functions are shared with nonfungibles pallet as well. Hence, I think refactor it in this PR as well would reduce noises in the nonfungibles pallet PR.

pallets/api/src/fungibles/tests.rs
```diff
@@ -823,4 +635,180 @@ mod ensure_codec_indexes {
 			);
 		})
 	}
+
+	#[test]
+	fn ensure_read_variant_indexes() {
```
chungquantin, line 640: `dispatchable` first then `read`

Daanvdplas (approved):

peterwht (approved):

PR #445, opened 2025-01-21
refactor: remove destroy witness

chungquantin:
Remove the witness data required as parameter from `destroy()` dispatchable function of nonfungibles pallet.
- Pros: Improve developer experience.
- Cons: Pre-dispatch weight charged up to `u32::MAX`.


peterwht:
Unfortunately, I think the devex of removing the witness (forcing high fees) does not outweigh the benefits of not requiring the witness.

A few things, `destroy` is likely something that is infrequently called. Additionally, I feel a better solution will be here: https://github.com/r0gue-io/pop-node/pull/432#discussion_r1929134391. However, to take that idea further, we can allow the smart contract to provide an optional `maybe_witness`. That way the choice is up to the SC developer.

PR #447, opened 2025-01-22
feat(mainnet): add XCM APIs, config tests, refactor mod

peterwht:
Add various improvements to the mainnet runtime, including:
- More config tests:  https://github.com/r0gue-io/pop-node/commit/5fd13cb70e76014eaa55e3c4510e79ebe567398c
- refactor to apis module: https://github.com/r0gue-io/pop-node/commit/3f86cd5a926a28448777c9aec061759f91d16d24
- add XCM runtime apis: https://github.com/r0gue-io/pop-node/commit/a6880e429507f16c6b6a196696a282bc7d5b7720

Recommend reviewing individual commits to more easily see the changes to the APIs.

supersedes https://github.com/r0gue-io/pop-node/pull/426

Credit to @evilrobot-01 as well as some of this code is based on prior work from him.

### DIFF
--- a/runtime/mainnet/src/lib.rs
+++ b/runtime/mainnet/src/lib.rs
@@ -16,6 +16,8 @@ use alloc::{borrow::Cow, vec::Vec};
 use config::xcm::{RelayLocation, XcmOriginToTransactDispatchOrigin};
 use cumulus_pallet_parachain_system::RelayNumberMonotonicallyIncreases;
 use cumulus_primitives_core::{AggregateMessageOrigin, ParaId};
+use cumulus_primitives_storage_weight_reclaim::StorageWeightReclaim;
+use frame_metadata_hash_extension::CheckMetadataHash;
 use frame_support::{
 	derive_impl,
 	dispatch::DispatchClass,
@@ -31,9 +33,11 @@ use frame_support::{
 };
 use frame_system::{
 	limits::{BlockLength, BlockWeights},
-	EnsureRoot,
+	CheckGenesis, CheckMortality, CheckNonZeroSender, CheckNonce, CheckSpecVersion, CheckTxVersion,
+	CheckWeight, EnsureRoot,
 };
 use pallet_balances::Call as BalancesCall;
+use pallet_transaction_payment::ChargeTransactionPayment;
 use pallet_xcm::{EnsureXcm, IsVoiceOfBody};
 use parachains_common::message_queue::{NarrowOriginToSibling, ParaIdToSibling};
 use polkadot_runtime_common::xcm_sender::NoPriceForMessageDelivery;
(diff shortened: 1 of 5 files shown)


runtime/mainnet/src/lib.rs
```diff
@@ -87,16 +92,16 @@ pub type BlockId = generic::BlockId<Block>;
...
-	frame_system::CheckSpecVersion<Runtime>,
-	frame_system::CheckTxVersion<Runtime>,
-	frame_system::CheckGenesis<Runtime>,
-	frame_system::CheckMortality<Runtime>,
-	frame_system::CheckNonce<Runtime>,
-	frame_system::CheckWeight<Runtime>,
-	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
-	cumulus_primitives_storage_weight_reclaim::StorageWeightReclaim<Runtime>,
-	frame_metadata_hash_extension::CheckMetadataHash<Runtime>,
+	CheckNonZeroSender<Runtime>,
```
peterwht, line 95: added imports to simplify unit test

evilrobot-01:
Seems to be missing test coverage on the APIs as per the coverage results. They can be awkward to implement, but the benefit is avoiding regressions longer term.

An example from a recent uplift of the SDK showed that the new unstable metadata was being emitted with 2412. Wouldn't have known that without the tests. Whilst this may not have any real effects, knowing that to be the case could be useful should tooling issues around metadata arise.

No view on whether this should be included now or added later, just mentioning it.

peterwht:
> No view on whether this should be included now or added later, just mentioning it.

My preference is to make it a stretch goal as a separate PR. I personally am not sure if I will be able to prioritize adding those tests in this sprint.

Daanvdplas (approved):
LGTM!

Currently it is a bit hard to figure out what is and what isn't tested in the `lib.rs`. One idea, which I have asked in the treasury pallet PR, is to add the tests for the types in a config of a pallet in the same order as the order of the types. These types get alphabetically ordered by the formatting so it would be consistent. This then makes it a lot easier to review (if we want all these types tested) and maintain. Just an idea, no need to start enforcing it now but I think doing some thinking in how structure these tests now that we are preparing the mainnet runtime could be valuable

runtime/mainnet/src/lib.rs
```diff
@@ -77,6 +72,12 @@ pub type AccountId = <<Signature as Verify>::Signer as IdentifyAccount>::Account
 /// The address format for describing accounts.
 pub type Address = MultiAddress<AccountId, ()>;
 
+/// Record of an event happening.
+pub type EventRecord = frame_system::EventRecord<
```
al3mart, line 76: It seems this type is not being used.

al3mart (changes requested):
It looks like `EventRecord` type is not being used in mainnet. Let's remove it if possible.

Aside from that everything is looking good :+1:

runtime/mainnet/src/lib.rs
peterwht: https://github.com/r0gue-io/pop-node/pull/447/commits/c23cf626382dd98b66e27fbe5c921099fb6ae1c5

al3mart (approved):

PR #456, opened 2025-01-31
fix(xcm): remove waived locations

al3mart:
XCM fees apply universally and are not waived for any specific location in mainnet runtime.

[sc-1353]
[sc-2205]

### DIFF
--- a/runtime/mainnet/src/config/xcm.rs
+++ b/runtime/mainnet/src/config/xcm.rs
@@ -340,19 +340,6 @@ mod tests {
 		use super::*;
 
 		#[test]
-		fn no_locations_are_waived() {
-			assert!(!<<XcmConfig as xcm_executor::Config>::FeeManager>::is_waived(
-				Some(&Location::new(1, [Parachain(1000)])),
-				FeeReason::TransferReserveAsset
-			));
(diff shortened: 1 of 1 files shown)


Daanvdplas (approved):
LGTM!

runtime/mainnet/src/config/xcm.rs
```diff
@@ -170,7 +163,7 @@ impl xcm_executor::Config for XcmConfig {
 	type Barrier = Barrier;
 	type CallDispatcher = RuntimeCall;
 	type FeeManager = XcmFeeManagerFromComponents<
-		WaivedLocations,
+		(),
```
peterwht, line 166: Can we add a simple comment saying: "No locations have waived fees".

peterwht (approved):
Approving. Requested one comment to be added.

runtime/mainnet/src/config/xcm.rs
al3mart: [65e07a7](https://github.com/r0gue-io/pop-node/pull/456/commits/65e07a783b350e30884bec9eebdf73cab744b9d5)

PR #459, opened 2025-01-31
refactor(xcm): Use `NativeAssetFrom<T>` as `XcmReserveTransferFilter`

al3mart:
Implements `Contains<(Location, Vec<Asset>)>` for `NativeAssetFrom<T>` such that only assets with id equals to `AssetId(Location::parent())` are whitelisted.

This `Contains` implementation only focuses on the asset, not the location. The definition of reserves already handles whether the relay asset is being transacted with the correct reserve location.

### DIFF
--- a/runtime/mainnet/src/config/xcm.rs
+++ b/runtime/mainnet/src/config/xcm.rs
@@ -4,7 +4,7 @@ use cumulus_primitives_core::{AggregateMessageOrigin, ParaId};
 use frame_support::{
 	parameter_types,
 	traits::{
-		tokens::imbalance::ResolveTo, ConstU32, ContainsPair, Everything, Get, Nothing,
+		tokens::imbalance::ResolveTo, ConstU32, Contains, ContainsPair, Everything, Get, Nothing,
 		TransformOrigin,
 	},
 	weights::Weight,
@@ -19,7 +19,7 @@ use parachains_common::{
 };
 use polkadot_parachain_primitives::primitives::Sibling;
 use polkadot_runtime_common::xcm_sender::NoPriceForMessageDelivery;
-use sp_runtime::traits::AccountIdConversion;
+use sp_runtime::{traits::AccountIdConversion, Vec};
 use xcm::latest::prelude::*;
 use xcm_builder::{
 	AccountId32Aliases, AllowKnownQueryResponses, AllowSubscriptionsFrom,
@@ -143,6 +143,16 @@ impl<T: Get<Location>> ContainsPair<Asset, Location> for NativeAssetFrom<T> {
 	}
 }
 
+impl<T: Get<Location>> Contains<(Location, Vec<Asset>)> for NativeAssetFrom<T> {
+	fn contains(t: &(Location, Vec<Asset>)) -> bool {
+		t.1.iter().all(|a| {
+			t.0 == T::get() &&
+				matches!(a, Asset { id: AssetId(asset_loc), fun: Fungible(_a) }
+				if *asset_loc == Location::from(Parent))
+		})
+	}
+}
+
 /// Combinations of (Asset, Location) pairs which we trust as reserves.
 pub type TrustedReserves = NativeAssetFrom<AssetHub>;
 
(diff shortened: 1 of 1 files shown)


runtime/mainnet/src/config/xcm.rs
```diff
@@ -862,8 +882,61 @@ mod tests {
 		fn reserve_transfer_filter_only_allows_dot_from_ah() {
 			assert_eq!(
 				TypeId::of::<<Runtime as pallet_xcm::Config>::XcmReserveTransferFilter>(),
-				TypeId::of::<Everything>(),
+				TypeId::of::<NativeAssetFrom<AssetHub>>(),
 			);
+
+			let location_assets = [
```
peterwht, line 888: 🔥

peterwht (approved):
great testing!

runtime/mainnet/src/config/xcm.rs
```diff
@@ -232,7 +242,7 @@ impl pallet_xcm::Config for Runtime {
 	type WeightInfo = pallet_xcm::TestWeightInfo;
 	type XcmExecuteFilter = Everything;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
-	type XcmReserveTransferFilter = Everything;
+	type XcmReserveTransferFilter = NativeAssetFrom<AssetHub>;
```
Daanvdplas, line 245: This will have to refactored once we will include non native assets but oke for now.

runtime/mainnet/src/config/xcm.rs
```diff
@@ -862,8 +882,59 @@ mod tests {
...
+
+			let asset_is_parent = [
+				(
+					AssetHub::get(),
+					vec![Asset { id: AssetId(Location::from(Parent)), fun: Fungible(0) }],
+				),
+				(
+					Location::parent(),
+					vec![Asset { id: AssetId(Location::from(Parent)), fun: Fungible(0) }],
+				),
```
Daanvdplas, line 931: Should this not supposed to work?

runtime/mainnet/src/config/xcm.rs
```diff
@@ -862,8 +882,59 @@ mod tests {
...
+					Location::parent(),
+					vec![
+						Asset { id: AssetId(Location::parent()), fun: Fungible(0) },
+						Asset { id: AssetId(AssetHub::get()), fun: Fungible(0) },
+						Asset {
+							id: AssetId(Location::new(
+								1,
+								[Parachain(1000), PalletInstance(50), GeneralIndex(1984)],
+							)),
+							fun: Fungible(0),
```
Daanvdplas, line 899: Can we clarify these cases with a comment. This should fail because we are trying to transfer assets that are not the relay asset?

runtime/mainnet/src/config/xcm.rs
```diff
@@ -862,8 +882,59 @@ mod tests {
 		fn reserve_transfer_filter_only_allows_dot_from_ah() {
```
Daanvdplas, line 882: `reserve_transfer_filter_only_allows_relay_asset_from_ah`

To be consistent with the other tests.

runtime/mainnet/src/config/xcm.rs
Daanvdplas: Or split the tests and have the tiltle explain it for itself.

Daanvdplas (commented):
Left some comments that need addressing

runtime/mainnet/src/config/xcm.rs
al3mart: impl checks that the assetid of every asset in the list is the parent asset.
Reserves impl then takes care that the concrete asset is transacted with the correct location.

As explained in the description of this PR.

I believe this test doesn't deviate from that behaviour.

runtime/mainnet/src/config/xcm.rs
```diff
@@ -143,6 +143,16 @@ impl<T: Get<Location>> ContainsPair<Asset, Location> for NativeAssetFrom<T> {
 	}
 }
 
+impl<T: Get<Location>> Contains<(Location, Vec<Asset>)> for NativeAssetFrom<T> {
+	fn contains(t: &(Location, Vec<Asset>)) -> bool {
+		t.1.iter().all(|a| {
+			// Asset id matches parent location.
+			matches!(a, Asset { id: AssetId(asset_loc), fun: Fungible(_a) }
+			if *asset_loc == Location::parent())
```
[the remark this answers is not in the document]
al3mart: I don't think that would be possible for this purpose. But maybe I'm just not seeing it.

Could you provide some further explanation ?

runtime/mainnet/src/config/xcm.rs
Daanvdplas: I would remove it from the test as it only creates confusion. You can add a comment that the case of another location should not be possible due to the `IsReserve` implementation that only allows AH as reserve.

PR #460, opened 2025-01-31
refactor: weight approve

Daanvdplas:
Adds my latest reviews to merge this PR

### DIFF
--- a/pallets/api/src/nonfungibles/mod.rs
+++ b/pallets/api/src/nonfungibles/mod.rs
@@ -7,7 +7,6 @@ extern crate alloc;
 use frame_support::{
 	dispatch::WithPostDispatchInfo,
 	traits::{nonfungibles_v2::Inspect, Currency},
-	weights::Weight,
 };
 use frame_system::pallet_prelude::BlockNumberFor;
 pub use pallet::*;
@@ -24,6 +23,7 @@ pub(crate) use weights::WeightInfo;
 mod benchmarking;
 #[cfg(test)]
 mod tests;
+/// Weights for non-fungibles dispatchables.
 pub mod weights;
 
 type AccountBalanceOf<T> = pallet_nfts::AccountBalance<T, NftsInstanceOf<T>>;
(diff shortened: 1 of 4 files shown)


pallets/api/src/nonfungibles/mod.rs
```diff
@@ -173,7 +173,12 @@ pub mod pallet {
 		/// - `deadline`: The optional deadline (in block numbers) specifying the time limit for the
 		///   approval, only required if `approved` is true.
 		#[pallet::call_index(4)]
-		#[pallet::weight(Weight::from_parts(200000000, 4326))]
+		#[pallet::weight(
+			NftsWeightInfoOf::<T>::approve_transfer()
+			.max(NftsWeightInfoOf::<T>::approve_collection_transfer())
+			.max(NftsWeightInfoOf::<T>::cancel_approval())
+			.max(NftsWeightInfoOf::<T>::cancel_collection_approval()))
+		]
```
Daanvdplas, line 181: We initially had a benchmark function but for some reason the weight would always return a very big value and the reads and writes were not separated from each other. In other words, the weights of all the above dispatchables was summed in stead of specifically the weight that was used for each case ([ref](https://github.com/r0gue-io/pop-node/pull/432#discussion_r1930697415)). We couldn't find the issue and have chosen to take the following route. This returns the max proof size and ref time found in the 4 dispatachables that could be called in this dispatchable. The post dispatch returns the actual weight per the dispatchable that is called.

pallets/api/src/nonfungibles/tests.rs
```diff
@@ -101,18 +101,6 @@ mod transfer {
 mod approve {
 	use super::*;
 
-	#[test]
-	fn ensure_origin_works() {
```
Daanvdplas, line 105: No origin check in the dispatachable.

pallets/api/src/nonfungibles/tests.rs
```diff
@@ -1365,14 +1353,3 @@ mod encoding_read_result {
 		assert_eq!(ReadResult::ItemMetadata::<Test>(data.clone()).encode(), data.encode());
 	}
 }
-
-#[test]
-fn ensure_approve_weight() {
```
Daanvdplas, line 1370: Initially I thought it would be bad to call each weight function and get the max proof size and ref time (the implementation suggested in this PR) but if we write a test we would either have to use a hard coded value or have to create a separate function that returns the hard coded value. All just not pretty, prefer to have what we have now.

PR #461, opened 2025-02-03
feat: pop-api nonfungibles implementation

chungquantin:
Add nonfungible feature to the Pop API library. Crate feature will be enabled on the smart contract side with `nonfungibles`. Size of the nonfungible contract (#462): 
```
Original wasm size: 43.8K, Optimized: 15.4K
```
Compare with size of the fungible contract (integration-test contract)
```
Original wasm size: 41.2K, Optimized: 14.2K
```

[sc-1243]


pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -0,0 +1,198 @@
...
+use crate::{macros::impl_codec_bitflags, primitives::AccountId};
+
+pub type ItemId = u32;
+pub type CollectionId = u32;
+type Balance = u32;
+
+/// Attribute namespaces for non-fungible tokens.
+#[derive(Debug, PartialEq, Eq)]
+#[ink::scale_derive(Encode, Decode, TypeInfo)]
+pub enum AttributeNamespace {
```
chungquantin, line 15: Types are ordered alphabetically, not following the order in the spec.

pop-api/src/primitives.rs
```diff
@@ -3,4 +3,5 @@ pub use pop_primitives::*;
 
 // Public due to integration tests crate.
 pub type AccountId = <DefaultEnvironment as Environment>::AccountId;
-pub(crate) type Balance = <DefaultEnvironment as Environment>::Balance;
+pub type Balance = <DefaultEnvironment as Environment>::Balance;
```
Daanvdplas, line 6: Docs

Is this required for integration tests? If so, could we see if we can add the `std` feature

pop-api/src/v0/nonfungibles/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+		.input::<(CollectionId, u32)>()
+		.output::<Result<()>, true>()
+		.handle_error_code::<StatusCode>()
+		.call(&(collection, limit))
+}
+
+/// Mints an item to the specified address.
+///
+/// # Parameters
+/// - `collection` - The collection.
```
Daanvdplas, line 388: Not inline with the actual parameters, please check all :)

pop-api/src/v0/nonfungibles/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+/// - `to` - The recipient account.
+/// - `item` - The ID for the item.
+#[inline]
+pub fn mint(
+	to: AccountId,
+	collection: CollectionId,
+	item: ItemId,
+	witness: Option<MintWitness>,
```
Daanvdplas, line 396: This parameter name is not developer friendly, can we improve this in some way?

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -0,0 +1,198 @@
...
+/// A witness data to cancel attributes approval operation.
+#[derive(Debug, PartialEq, Eq)]
+#[ink::scale_derive(Encode, Decode, TypeInfo)]
+pub struct CancelAttributesApprovalWitness {
+	/// An amount of attributes previously created by account.
+	pub account_attributes: u32,
+}
+
+/// Collection's configuration.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
Daanvdplas, line 36: Clone and Copy trait derived, any reason?

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -0,0 +1,198 @@
...
+
+use super::*;
+use crate::{macros::impl_codec_bitflags, primitives::AccountId};
+
+pub type ItemId = u32;
+pub type CollectionId = u32;
+type Balance = u32;
+
+/// Attribute namespaces for non-fungible tokens.
+#[derive(Debug, PartialEq, Eq)]
```
Daanvdplas, line 13: We should be concious with adding these derives, only if they are really necessary should they be added.

pop-api/src/v0/nonfungibles/types.rs
Daanvdplas: They should be inline with the type in pallet nfts and that is what it will be decoded to.

pop-api/src/v0/nonfungibles/types.rs
Daanvdplas: Same for others

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -0,0 +1,198 @@
...
+	/// When this isn't set then the deposit is required to hold the items of this collection.
+	DepositRequired,
+}
+
+/// Wrapper type for `BitFlags<CollectionSetting>` that implements `Codec`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct CollectionSettings(pub BitFlags<CollectionSetting>);
+
+impl CollectionSettings {
+	pub fn from_disabled(settings: BitFlags<CollectionSetting>) -> Self {
```
Daanvdplas, line 89: I find this dx very bad as it is not clear whether I provide settings and they will be included or excluded from the CollectionSettings of the collection I create. Not sure whether it is even useful here. I remember the implementation in pallet nfts also to be weird, please implement the best devex you can implement.

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -0,0 +1,198 @@
...
+		Self(settings)
+	}
+
+	#[cfg(feature = "std")]
+	pub fn all_enabled() -> Self {
```
Daanvdplas, line 94: If no tests use it, remove

Daanvdplas: Also check ItemSettings

pop-api/src/macros.rs
```diff
@@ -0,0 +1,53 @@
...
+/// `$size` is an integer type (e.g., u8, u16, u32) that can represent the bitflags.
+/// The `$bitflag_enum` type is the enumeration type that defines the individual bitflags.
+///
+/// This macro provides implementations for the following traits:
+/// - `MaxEncodedLen`: Calculates the maximum encoded length for the wrapper type.
+/// - `Encode`: Encodes the wrapper type using the provided encoding function.
+/// - `EncodeLike`: Trait indicating the type can be encoded as is.
+/// - `Decode`: Decodes the wrapper type from the input.
+/// - `TypeInfo`: Provides type information for the wrapper type.
+macro_rules! impl_codec_bitflags {
```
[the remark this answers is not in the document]
Daanvdplas: Also I am interested to hear what effect this macro has on the contract size.

Daanvdplas (changes requested):
PR looking great! A few things that are important and need closer attention:
- Large types such as enums and structs increase the contract size fast so should be minimised as much as possible. 
- Only code that is needed for the smart contracts should be added, e.g. any unnecessary trait derived is code added which increases the contract size.
- Please reassess the types and implementations copied from pallet nfts as the requirements in the pop api are different.

To remind ourselves of the pop api design goals ([ref](https://app.shortcut.com/r0gue/write/IkRvYyI6I3V1aWQgIjY2MzQwMzUyLWI1MWUtNDZmNy05NWJmLWNkOTBlODQ4NDk2YiI=)):
```
The Pop API must be designed to be:

Stable: contracts are generally immutable and therefore need a stable interface, especially as the runtime may be evolving over time.
Future Proof: technologies improve, so the API implementation should be adaptable to new approaches without affecting the interface. The API implementation should also select the optimal approach to realise a use case from a range of possible technical implementations.
Versioned: not every future technology can be retro-fitted, so the API will be versioned.
Simple: we want to make the use of the API by smart contract developers as easy to use as possible. A one-to-one mapping to pallet functionality would echo complexity up the stack. Pop API strives to shield dApp developers from this complexity.
Storage Efficient: The contracts size impacts the cost to deploy.
```

Especially point 4 & 5 are important for this PR!

Last, it would be good to see the contract size compared to the fungibles.

pop-api/src/v0/nonfungibles/mod.rs
```diff
@@ -0,0 +1,466 @@
...
+		.handle_error_code::<StatusCode>()
+		.call(&(collection, item, namespace, key))
+}
+
+/// Returns the details of the `collection`.
+///
+/// # Parameters
+/// - `collection` - The collection.
+#[inline]
+pub fn collection(collection: CollectionId) -> Result<Option<CollectionDetails>> {
```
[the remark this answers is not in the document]
chungquantin: Got an issue with the integration test for `collection` read query that the collection details can't be decoded (maybe because it is too large). Below fields can be queried using other existing reads.
- `owner -> collection_owner`
- `items -> collection_items`

These fields can't be queried yet: 
```rs
/// The total number of outstanding item metadata of this collection.
pub item_metadatas: u32,
/// The total number of outstanding item configs of this collection.
pub item_configs: u32,
/// The total number of attributes for this collection.
pub attributes: u32,
```
My suggestion is to have additional reads to return these fields separately. Hence, reducing the contract size and make it more efficient, devs only need to query a field they need instead of returning everything to them.

pop-api/src/primitives.rs
chungquantin: Yes, there's no doc because there is a comment above already.

pop-api/src/v0/nonfungibles/types.rs
chungquantin: If bitflag is used, these must be provided

pop-api/src/v0/nonfungibles/traits.rs
```diff
@@ -0,0 +1,85 @@
...
+#[ink::trait_definition]
+pub trait Psp34Metadata {
+	/// Returns the attribute of `item` for the given `key`.
+	///
+	/// # Parameters
+	/// * `item` - The item. If `None` the attributes for the collection are queried.
+	/// * `namespace` - The attribute's namespace.
+	/// * `key` - The key of the attribute.
+	#[ink(message, selector = 0xf19d48d1)]
+	fn get_attribute(&self, id: ItemId, key: Vec<u8>) -> Option<Vec<u8>>;
```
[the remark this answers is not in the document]
chungquantin: This does not follow the standard, I know we already deviated from the standard by introducing the collection parameter and namespace...But I thought these traits are provided as additional things as there was no official library that provide these traits? 

If we add `namespace` to `get_attribute`, this breaks the standard completely and contract deployed on Pop won't be compatible with other ink! chains. Plus, the method in spec does not have `namespace`.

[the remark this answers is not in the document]
Daanvdplas: Yes you are right, it should not be added to the psp34 trait as the standard doesn't have it. The `namespace` parameter in the comment above confused me. 

> This does not follow the standard, I know we already deviated from the standard by introducing the collection parameter

If you look at the fungibles example (psp22 standard contract), you can see that the contract is compliant per its function definitions, not the api that it is interacting with.

pop-api/src/v0/nonfungibles/mod.rs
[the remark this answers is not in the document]
Daanvdplas: I think those fields are only important to verify the destroy witness when wanting to destroy an asset. Please verify but perhaps these are never needed for dapp developers.

pop-api/src/v0/nonfungibles/types.rs
Daanvdplas: Good point.

pop-api/src/v0/nonfungibles/mod.rs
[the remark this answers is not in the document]
chungquantin: Removed in the main PR and created an issue #464 so we can revisit this later. I agree that other fields might not be necessary for the use case of the dapps.

pop-api/src/v0/nonfungibles/types.rs
chungquantin: Resolved by reordering these types following the order in pallet-nfts

pop-api/src/macros.rs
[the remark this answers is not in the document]
chungquantin: Based on the last observation on the [PR merged to sub0](https://github.com/r0gue-io/pop-node/pull/353), the contract size is still quite small.

pop-api/src/v0/nonfungibles/types.rs
chungquantin: `Debug` and `Eq` provides basic use case of the struct. We don't know when these will be used by the developer but to me, providing them won't increase the contract size much.

pop-api/src/macros.rs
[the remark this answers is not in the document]
chungquantin: > It would be good if you could investigate whether all the traits derived here are necessary for the smart contract implementation.

All types are required by ink! contract.

pop-api/src/v0/nonfungibles/traits.rs
```diff
@@ -0,0 +1,100 @@
...
+	/// # Parameters
+	/// - `item` - The item. If `None` the attributes for the collection are queried.
+	/// - `key` - The key of the attribute.
+	#[ink(message, selector = 0xf19d48d1)]
+	fn get_attribute(&self, item: ItemId, key: Vec<u8>) -> Option<Vec<u8>>;
+}
+
+/// The PSP34 Mintable trait.
+#[ink::trait_definition]
+pub trait Psp34Mintable {
```
chungquantin, line 81: These traits are not in the spec. I researched other Psp34 trait of [AZ](https://github.com/Cardinal-Cryptography/PSP34/blob/main/traits.rs) and [Openbrush](https://github.com/Brushfam/openbrush-contracts/blob/main/contracts/src/token/psp34/extensions/mintable.rs) and both have different interfaces.

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -0,0 +1,170 @@
+//! A set of data types for use in smart contracts interacting with the non-fungibles API.
```
Daanvdplas, line 1: One thing that we need to make sure here is that if a change in pallet nfts occurs with one of those types, then our tests should fail somewhere. We have the primitives crate where we could place those types and add tests but this needs some thinking.

pop-api/src/v0/nonfungibles/types.rs
Daanvdplas: Why don't we remove this one?

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -0,0 +1,170 @@
...
+	UnlockedAttributes,
+}
+
+/// Wrapper type for `BitFlags<ItemSetting>` that implements `Codec`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct ItemSettings(pub BitFlags<ItemSetting>);
+
+impl ItemSettings {
+	/// Enable all features on an item.
+	pub fn all_enabled() -> Self {
```
Daanvdplas, line 160: Same, do we need this if the nfts pallet only used it for tests?

Daanvdplas (commented):
PR looking really good. We need to agree on where the types are placed and how we ensure that we get alerted when something changes about these types. These types need to be in primitives I think because when an older smart contract using the api with an older version (with an outdated type), the runtime also needs to know how to decode this outdated type.

Side note: could I resolve my comments?

pop-api/src/v0/nonfungibles/types.rs
chungquantin: It's weird to me why this is only used for test in the pallet-nfts. To me, this method is necessary as a default option for developers to set all features enabled.

pop-api/src/v0/nonfungibles/errors.rs
```diff
@@ -0,0 +1,55 @@
...
+	TokenNotExists,
+	/// Returned if safe transfer check fails.
+	SafeTransferCheckFailed(String),
+}
+
+#[cfg(feature = "std")]
+impl From<Psp34Error> for u32 {
+	fn from(value: Psp34Error) -> u32 {
+		match value {
+			Psp34Error::NotApproved => u32::from_le_bytes([MODULE_ERROR, NFTS, 0, 0]),
```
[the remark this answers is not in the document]
chungquantin: Makes sense. So no changes need to be made as it is already converted to `NoPermission`

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -26,25 +27,35 @@ mod tests;
...
 type ItemIdOf<T> = <NftsOf<T> as Inspect<<T as frame_system::Config>::AccountId>>::ItemId;
 type ItemPriceOf<T> = BalanceOf<T>;
-type NextCollectionIdOf<T> = pallet_nfts::NextCollectionId<T, NftsInstanceOf<T>>;
 type NftsErrorOf<T> = pallet_nfts::Error<T, NftsInstanceOf<T>>;
 type NftsInstanceOf<T> = <T as Config>::NftsInstance;
 type NftsOf<T> = pallet_nfts::Pallet<T, NftsInstanceOf<T>>;
 type NftsWeightInfoOf<T> = <T as pallet_nfts::Config<NftsInstanceOf<T>>>::WeightInfo;
 type WeightOf<T> = <T as Config>::WeightInfo;
+// Public due to pop-api integration tests crate.
+pub type AccountBalanceOf<T> = pallet_nfts::AccountBalance<T, NftsInstanceOf<T>>;
```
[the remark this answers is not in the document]
Daanvdplas: That doesn't answer my first part of my comment.

pallets/api/src/nonfungibles/mod.rs
```diff
@@ -207,7 +218,7 @@ pub mod pallet {
 		pub fn create(
 			origin: OriginFor<T>,
 			admin: AccountIdOf<T>,
-			config: CollectionConfigOf<T>,
+			config: CollectionConfigFor<T>,
```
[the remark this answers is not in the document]
Daanvdplas: But didn't we already said previously that we want consistency and have `...Of`

pop-api/Cargo.toml
```diff
@@ -9,7 +9,9 @@ version = "0.0.0"
 bitflags = { version = "1.3.2" }
 enumflags2 = "0.7.9"
 ink = { version = "5.1.0", default-features = false }
+pallet-nfts = { path = "../pallets/nfts", default-features = false, optional = true }
```
Daanvdplas, line 12: This is incorrect, we should not have a pallet dependency in a contract library. 

In addition, we do need to know those types in the runtime. Imagine a contract that uses old nft types interact with the runtime, aka a contract using an older version. The runtime needs to know these older types in order to decode the bytes correctly. Therefore, those types should be in the primitives crate, also with the tests. 

Please do some thinking on this yourself and take a longer term view

pop-api/src/v0/nonfungibles/types.rs
Daanvdplas: Can you check whether all these features should be able to be enabled. I have a vague memory of an edge case there, something along the lines of a feature that only root can enable. Please ignore if you cant find anything.

pallets/api/src/nonfungibles/mod.rs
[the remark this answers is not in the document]
Daanvdplas: Understood

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -0,0 +1,335 @@
...
+	/// Provide `settings` bit flags indicate which features are turned off.
+	pub fn from_disabled(settings: BitFlags<ItemSetting>) -> Self {
+		Self(settings)
+	}
+}
+
+impl_codec_bitflags!(ItemSettings, u64, ItemSetting);
+
+#[cfg(test)]
+mod tests {
```
peterwht, line 173: love these tests!

pop-api/Cargo.toml
peterwht: should be moved to dev-dependencies if only used for tests.

peterwht (changes requested):
Excellent job. Pretty close to merging IMO!

Left a few nitpicks that should be resolved.

Regarding the type issue, I think currently the types should remain in the `pop-api` crate under version 0 of non-fungibles.

Daan makes a valid point, however, I think we should cross that bridge when we get there. In this case, "kicking the can down the road" makes sense because this pop-api will change with precompiles. 

Additionally, it's an easy enough refactor if we face that problem in the future (a concrete problem). While right now we are spending time on a problem that does not exist yet. The existing testing is solid to make sure we know if a type changes.

pop-api/src/v0/nonfungibles/traits.rs
```diff
@@ -0,0 +1,100 @@
...
+/// The PSP34 Burnable trait.
+#[ink::trait_definition]
+pub trait Psp34Burnable {
+	/// Destroys the specified item. Clearing the corresponding approvals.
+	///
+	/// # Parameters
+	/// - `item` - The item.
+	#[ink(message, selector = 0x63c9877a)]
+	fn burn(&mut self, item: ItemId) -> Result<(), Psp34Error>;
+}
```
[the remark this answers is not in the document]
chungquantin: We did not provide this trait for fungibles as well. I can't recall what was the reason so would be great to have an input from @Daanvdplas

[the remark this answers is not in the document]
Daanvdplas: Management interface is coming from us not the psp standard

pallets/api/src/nonfungibles/mod.rs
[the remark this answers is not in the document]
chungquantin: Resolved by removing those types

Daanvdplas (approved):
Well done! PR is looking good!

pop-api/src/v0/nonfungibles/traits.rs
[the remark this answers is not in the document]
peterwht: where does a contract implement the management interface though? Do they just have to manually create the functions themselves?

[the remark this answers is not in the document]
chungquantin: Here we separate those dispatchable in sets for better management but with traits expected to be used by the smart contract developers, I don't think providing `Management` is useful. There are a lot of functions in the `Management` set that the trait won't be that helpful. Are any opinions would be valuable here. 

cc: @Daanvdplas

pop-api/src/v0/nonfungibles/types.rs
chungquantin: Removed the feature gate but still keep the method because I checked the code in pallet-nfts and this method is needed: https://docs.rs/pallet-nfts/latest/src/pallet_nfts/types.rs.html#274

peterwht (approved):

PR #466, opened 2025-02-06
refactor: integration tests mainnet xcm

Daanvdplas:
The tests need to be improved in general but refactored it so that there is no mainnet specific code to a runtime generic test.

### DIFF
--- a/integration-tests/src/lib.rs
+++ b/integration-tests/src/lib.rs
@@ -6,7 +6,10 @@ use chains::{
 		genesis::ED as ASSET_HUB_ED, runtime::xcm_config::XcmConfig as AssetHubXcmConfig, AssetHub,
 		AssetHubParaPallet,
 	},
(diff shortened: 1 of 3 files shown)


integration-tests/src/lib.rs
```diff
@@ -315,23 +313,8 @@ fn reserve_transfer_native_asset_from_para_to_system_para() {
...
-		)
-	});
-
-	// Mainnet ED is 100 times higher.
-	let pop_mainnet_ed = POP_ED * 100;
-	// We need to keep `delivery_fees` in balance to pay for these fees locally.
-	// To avoid the account from being reaped we need to ensure its balance is >= ED.
-	let amount_to_send = amount_to_send - (pop_mainnet_ed + delivery_fees);
-	let assets: Assets = (Parent, amount_to_send).into();
+	let amount_to_send = PopNetworkPara::account_data_of(PopNetworkParaReceiver::get()).free / 4; // bob on pop balance
```
al3mart, line 316: what's the reason for the division ?

PR #469, opened 2025-02-14
chore(utility): make deposit calc explicit.

al3mart:
Improves how deposits are presented in utility config for mainnet runtime.


Daanvdplas (approved):
Nice thank you!

PR #470, opened 2025-02-17
messaging spec wip

f-gate:


messaging-spec.md
```diff
@@ -0,0 +1,351 @@
+# Maximus Messaging Monilith!
+The messaging feature enables smart contracts on pop-net to seamlessly interact with external chains using XCM, ISMP, GMP, and other Substrate-compatible messaging protocols. By providing a flexible and efficient messaging framework, it expands multichain use cases while giving developers full control over request handling and response processing. This system enhances interoperability and significantly improves the developer experience.
```
peterwht, line 2: for the GMP part, do you think it makes sense to rephrase similar to: "XCM, ISMP, and support for other adding other protocols, like GMP"

messaging-spec.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Non-deliverables
+- A high-level abstraction layer for developers (this version focuses on raw ISMP/XCM capabilities)
+- Prebuilt integrations with specific parachains.
+- Automated retries or reliability mechanisms beyond ISMP/XCM’s existing guarantees.
+- Supporting ISMP data processing for RLP encoded contracts.
+- Continuously running relayer (e.g. in pop-node that automatically relays messages).
+- Adding other protocols beyond ISMP and XCM.
+- Drink! e2e testing.
+- Refunding unused weight to developers.
```
peterwht, line 45: Depending on how you tackle the weight, this might be kinda easy to do. With fresh eyes, this seems like we should try to make this possible.

messaging-spec.md
```diff
@@ -0,0 +1,351 @@
...
+## Architecture
+todo!();
+
+
+
+## Considerations: 
+### Weight + Benchmarking
+#### Weight Flow inspired from POC:
+0. Contract is called with "to be charged" weight included.
+1. Weight charged in contract by pallet-contracts metering.
```
peterwht, line 59:
```suggestion
1. Weight charged in contract by pallet-revive metering.
```

messaging-spec.md
```diff
@@ -0,0 +1,351 @@
...
+#### On charging for future execution on the current block
+Consider the argument.
+
+1. When the chain extension invokes a dispatchable function, the weight of the associated callback is already included in its calculation.
+2. The chain extension charges the contract environment for the full weight of the dispatchable call.
+3. This weight contributes to filling the current block (Block 1), meaning some of the available blockspace is consumed.
+4. However, the callback itself is scheduled for execution in a future block (Block 2).
+5. When Block 2 arrives, it must also allocate space to execute the callback, consuming additional blockspace.
+
+Conclusion: Since blockspace is consumed both when the dispatchable is initially called and when the callback executes, we effectively pay for the callback's weight twice, once in Block 1 and again in Block 2.
```
peterwht, line 82: I don't fully follow this. The runtime controls when the callback executes, which means the runtime can choose to charge weight or not -- so we wouldn't charge the callback twice.

messaging-spec.md
```diff
@@ -0,0 +1,351 @@
...
+When pallet-xcm recieves the response it will decode the runtime call and call our extrinsic for response handling.
+
+### ISMP and Queries
+We can use the `IsmpDispatcher` to dispatch all outgoing calls.
+
+### Message IDs
+Message ids are used as a key for outgoing requests, they can be queried to find the status of a message.
+A message id can either be specified by the runtime or the contract.
+
+If the message id is specified by the runtime, the runtime will have to return that data to the contract to emit or emit an event itself. The event emitted must contain data that can relate the request to a response.
```
peterwht, line 142: I think we should choose one or the other (contract manages its own request IDs, or runtime manages request IDs). But not both as that adds complexity, collision checking, etc. 

- Contract passes request id seems to be simplest route for our pallet. 
- Runtime returns request id
    - Contract can still create their own system to map their requests to the pallets request ID
    - We would have to modify the API to support executing a dispatchable, and then returning a request ID. This may become too complicated

messaging-spec.md
```diff
@@ -0,0 +1,351 @@
...
+
+```rust 
+        /// Send an ismp get message and optionally specify a callback.
+		#[pallet::call_index(1)]
+		#[pallet::weight(
+			T::WeightInfo::ismp_get() + 
+            T::IsmpModuleWeights::on_response() + 
+            callback.map(|cb| cb.weight + T::CallbackExecutor::execution_weight()).unwrap_or_default()
+		)]
+		pub fn ismp_get(
```
peterwht, line 172: is it possible to have just a few high-level dispatchables?

Something like:
```
// pseudo code
enum ProtocolMessage;
enum ProtocolResponse;

fn send(message: ProtocolMessage);
fn query(message_id) -> ProtocolResponse;
fn remove(message_id);
```

The higher level architecture might look like:
Screenshot 2025-02-17 at 10 31 48 AM

peterwht (commented):
Looking really great! Left some comments throughout

messaging-spec.md
f-gate: i have added it to deliverables, i agree with the weights resolved properly this should be ok.

messaging-spec.md
f-gate: collision checking will happen by default (through seeing if the message exists before creating one).
Prevention of collision perhaps is unnecessary.

I will change to an active voice.

messaging-spec.md
f-gate: ill have something liek this ready for our call

messaging-spec.md
Daanvdplas: Perhaps how the scheduler pallet does this. We shouldn't be paying twice

f-gate:
Has been added to the docs on notion

PR #471, opened 2025-02-17
refactor(genesis): use runtime genesis presets

al3mart:
Closes: #246

Refactor genesis config to use presets for all runtimes:

- [x] Use genesis presets for devnet runtime: this PR
- [x] Use genesis presets for testnet runtime: #472 
- [x] Use genesis presets for mainnet runtime: #473

Each PR is been based on the one above as per this list. Merging should be easier if we waterfall starting from #473  and continue with #472 then this PR. Reviews should still be easier in order devnet -> mainnet as the changes are incremental.

These are the different spec that will use the devnet runtime. These are the ones we can pass to the node via `--chain <spec>`.

```rust
/// A development chain running on a single node, using the `devnet` runtime.
pub const DEVNET_DEV: &str = "pop-devnet-dev";
/// Configures a local chain running on multiple nodes for testing purposes, using the `devnet`
/// runtime.
pub const DEVNET_LOCAL: &str = "pop-devnet-local";
/// A live chain running on multiple nodes on private devnet, using the `devnet` runtime.
pub const DEVNET: &str = "pop-devnet";
```

Each of these use a different genesis configuration.

---

### Requirements:

- [x] There's an asset reserved for DOT on every config.

[sc-2373]

PR #472, opened 2025-02-17
refactor(testnet): use genesis presets for testnet runtime.

al3mart:
Based on #471 

This PR mainly takes the configuration we had in genesis and refactors the code to use presets. Some configs might need to be changed, but those are not included in this PR.

These are the different spec that will use the testnet runtime. These are the ones we can pass to the node via `--chain  <spec>`.

```rust
/// A development chain running on a single node, using the `testnet` runtime.
pub const TESTNET_DEV: &str = "pop-testnet-dev";
/// Configures a local chain running on multiple nodes for testing purposes, using the `testnet`
/// runtime.
pub const TESTNET_LOCAL: &str = "pop-testnet-local";
/// A live chain running on multiple nodes on private testnet, using the `testnet` runtime.
pub const TESTNET: &str = "pop-testnet";
```

---

### Requirements:

- [x] There's an asset reserved for DOT on every config.

[sc-2373]

### DIFF
--- a/runtime/testnet/src/genesis.rs
+++ b/runtime/testnet/src/genesis.rs
@@ -1,3 +1,5 @@
+use alloc::{vec, vec::Vec};
+
 use cumulus_primitives_core::ParaId;
 use parachains_common::{AccountId, AuraId, Balance};
 use pop_runtime_common::genesis::*;
@@ -26,8 +28,13 @@ const ENDOWMENT: Balance = 10_000_000 * UNIT;
 const SAFE_XCM_VERSION: u32 = xcm::prelude::XCM_VERSION;
 
 /// SUDO account set at genesis.
-const TESTNET_SUDO_ACCOUNT: AccountId =
-	AccountId::from_ss58check("5FPL3ZLqUk6MyBoZrQZ1Co29WAteX6T6N68TZ6jitHvhpyuD").unwrap();
+const TESTNET_SUDO_ACCOUNT: &str = "5FPL3ZLqUk6MyBoZrQZ1Co29WAteX6T6N68TZ6jitHvhpyuD";
+
+// Returns the SUDO account's address as an `AccountId`.
+// This function will return an error if the SS58 address is invalid.
+fn sudo_account_id() -> AccountId {
+	AccountId::from_ss58check(TESTNET_SUDO_ACCOUNT).expect("sudo address is valid SS58")
+}
 
 /// Returns a JSON blob representation of the built-in `RuntimeGenesisConfig` identified by `id`.
 pub(crate) fn get_preset(id: &PresetId) -> Option<Vec<u8>> {
@@ -67,7 +74,7 @@ fn development_config() -> Value {
 /// runtime.
 fn local_config() -> Value {
 	let mut endowed_accounts = dev_accounts();
-	endowed_accounts.push(TESTNET_SUDO_ACCOUNT);
+	endowed_accounts.push(sudo_account_id());
 
 	genesis(
 		// Initial collators.
@@ -77,7 +84,7 @@ fn local_config() -> Value {
 			(Keyring::Bob.to_account_id(), Keyring::Bob.public().into()),
 		]),
 		endowed_accounts,
-		TESTNET_SUDO_ACCOUNT,
+		sudo_account_id(),
 		PARA_ID,
 	)
 }
@@ -97,8 +104,6 @@ fn live_config() -> Value {
 		AccountId::from_ss58check("5GMqrQuWpyyBBK7LAWXR5psWvKc1QMqtiyasjp23VNKZWgh6").unwrap();
 	let collator_2_aura_id: AuraId =
 		AuraId::from_ss58check("5GMqrQuWpyyBBK7LAWXR5psWvKc1QMqtiyasjp23VNKZWgh6").unwrap();
-	let sudo_account_id: AccountId =
-		AccountId::from_ss58check("5FPL3ZLqUk6MyBoZrQZ1Co29WAteX6T6N68TZ6jitHvhpyuD").unwrap();
 
 	genesis(
 		// Initial collators.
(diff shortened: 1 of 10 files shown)


runtime/testnet/src/genesis.rs
```diff
@@ -0,0 +1,170 @@
...
+use sp_genesis_builder::PresetId;
+
+use crate::{AssetsConfig, BalancesConfig, SessionKeys, EXISTENTIAL_DEPOSIT, UNIT};
+
+/// A development chain running on a single node, using the `testnet` runtime.
+pub const TESTNET_DEV: &str = "pop-testnet-dev";
+/// Configures a local chain running on multiple nodes for testing purposes, using the `testnet`
+/// runtime.
+pub const TESTNET_LOCAL: &str = "pop-testnet-local";
+/// A live chain running on multiple nodes on private testnet, using the `testnet` runtime.
```
Daanvdplas, line 16:
```suggestion
/// A live chain running on multiple nodes, using the `testnet` runtime.
```

runtime/testnet/src/genesis.rs
```diff
@@ -0,0 +1,170 @@
...
+
+	genesis(
+		// Initial collators.
+		Vec::from([
+			// Multiple collators for local development chain.
+			(Keyring::Alice.to_account_id(), Keyring::Alice.public().into()),
+			(Keyring::Bob.to_account_id(), Keyring::Bob.public().into()),
+		]),
+		endowed_accounts,
+		sudo_account_id(),
```
Daanvdplas, line 87: I'm of the opinion that we should just make this alice. We are not using the local network so I would keep it as simple as possible. Same for the endowed_accounts above, I would do dev_accounts.

runtime/testnet/src/genesis.rs
```diff
@@ -0,0 +1,170 @@
...
+		vec![
+			// POP COLLATOR 0
+			(collator_0_account_id, collator_0_aura_id),
+			// POP COLLATOR 1
+			(collator_1_account_id, collator_1_aura_id),
+			// POP COLLATOR 2
+			(collator_2_account_id, collator_2_aura_id),
+		],
+		vec![],
+		sudo_account_id(),
```
Daanvdplas, line 119: Didn't we want to have our upon agreed governance on testnet as well so that we get the hang of it?

runtime/testnet/src/genesis.rs
```diff
@@ -0,0 +1,170 @@
...
+	invulnerables: Vec<(AccountId, AuraId)>,
+	endowed_accounts: Vec<AccountId>,
+	sudo_key: AccountId,
+	id: ParaId,
+) -> Value {
+	json!({
+		"assets": AssetsConfig {
+			// Genesis assets: Vec<(id, owner, is_sufficient, min_balance)>
+			assets: Vec::from([
+				(0, asset_hub_sa_on_pop(), false, EXISTENTIAL_DEPOSIT),	// Relay native asset from Asset Hub
```
Daanvdplas, line 135: change the owner account

runtime/testnet/src/genesis.rs
```diff
@@ -0,0 +1,170 @@
...
+/// Configures a local chain running on multiple nodes for testing purposes, using the `testnet`
+/// runtime.
+pub const TESTNET_LOCAL: &str = "pop-testnet-local";
+/// A live chain running on multiple nodes on private testnet, using the `testnet` runtime.
+pub const TESTNET: &str = "pop-testnet";
+/// The available genesis config presets;
+const PRESETS: [&str; 3] = [TESTNET_DEV, TESTNET_LOCAL, TESTNET];
+
+/// The parachain identifier to set in genesis config.
+pub const PARA_ID: ParaId = ParaId::new(4_001);
```
Daanvdplas, line 22: Same as devnet and could be a problem if we both launch on paseo.

Daanvdplas (commented):
Left a few comments

runtime/testnet/src/genesis.rs
al3mart: Kind of scope creep for this PR, isn't it. Agree that the testnet runtime needs to change, but this PR is only refactoring the code so that we can have presets set for the runtime.

runtime/testnet/src/genesis.rs
al3mart: Yeah, this PR mainly takes what we had in genesis and refactors the code to use presets, that we were not using before.

I agree that the runtime should mirror mainnet in terms on pallets features and this configuration should change in consequence.

Daanvdplas: > this PR mainly takes what we had in genesis and refactors the code to use presets

Missed this, can we make a clear note in the PR descriptions for other reviewers.

peterwht (approved):
Nicely done. Giving an approval. Would be nice to have the local config account be a dev account before merging though :)

runtime/testnet/src/genesis.rs
al3mart: Better that way [97e18c5](https://github.com/r0gue-io/pop-node/pull/472/commits/97e18c5134055f674960fea658b484d339730d19)

runtime/testnet/src/genesis.rs
al3mart: Changes here: [ab13ef5](https://github.com/r0gue-io/pop-node/pull/472/commits/ab13ef5f4f89d892f3ab8ae21525038bed18355b) :+1:

runtime/testnet/src/genesis.rs
al3mart: Pushed this change in: [97e18c5](https://github.com/r0gue-io/pop-node/pull/472/commits/97e18c5134055f674960fea658b484d339730d19)

Daanvdplas (approved):
Looking good! Lets wait for the CI to be green and then we can merge!

Daanvdplas:
Note that the PR that has been merged involves mainnet and we agreed to have minimal 2 reviews for that. Please uphold to this requirement as we are responsible for a secure runtime!

peterwht (approved):

peterwht:
Good to merge! Mainnet checks out as well

PR #473, opened 2025-02-17
refactor(mainnet): use genesis presets for mainnet runtime.

al3mart:
Based on #472 

Moves genesis configuration for mainnet into the runtime.

These are the different spec that will use the mainnet runtime. These are the ones we can pass to the node via `--chain  <spec>`.

```rust
/// A development chain running on a single node, using the `mainnet` runtime.
pub const MAINNET_DEV: &str = "pop-dev";
/// Configures a local chain running on multiple nodes for testing purposes, using the `mainnet`
/// runtime.
pub const MAINNET_LOCAL: &str = "pop-local";
/// A live chain running on multiple nodes, using the `mainnet` runtime.
pub const MAINNET: &str = "pop";
```

---

### Requirements:

- [x] There's an asset reserved for DOT on development and local configuration.
- [x] Council seats are populated in genesis state

[sc-2373]

### DIFF
--- a/runtime/mainnet/src/genesis.rs
+++ b/runtime/mainnet/src/genesis.rs
@@ -86,7 +86,7 @@ fn development_config() -> Value {
 		// AssetId reserved for DOT from AH.
 		vec![GenesisAsset {
 			id: 0,
-			owner: asset_hub_sa_on_pop(),
+			owner: Keyring::Alice.to_account_id(),
 			is_sufficient: false,
 			min_balance: ExistentialDeposit::get(),
 			name: "DOT".into(),
@@ -134,7 +134,7 @@ fn local_config() -> Value {
 		// AssetId reserved for DOT from AH.
 		vec![GenesisAsset {
 			id: 0,
-			owner: asset_hub_sa_on_pop(),
+			owner: SudoAddress::get(),
 			is_sufficient: false,
 			min_balance: ExistentialDeposit::get(),
 			name: "DOT".into(),
@@ -299,6 +299,10 @@ mod tests {
 
 			let sudo_key = genesis["sudo"]["key"].as_str().unwrap();
 			assert_eq!(sudo_key, "5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY");
+			assert_eq!(
+				AccountId::from_ss58check(sudo_key).unwrap(),
+				Keyring::Alice.to_account_id()
+			);
 		}
 
 		#[test]
@@ -329,12 +333,10 @@ mod tests {
 
 			// AssetId is 0.
 			assert_eq!(assets[0].0, 0);
-			// Owner is ah_sa_on_pop
-			assert_eq!(assets[0].1, "5Eg2fntNprdN3FgH4sfEaaZhYtddZQSQUqvYJ1f2mLtinVhV");
+			// Owner is SudoAccount
 			assert_eq!(
-				asset_hub_sa_on_pop(),
-				AccountId::from_ss58check("5Eg2fntNprdN3FgH4sfEaaZhYtddZQSQUqvYJ1f2mLtinVhV")
-					.unwrap()
+				AccountId::from_ss58check(assets[0].1.as_str()).unwrap(),
+				Keyring::Alice.to_account_id()
 			);
 			// Asset is not sufficient
 			assert_eq!(assets[0].2, false);
@@ -449,12 +451,10 @@ mod tests {
 
 			// AssetId is 0.
 			assert_eq!(assets[0].0, 0);
-			// Owner is ah_sa_on_pop
-			assert_eq!(assets[0].1, "5Eg2fntNprdN3FgH4sfEaaZhYtddZQSQUqvYJ1f2mLtinVhV");
+			// Owner is SudoAddress
 			assert_eq!(
-				asset_hub_sa_on_pop(),
-				AccountId::from_ss58check("5Eg2fntNprdN3FgH4sfEaaZhYtddZQSQUqvYJ1f2mLtinVhV")
-					.unwrap()
+				AccountId::from_ss58check(assets[0].1.as_str()).unwrap(),
+				SudoAddress::get()
 			);
 			// Asset is not sufficient
 			assert_eq!(assets[0].2, false);
--- a/runtime/mainnet/src/genesis.rs
+++ b/runtime/mainnet/src/genesis.rs
@@ -5,9 +5,14 @@ use parachains_common::{AccountId, AuraId, Balance};
 use pop_runtime_common::genesis::*;
 use sp_core::crypto::Ss58Codec;
 use sp_genesis_builder::PresetId;
+use sp_runtime::traits::AccountIdConversion;
 
 use crate::{
-	config::{governance::SudoAddress, monetary::ExistentialDeposit},
+	config::{
+		collation::PotId,
+		governance::SudoAddress,
+		monetary::{ExistentialDeposit, MaintenanceAccount, TreasuryAccount},
+	},
 	AssetsConfig, BalancesConfig, CouncilConfig, SessionKeys, EXISTENTIAL_DEPOSIT, UNIT,
 };
 
@@ -63,13 +68,19 @@ pub(crate) fn presets() -> Vec<PresetId> {
 
 /// Configures a development chain running on a single node, using the `mainnet` runtime.
 fn development_config() -> Value {
+	let mut endowed_accounts = dev_accounts();
+	endowed_accounts.push(asset_hub_sa_on_pop());
+	endowed_accounts.push(MaintenanceAccount::get());
+	endowed_accounts.push(PotId::get().into_account_truncating());
+	endowed_accounts.push(TreasuryAccount::get());
+
 	genesis(
 		// Initial collators.
 		Vec::from([
 			// Single collator for development chain
 			(Keyring::Alice.to_account_id(), Keyring::Alice.public().into()),
 		]),
-		dev_accounts(),
+		endowed_accounts,
 		Keyring::Alice.to_account_id(),
 		PARA_ID,
 		// AssetId reserved for DOT from AH.
@@ -95,8 +106,20 @@ fn development_config() -> Value {
 /// Configures a local chain running on multiple nodes for testing purposes, using the `mainnet`
 /// runtime.
 fn local_config() -> Value {
+	let council_members = vec![
+		AccountId::from_ss58check("142zako1kfvrpQ7pJKYR8iGUD58i4wjb78FUsmJ9WcXmkM5z").unwrap(),
+		AccountId::from_ss58check("15VPagCVayS6XvT5RogPYop3BJTJzwqR2mCGR1kVn3w58ygg").unwrap(),
+		AccountId::from_ss58check("14G3CUFnZUBnHZUhahexSZ6AgemaW9zMHBnGccy3df7actf4").unwrap(),
+		AccountId::from_ss58check("15k9niqckMg338cFBoz9vWFGwnCtwPBquKvqJEfHApijZkDz").unwrap(),
+		AccountId::from_ss58check("13BL7T6bTgeEdfEdZqLCKJZPN8ncyFNxxHRKFb2YMATvyfH4").unwrap(),
+	];
 	let mut endowed_accounts = dev_accounts();
+	endowed_accounts.push(asset_hub_sa_on_pop());
+	endowed_accounts.extend(council_members.clone());
+	endowed_accounts.push(MaintenanceAccount::get());
+	endowed_accounts.push(PotId::get().into_account_truncating());
 	endowed_accounts.push(SudoAddress::get());
+	endowed_accounts.push(TreasuryAccount::get());
 
 	genesis(
 		// Initial collators.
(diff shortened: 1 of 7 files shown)


runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,538 @@
...
+	genesis(
+		// Initial collators.
+		vec![
+			// POP COLLATOR 0
+			(collator_0_account_id, collator_0_aura_id),
+		],
+		vec![],
+		SudoAddress::get(),
+		PARA_ID,
+		vec![],
```
Daanvdplas, line 165: Don't we want the same? Reserving an asset id for dot?

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,538 @@
...
+
+// The initial balances at genesis.
+fn balances(endowed_accounts: Vec<AccountId>) -> Vec<(AccountId, Balance)> {
+	let balances = endowed_accounts
+		.iter()
+		.cloned()
+		.map(|k| {
+			if k == asset_hub_sa_on_pop() {
+				// Onboarded DOT needs to be held on AH sa account on pop.
+				(k, ENDOWMENT.saturating_mul(endowed_accounts.len() as Balance))
```
Daanvdplas, line 230: I don't follow this?

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,538 @@
...
+			assert!(assets.is_empty());
+		}
+
+		#[test]
+		fn ensure_council_members_are_not_set() {
+			let genesis = live_config();
+
+			let council = genesis["council"]["members"].as_array().unwrap();
+			assert!(council.is_empty());
+		}
```
Daanvdplas, line 536: What about tests for the session genesis and collator selection?

Daanvdplas (commented):
Well done, left some comments. 

This PR is heavily correlated with the launch of devnet and testnet. Would be nice to hear what the plan is in terms of updating keys etc.

runtime/mainnet/src/genesis.rs
al3mart: Pop is already registered in Polkadot.
This genesis config is the translation of what we had into the newer version of the code. Arguably unnecessary work, but will make the transition to the new testnet way easier.

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,538 @@
...
+		// Initial collators.
+		vec![
+			// POP COLLATOR 0
+			(collator_0_account_id, collator_0_aura_id),
+		],
+		vec![],
+		SudoAddress::get(),
+		PARA_ID,
+		vec![],
+		vec![],
```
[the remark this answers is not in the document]
al3mart: Same as my answer above.

runtime/mainnet/src/genesis.rs
al3mart: yeah, that's just wrong, don't  know what I was thinking here.

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,538 @@
...
+			Keyring::Eve.to_account_id(),
+		],
+	)
+}
+
+/// Configures a local chain running on multiple nodes for testing purposes, using the `mainnet`
+/// runtime.
+fn local_config() -> Value {
+	let council_members = vec![
+		AccountId::from_ss58check("142zako1kfvrpQ7pJKYR8iGUD58i4wjb78FUsmJ9WcXmkM5z").unwrap(),
```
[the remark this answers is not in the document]
al3mart: Happy to adapt

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,538 @@
...
+	)
+}
+
+#[allow(clippy::too_many_arguments)]
+fn genesis(
+	invulnerables: Vec<(AccountId, AuraId)>,
+	endowed_accounts: Vec<AccountId>,
+	sudo_key: AccountId,
+	id: ParaId,
+	genesis_assets: Vec<GenesisAsset>,
```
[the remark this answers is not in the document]
al3mart: Registred state for live Pop doesn't have any asset registered at genesis. That's the reason for the parameter, you can see the dev and local config using Sudo as owner.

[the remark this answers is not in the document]
Daanvdplas: But we should

runtime/mainnet/src/genesis.rs
Daanvdplas: I don't follow, we want the asset identifier reserved in pallet assets so that we can use it for the pop api. We will need this for mainnet as well.

runtime/mainnet/src/genesis.rs
al3mart: Included in [03c47e6](https://github.com/r0gue-io/pop-node/pull/473/commits/03c47e66cfda81815c872be901726e6c11f2ee5c)

runtime/mainnet/src/genesis.rs
al3mart: Fixed in: [03c47e6](https://github.com/r0gue-io/pop-node/pull/473/commits/03c47e66cfda81815c872be901726e6c11f2ee5c)

runtime/mainnet/src/genesis.rs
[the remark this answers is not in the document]
al3mart: We are now using dev accounts: [03c47e6](https://github.com/r0gue-io/pop-node/pull/473/commits/03c47e66cfda81815c872be901726e6c11f2ee5c)

runtime/mainnet/src/genesis.rs
al3mart: Pop genesis state is already registered in Polkadot.

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,610 @@
...
+pub const MAINNET_LOCAL: &str = "pop-local";
+/// A live chain running on multiple nodes, using the `mainnet` runtime.
+pub const MAINNET: &str = "pop";
+/// The available genesis config presets;
+const PRESETS: [&str; 3] = [MAINNET_DEV, MAINNET_LOCAL, MAINNET];
+
+/// The parachain identifier to set in genesis config.
+pub const PARA_ID: ParaId = ParaId::new(3395);
+
+/// Initial balance for genesis endowed accounts.
```
peterwht, line 32: Nit: because this is the mainnet config, we this might be misleading at first glance and make someone thing we are endowing accounts with balance at genesis.

Could be worthwhile to clarify this is genesis endowed accounts for local development only.
```suggestion
/// Initial balance for genesis endowed accounts for local testing only.
```

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,610 @@
...
+pub(crate) fn presets() -> Vec<PresetId> {
+	PRESETS.map(PresetId::from).to_vec()
+}
+
+/// Configures a development chain running on a single node, using the `mainnet` runtime.
+fn development_config() -> Value {
+	let mut endowed_accounts = dev_accounts();
+	endowed_accounts.push(MaintenanceAccount::get());
+	endowed_accounts.push(PotId::get().into_account_truncating());
+	endowed_accounts.push(TreasuryAccount::get());
```
peterwht, line 74: Thoughts on giving these accounts only ED?

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,610 @@
...
+		// AssetId reserved for DOT from AH.
+		vec![GenesisAsset {
+			id: 0,
+			owner: Keyring::Alice.to_account_id(),
+			is_sufficient: false,
+			min_balance: ExistentialDeposit::get(),
+			name: "DOT".into(),
+			symbol: "DOT".into(),
+			decimals: 10,
+		}],
```
peterwht, line 94: I realize you are getting conflicting code direction. However, for mainnet I would like to exclude this. 

I am still not sold that we should have devs use the fungibles API for DOT. It introduces complexity due to pallet-balances missing functionality that pallet-assets have.

By including, we can forget its there. By excluding, we have to consciously add it in -- when we confirm it works.  

If we want to reserve asset id 0, I recommend using another name than "DOT" as a placeholder.

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,610 @@
...
+	)
+}
+
+/// Configures a local chain running on multiple nodes for testing purposes, using the `mainnet`
+/// runtime.
+fn local_config() -> Value {
+	let mut endowed_accounts = dev_accounts();
+	endowed_accounts.push(MaintenanceAccount::get());
+	endowed_accounts.push(PotId::get().into_account_truncating());
+	endowed_accounts.push(SudoAddress::get());
```
peterwht, line 111: SudoAddress is a multisig, right? That would mean for the local_config we wouldn't be able to use Sudo.

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,610 @@
...
+		// AssetId reserved for DOT from AH.
+		vec![GenesisAsset {
+			id: 0,
+			owner: SudoAddress::get(),
+			is_sufficient: false,
+			min_balance: ExistentialDeposit::get(),
+			name: "DOT".into(),
+			symbol: "DOT".into(),
+			decimals: 10,
+		}],
```
peterwht, line 142: Similarly, (for mainnet only) please remove or change the name from "DOT"

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,610 @@
...
+						SessionKeys { aura},// session keys
+					)
+				})
+				.collect::<Vec<_>>(),
+		},
+		"sudo" : { "key" : sudo_key },
+	})
+}
+
+// The initial balances at genesis.
```
peterwht, line 227:
```suggestion
// The initial balances at genesis for local testing only.
```

might consider adding this as well.

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,610 @@
...
+					(session_key, aura_key)
+				})
+				.collect::<Vec<_>>();
+			assert!(initial_collators
+				.iter()
+				.all(|c| session_keys.contains(&(c.0.to_string(), c.1.to_string()))));
+		}
+
+		#[test]
+		fn endows_given_accounts() {
```
peterwht, line 570:
```suggestion
		fn endowed_accounts_are_empty() {
```

peterwht (changes requested):
Left a few nits, should be close to approving

runtime/mainnet/src/genesis.rs
al3mart: Happy to remove from mainnet.

al3mart:
Endowments have been improved in: [1ea9626](https://github.com/r0gue-io/pop-node/pull/473/commits/1ea962642dad3a32b4908d8f0a501df8fc0e527c)

Such that only dev accounts get ENDOWMENT amount and all the rest are only funded with ED.
While it might have been handy to keep sudo mutlisig with more funds than ED, we can easily override this in the zombienet spec. The SS58 account is provided in a test for easy access.

No assets are stores at genesis anymore. The asset key however remain in genesis empty, as we have seen previously that for us to override any state in the zn spec we need the key to exist first.

This should solve all the concerns above.

runtime/mainnet/src/genesis.rs
al3mart: [1ea9626](https://github.com/r0gue-io/pop-node/pull/473/commits/1ea962642dad3a32b4908d8f0a501df8fc0e527c)

runtime/mainnet/src/genesis.rs
al3mart: [1ea9626](https://github.com/r0gue-io/pop-node/pull/473/commits/1ea962642dad3a32b4908d8f0a501df8fc0e527c)

runtime/mainnet/src/genesis.rs
al3mart: [57f4d96](https://github.com/r0gue-io/pop-node/pull/473/commits/57f4d964842e71ee4bc1a70c7ca184d4677dd56a)

runtime/mainnet/src/genesis.rs
al3mart: [1ea9626](https://github.com/r0gue-io/pop-node/pull/473/commits/1ea962642dad3a32b4908d8f0a501df8fc0e527c)

runtime/mainnet/src/genesis.rs
al3mart: [1ea9626](https://github.com/r0gue-io/pop-node/pull/473/commits/1ea962642dad3a32b4908d8f0a501df8fc0e527c)

runtime/mainnet/src/genesis.rs
al3mart: [1ea9626](https://github.com/r0gue-io/pop-node/pull/473/commits/1ea962642dad3a32b4908d8f0a501df8fc0e527c)

runtime/mainnet/src/genesis.rs
```diff
@@ -224,9 +226,21 @@ fn genesis(
 	})
 }
 
-// The initial balances at genesis.
+// The initial balances at genesis; Used for local testing only.
```
Daanvdplas, line 186: // The initial balances at genesis. Used for local testing only.

runtime/mainnet/src/genesis.rs
```diff
@@ -29,7 +29,7 @@ const PRESETS: [&str; 3] = [MAINNET_DEV, MAINNET_LOCAL, MAINNET];
 /// The parachain identifier to set in genesis config.
 pub const PARA_ID: ParaId = ParaId::new(3395);
 
-/// Initial balance for genesis endowed accounts.
+/// Initial balance for genesis endowed accounts; Used for local testing only.
```
Daanvdplas, line 32: /// Initial balance for genesis endowed accounts. Used for local testing only.

Daanvdplas (approved):
LGTM! Do we need to ensure that the genesis hash is the same as the one already registered on chain?

al3mart:
> Do we need to ensure that the genesis hash is the same as the one already registered on chain?

No, it won't be.
We can use all this work for the testnet mirroring mainnet.

runtime/mainnet/src/genesis.rs
```diff
@@ -0,0 +1,559 @@
...
+						SessionKeys { aura},// session keys
+					)
+				})
+				.collect::<Vec<_>>(),
+		},
+		"sudo" : { "key" : sudo_key },
+	})
+}
+
+// The initial balances at genesis; Used for local testing only.
```
al3mart, line 186:
```suggestion
// The initial balances at genesis. Used for local testing only.
```

PR #474, opened 2025-02-18
spike foreign fungibles v2

Daanvdplas:
(Close annotations in the `Files changed` section by clicking the right corner 3 dots when reviewing because I only pushed the relevant bits)

Shows how we can implement the foreign fungibles functionality by creating a wrapper type in the fungibles pallet & api. This reduces the amount of duplicate code but adds some complexity to the pallet (i.e. benchmarking, types).

See latest commit to get a feel for the contract. This is just the fungibles contract updated and can both be used for the fungibles as foreign fungibles use case.

Fungibles contract without `foreign_fungibles` feature enabled: **14.5K** (see latest commit), disabled: **22.8K** (fungibles contract was **14.2K**)

Integrations tests for `transfer` and `balance_of` are working.

### DIFF
--- a/pop-api/Cargo.toml
+++ b/pop-api/Cargo.toml
@@ -33,6 +33,7 @@ path = "src/lib.rs"
 
 [features]
 default = [ "std" ]
+foreign_fungibles = [  ]
 fungibles = [  ]
 nonfungibles = [  ]
 std = [
--- a/pop-api/integration-tests/contracts/fungibles/Cargo.toml
+++ b/pop-api/integration-tests/contracts/fungibles/Cargo.toml
@@ -8,7 +8,7 @@ version = "0.1.0"
 
 [dependencies]
 ink = { version = "5.1.0", default-features = false }
-pop-api = { path = "../../../../pop-api", default-features = false, features = [ "fungibles" ] }
+pop-api = { path = "../../../../pop-api", default-features = false, features = [ "foreign_fungibles", "fungibles" ] }
 
 [lib]
 path = "lib.rs"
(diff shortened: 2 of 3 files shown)


pallets/api/src/fungibles/mod.rs
```diff
@@ -499,7 +594,7 @@ pub mod pallet {
 		fn weight(request: &Self::Read) -> Weight {
 			use Read::*;
 			match request {
-				TotalSupply(_) => <T as Config>::WeightInfo::total_supply(),
+				TotalSupply(id) => <T as Config>::WeightInfo::total_supply(weight_param(id)),
```
Daanvdplas, line 597: This hasn't been tested but applied to show how the weight can be handled for the read functions.

pallets/api/src/fungibles/benchmarking.rs
```diff
@@ -108,10 +108,14 @@ mod benchmarks {
 
 	#[benchmark]
 	// Storage: `Assets`
-	fn total_supply() {
+	fn total_supply(a: Linear<0, 1>) {
```
Daanvdplas, line 111: Not tested but I was thinking to implement it this way if we went this route.

pallets/api/src/fungibles/mod.rs
```diff
@@ -42,13 +80,44 @@ pub mod pallet {
...
 		/// Because this pallet emits events, it depends on the runtime's definition of an event.
 		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
-		/// The instance of pallet-assets.
-		type AssetsInstance;
-		/// Weight information for dispatchables in this pallet.
+		/// The asset instance for trust‑backed assets (renamed from the old AssetsInstance).
+		type TrustBackedAssetsInstance;
+		/// The asset instance for foreign assets.
+		type ForeignAssetsInstance;
+		type Balance: Parameter
```
Daanvdplas, line 99: Complex Balance type unfortunately

PR #475, opened 2025-02-18
spike foreign fungibles

Daanvdplas:
(Close annotations in the `Files changed` section by clicking the right corner 3 dots when reviewing because I only pushed the relevant bits)

Shows how we can implement the foreign fungibles functionality by having a separate instance of the fungibles pallet. This adds duplicate code to the pop api crate but reduces the added complexity in the fungibles pallet.

See latest commit to get a feel for the contract.

fungibles contract: **14.2K**
foreign fungibles contract: **22.8K**

Integrations tests for transfer and balance_of are working.


pop-api/src/v0/foreign_fungibles.rs
```diff
@@ -0,0 +1,243 @@
...
+pub type TokenId = xcm::Location;
+
+/// Transfers `value` amount of tokens from the caller's account to account `to`.
+///
+/// # Parameters
+/// - `token` - The token to transfer.
+/// - `to` - The recipient account.
+/// - `value` - The number of tokens to transfer.
+#[inline]
+pub fn transfer(token: TokenId, to: AccountId, value: Balance) -> Result<()> {
```
Daanvdplas, line 15: Only added `transfer` and `balance_of` to make the integrations tests work. This will be a duplication of the functions in the `fungibles` module.

PR #479, opened 2025-02-24
feat(mainnet): include xcm benchmarks

al3mart:
Needed for: #478

Adds benchmarks related to XCM in mainnet runtime.

- [x] `pallet_xcm` extrinsics benchmark
- [x] `pallet_xcm_benchmarks::generic` benchmark
- [x] `pallet_xcm_benchmarks::fungible` benchmark
- [x] Configure `Weigher` using `WeightInfoBounds`
- [x] Remove default weights from `pallet_xcm` and use benchmarks results.

Note that the benchmarks have been run locally, and they need to be re-run in the collator machines. A script will be provided in #478 such that we can do that easily.


runtime/mainnet/src/weights/xcm/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+
+use crate::{config::xcm::MaxAssetsIntoHolding, Runtime};
+
+trait WeighAssets {
+	fn weigh_assets(&self, weight: Weight) -> Weight;
+}
+
+const MAX_ASSETS: u64 = 100;
+
+impl WeighAssets for AssetFilter {
```
al3mart, line 22: Implemented this based on: https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/asset-hubs/asset-hub-polkadot/src/weights/xcm/mod.rs

Might need further consideration.

runtime/mainnet/src/weights/xcm/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+	fn weigh_assets(&self, weight: Weight) -> Weight {
+		match self {
+			Self::Definite(assets) => weight.saturating_mul(assets.inner().iter().count() as u64),
+			Self::Wild(asset) => match asset {
+				All => weight.saturating_mul(MAX_ASSETS),
+				AllOf { fun, .. } => match fun {
+					WildFungibility::Fungible => weight,
+					// Magic number 2 has to do with the fact that we could have up to 2 times
+					// MaxAssetsIntoHolding in the worst-case scenario.
```
al3mart, line 31: Reason for this is:

> // worst-case, holding.len becomes 2 * holding_limit.
> // this guarantees that if holding.len() == holding_limit and you have more than
> // `holding_limit` items (which has a best case outcome of holding.len() == holding_limit),
> // then the operation is guaranteed to succeed.

[source](https://github.com/paritytech/polkadot-sdk/blob/4a400dc1866f11707331fb6408df1055d0f42a70/polkadot/xcm/xcm-executor/src/lib.rs#L470-L473)

Also, even though I've found this comment: https://github.com/paritytech/cumulus/pull/2102#discussion_r1100706750
The reason why this only is applied for `NonFungible` is not that apparent to me. It seems that it would apply to both cases.

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +67,183 @@ frame_benchmarking::define_benchmarks!(
...
+
+		let who = frame_benchmarking::whitelisted_caller();
+		// Give some multiple of the existential deposit.
+		let balance = ExistentialDeposit::get() * 10_000;
+		let _ =
+			<Balances as frame_support::traits::Currency<_>>::make_free_balance_be(&who, balance);
+		let ah_on_pop: AccountId = LocationToAccountId::convert_location(&AssetHub::get())?;
+		let _ = <Balances as frame_support::traits::Currency<_>>::make_free_balance_be(
+			&ah_on_pop, balance,
+		);
```
al3mart, line 110: Funding AH sovereign account on pop because the benchmark tries to [withdraw from destination](https://github.com/paritytech/polkadot-sdk/blob/4a400dc1866f11707331fb6408df1055d0f42a70/polkadot/xcm/pallet-xcm/src/benchmarking.rs#L254) as a way to verify the benchmark.

I believe that makes sense for reserve transfers happening within the reserve itself. Not quite that for our case.

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +67,183 @@ frame_benchmarking::define_benchmarks!(
...
+	fn alias_origin() -> Result<(Location, Location), BenchmarkError> {
+		// Pop's `Aliasers` is configured to `Nothing`.
+		Err(BenchmarkError::Skip)
+	}
+}
+
+parameter_types! {
+	pub RelayAsset: Asset = Asset::from((RelayLocation::get(), 1 * UNIT));
+	pub TrustedReserve: Option<(Location, Asset)> = Some((AssetHub::get(), RelayAsset::get()));
+	// We don't set any trusted teleporters in our XCM config, but we need this for the benchmarks.
```
peterwht, line 234: ❤️

runtime/mainnet/src/weights/pallet_xcm.rs
```diff
@@ -0,0 +1,318 @@
...
+			.saturating_add(T::DbWeight::get().writes(1))
+	}
+	/// Storage: `Benchmark::Override` (r:0 w:0)
+	/// Proof: `Benchmark::Override` (`max_values`: None, `max_size`: None, mode: `Measured`)
+	fn teleport_assets() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `0`
+		//  Estimated: `0`
+		// Minimum execution time: 18_446_744_073_709_551_000 picoseconds.
+		Weight::from_parts(18_446_744_073_709_551_000, 0)
```
peterwht, line 54: For future reviewers: this value is so large because we don't support teleporting.

runtime/mainnet/src/weights/xcm/mod.rs
peterwht: can we include that link the in the source code please

runtime/mainnet/src/weights/xcm/mod.rs
peterwht: Worth pinging a Cisco or someone about?

runtime/mainnet/src/weights/xcm/pallet_xcm_benchmarks_generic.rs
```diff
@@ -0,0 +1,335 @@
...
+	/// Storage: `ParachainSystem::HostConfiguration` (r:1 w:0)
+	/// Proof: `ParachainSystem::HostConfiguration` (`max_values`: Some(1), `max_size`: None, mode: `Measured`)
+	/// Storage: `ParachainSystem::PendingUpwardMessages` (r:1 w:1)
+	/// Proof: `ParachainSystem::PendingUpwardMessages` (`max_values`: Some(1), `max_size`: None, mode: `Measured`)
+	pub(crate) fn report_holding() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `111`
+		//  Estimated: `1596`
+		// Minimum execution time: 291_401_000 picoseconds.
+		Weight::from_parts(291_401_000, 0)
```
peterwht, line 45: A little expensive compared to westend: https://github.com/paritytech/polkadot-sdk/blob/3dc3a11cd68762c2e5feb0beba0b61f448c4fc92/polkadot/runtime/westend/src/weights/xcm/pallet_xcm_benchmarks_generic.rs#L72

Likely just hardware, but double checking regardless.

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +67,183 @@ frame_benchmarking::define_benchmarks!(
...
+	pub ExistentialDepositAsset: Option<Asset> = Some((
+		RelayLocation::get(),
+		ExistentialDeposit::get()
+	).into());
+	pub const AssetHubParaId: ParaId = ParaId::new(1000);
+}
+
+impl pallet_xcm::benchmarking::Config for Runtime {
+	type DeliveryHelper = (
+		polkadot_runtime_common::xcm_sender::ToParachainDeliveryHelper<
```
peterwht, line 81: should this also have a `ToParentDelivery`?

https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L1128

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +67,183 @@ frame_benchmarking::define_benchmarks!(
...
+		))
+	}
+
+	fn set_up_complex_asset_transfer(
+	) -> Option<(xcm::prelude::Assets, u32, Location, Box<dyn FnOnce()>)> {
+		ParachainSystem::open_outbound_hrmp_channel_for_benchmarks_or_tests(ParaId::from(
+			AssetHubParaId::get(),
+		));
+		// Pop can only reserve transfer DOT.
+		// This test needs to be adapted as the features grow.
```
peterwht, line 125: could you please add a comment on one of the XCM config unit tests?

Basically that way we have this flow:
1.  we change XCM to support other asset transfers 
2. We now have a failing unit test
3. We fix unit test and see comment that says: "update set_up_complex_asset_transfer ..."

peterwht (changes requested):
Very clean PR, almost ready to approve.

runtime/mainnet/src/benchmarks.rs
al3mart: makes sense, thanks!

runtime/mainnet/src/weights/xcm/pallet_xcm_benchmarks_generic.rs
al3mart: Should be, I've did these benchmarks in my laptop :)
Recreating weights in our collators will give us a better point to compare.

runtime/mainnet/src/weights/pallet_xcm.rs
al3mart: Thanks!

Indeed, max weight is applied for unsupported actions.

runtime/mainnet/src/weights/xcm/mod.rs
al3mart: [fe79c4b](https://github.com/r0gue-io/pop-node/pull/479/commits/fe79c4b7fcabdf6c818b513ac03060da22adf681)

runtime/mainnet/src/benchmarks.rs
al3mart: [25a3bf4](https://github.com/r0gue-io/pop-node/pull/479/commits/25a3bf42739f1fdbddf37ce5cc808231316b897e)

runtime/mainnet/src/benchmarks.rs
al3mart: It will be better with it :+1:  [375e9d1](https://github.com/r0gue-io/pop-node/pull/479/commits/375e9d194d8a69a2e2ca9194c1594d189ea703e0)

runtime/mainnet/src/weights/xcm/mod.rs
al3mart: Clarified in: [0a50386](https://github.com/r0gue-io/pop-node/pull/479/commits/0a5038643b8521466f79fd261800133a00800c01)

peterwht (approved):
Great job. Approving!

peterwht:
do we need cumulus_pallet_weight_reclaim : https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/assets/asset-hub-westend/src/lib.rs#L1486?

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +70,184 @@ frame_benchmarking::define_benchmarks!(
...
+	}
+
+	fn alias_origin() -> Result<(Location, Location), BenchmarkError> {
+		// Pop's `Aliasers` is configured to `Nothing`.
+		Err(BenchmarkError::Skip)
+	}
+}
+
+parameter_types! {
+	pub RelayAsset: Asset = Asset::from((RelayLocation::get(), 1 * UNIT));
```
Daanvdplas, line 236: Consistancy:
```suggestion
	pub RelayAsset: Asset = Asset::from((Location::parent(), 1 * UNIT));
```

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +70,184 @@ frame_benchmarking::define_benchmarks!(
 	[pallet_nfts, Nfts]
 	[pallet_assets, TrustBackedAssets]
 );
+
+parameter_types! {
+	pub ExistentialDepositAsset: Option<Asset> = Some((
+		RelayLocation::get(),
+		ExistentialDeposit::get()
```
Daanvdplas, line 77: Some docs would be good. For example mentioning that the ExistentialDeposit can be used from mainnet config because we have the same existential deposit.

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +70,184 @@ frame_benchmarking::define_benchmarks!(
...
+		let _ =
+			<Balances as frame_support::traits::Currency<_>>::make_free_balance_be(&who, balance);
+		let ah_on_pop: AccountId = LocationToAccountId::convert_location(&AssetHub::get())?;
+		let _ = <Balances as frame_support::traits::Currency<_>>::make_free_balance_be(
+			&ah_on_pop, balance,
+		);
+
+		// We can do reserve transfers of relay native asset to AH.
+		Some((
+			Asset { fun: Fungible(ExistentialDeposit::get()), id: AssetId(Location::from(Parent)) },
```
Daanvdplas, line 124: You have a lot of ways you are writing the same thing. Could you please choose a way and be consistent :)

```suggestion
			Asset { fun: Fungible(ExistentialDeposit::get()), id: AssetId(Location::parent()) },
```

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +70,184 @@ frame_benchmarking::define_benchmarks!(
...
+		Err(BenchmarkError::Skip)
+	}
+
+	fn alias_origin() -> Result<(Location, Location), BenchmarkError> {
+		// Pop's `Aliasers` is configured to `Nothing`.
+		Err(BenchmarkError::Skip)
+	}
+}
+
+parameter_types! {
```
Daanvdplas, line 235: Perhaps unnecessary?

runtime/mainnet/src/config/xcm.rs
```diff
@@ -1037,6 +1061,45 @@ mod tests {
...
+				let b: u128 = BaseDeliveryFee::get();
+				let m: u128 = TransactionByteFee::get();
+
+				// F * (B + msg_length * M)
+				// A: RelayLocation
+				// B: BaseDeliveryFee
+				// M: TransactionByteFee
+				// F: ParachainSystem
+				//
+				// message_length = 1
```
Daanvdplas, line 1094: Always better to use a bigger value to be sure that it works (e.g. 2).

runtime/mainnet/src/weights/cumulus_pallet_xcmp_queue.rs
```diff
@@ -0,0 +1,136 @@
...
+// Executed Command:
+// ./target/release/pop-node
+// benchmark
+// pallet
+// --chain=./scripts/pop-raw.json
+// --genesis-builder-preset=pop-mainnet
+// --pallet=*
+// --extrinsic=*
+// --steps=2
+// --repeat=1
```
Daanvdplas, line 19: I am used to using 50 steps and 20 repeats. Or should we do this on the actual hardware?

runtime/mainnet/src/weights/cumulus_pallet_xcmp_queue.rs
```diff
@@ -0,0 +1,136 @@
...
+//! HOSTNAME: `tux`, CPU: `12th Gen Intel(R) Core(TM) i7-12700H`
+//! WASM-EXECUTION: `Compiled`, CHAIN: `Some("./scripts/pop-raw.json")`, DB CACHE: 1024
+
+// Executed Command:
+// ./target/release/pop-node
+// benchmark
+// pallet
+// --chain=./scripts/pop-raw.json
+// --genesis-builder-preset=pop-mainnet
+// --pallet=*
```
Daanvdplas, line 16: why not specify the pallet?

runtime/mainnet/src/weights/pallet_message_queue.rs
```diff
@@ -0,0 +1,167 @@
...
+//! HOSTNAME: `tux`, CPU: `12th Gen Intel(R) Core(TM) i7-12700H`
+//! WASM-EXECUTION: `Compiled`, CHAIN: `Some("./scripts/pop-raw.json")`, DB CACHE: 1024
+
+// Executed Command:
+// ./target/release/pop-node
+// benchmark
+// pallet
+// --chain=./scripts/pop-raw.json
+// --genesis-builder-preset=pop-mainnet
+// --pallet=*
```
Daanvdplas, line 16: same

runtime/mainnet/src/weights/pallet_message_queue.rs
```diff
@@ -0,0 +1,167 @@
...
+// --extrinsic=*
+// --steps=2
+// --repeat=1
```
Daanvdplas, line 19: same, should be increased

runtime/mainnet/src/weights/xcm/pallet_xcm_benchmarks_generic.rs
```diff
@@ -0,0 +1,363 @@
...
+//! HOSTNAME: `tux`, CPU: `12th Gen Intel(R) Core(TM) i7-12700H`
+//! WASM-EXECUTION: `Compiled`, CHAIN: `Some("pop")`, DB CACHE: 1024
+
+// Executed Command:
+// ./target/release/pop-node
+// benchmark
+// pallet
+// --chain=pop
+// --pallet=pallet_xcm_benchmarks::generic
+// --extrinsic=report_holding,unpaid_execution,set_fees_mode,clear_topic,set_topic,clear_transact_status,report_transact_status,expect_pallet,query_pallet,expect_transact_status,expect_error,expect_origin,expect_asset,burn_asset,unsubscribe_version,subscribe_version,trap,claim_asset,report_error,clear_origin,execute_with_origin,descend_origin,clear_error,set_appendix,set_error_handler,refund_surplus,transact,query_response,asset_claimer,pay_fees,buy_execution,report_holding
```
Daanvdplas, line 16: Why not all? I realise now that these weird commands might come from pop cli and perhaps worth forwarding

runtime/mainnet/src/benchmarks.rs
```diff
@@ -1,7 +1,34 @@
-use crate::{config::assets::TrustBackedAssetsInstance, Runtime};
```
Daanvdplas, line 1: The code you wrote here, is there any reference material I can take a look at? I couldn't find for asset hub

Daanvdplas (commented):
Nice job!!!

I must say that I didn't expect this complex review and have to finish it tomorrow. Will need some more reference material to give a proper review. Already left some comments you can take a look at.

runtime/mainnet/src/weights/xcm/pallet_xcm_benchmarks_generic.rs
al3mart: We don't configure certain types within our xcm configuration. Which makes the following benchmarks fail

- pallet_xcm_benchmarks::generic::alias_origin
- pallet_xcm_benchmarks::generic::exchange_asset
- pallet_xcm_benchmarks::generic::export_message
- pallet_xcm_benchmarks::generic::lock_asset
- pallet_xcm_benchmarks::generic::note_unlockable
- pallet_xcm_benchmarks::generic::request_unlock
- pallet_xcm_benchmarks::generic::universal_origin
- pallet_xcm_benchmarks::generic::unlock_asset


There are features that we are not supporting yet.
For these cases, [max weight is used](https://github.com/r0gue-io/pop-node/pull/479/files/0a5038643b8521466f79fd261800133a00800c01?show-viewed-files=true&file-filters%5B%5D=#diff-177869dac5e632df6b7f0346b03a2864ec2da591425313bd42426df8a102161fR236). This is configured in `runtime/mainnet/src/weights/xcm/mod.rs`

runtime/mainnet/src/weights/pallet_message_queue.rs
al3mart: Yeah, the script I have included in the base branch is configured to run 50 steps and 20 repetitions.
This is the script to be run in the collators.

What I have pushed here was run in my local machine, so it is no representative and chose values that allowed me to have the files ready such that they could be substituted after a proper bench run.

runtime/mainnet/src/weights/cumulus_pallet_xcmp_queue.rs
al3mart: The script I've prepared for the collator will execute the benchmark per pallet, the resulting file should contain the name of the pallet.

These files are the result of a general run. I executed all benchmarks for all pallets.

runtime/mainnet/src/weights/cumulus_pallet_xcmp_queue.rs
al3mart: Yeah, that is a better config once when we run the benchmarks in the proper hw. 
A script correctly configured is provided in: https://github.com/r0gue-io/pop-node/pull/478/files?show-viewed-files=true&file-filters%5B%5D=#diff-30a6df322b49f199f5f3d540d92036730f6a9c83ad89909408d328e99759ece8

runtime/mainnet/src/benchmarks.rs
al3mart: You can find `pallet_xcm::benchmarking` implementation for asset_hub [here](https://github.com/polkadot-fellows/runtimes/blob/80ee7dce11ed68a68abf89512bd9886ffcabe192/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L1126).

And `pallet_xcm_benchmarks` [here](https://github.com/polkadot-fellows/runtimes/blob/80ee7dce11ed68a68abf89512bd9886ffcabe192/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L1214)

runtime/mainnet/src/apis.rs
```diff
@@ -505,3 +490,18 @@ impl_runtime_apis! {
...
+};
+
+// Local module imports
+use super::{
+	config::{monetary::fee::WeightToFee, system::RuntimeBlockWeights, xcm as xcm_config},
+	AccountId, Balance, Balances, Block, BlockNumber, BlockWeights, EventRecord, Executive,
+	ExtrinsicInclusionMode, InherentDataExt, Nfts, Nonce, OriginCaller, ParachainSystem,
+	PolkadotXcm, Revive, Runtime, RuntimeCall, RuntimeEvent, RuntimeGenesisConfig, RuntimeOrigin,
+	SessionKeys, System, TransactionPayment, UncheckedExtrinsic, VERSION,
+};
```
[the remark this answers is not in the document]
al3mart: Back where they should be :+1: 
[006c221](https://github.com/r0gue-io/pop-node/pull/479/commits/006c22172b4a65d848da88ed0a01c050445eb9f0)

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +70,184 @@ frame_benchmarking::define_benchmarks!(
...
+		RelayLocation::get(),
+		ExistentialDeposit::get()
+	).into());
+	pub const AssetHubParaId: ParaId = ParaId::new(1000);
+}
+
+type DeliveryHelper = (
+	cumulus_primitives_utility::ToParentDeliveryHelper<
+		XcmConfig,
+		ExistentialDepositAsset,
```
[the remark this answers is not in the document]
al3mart: These ED should be Pop's ED, as the delivery fees are paid locally.
The benchmarks will deposit this ED into a test account in our runtime.

runtime/mainnet/src/benchmarks.rs
al3mart: Kinda helps keeping `pallet_xcm_benchmarks::fungible` impl a bit cleaner, but happy to receive suggestions.

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +70,184 @@ frame_benchmarking::define_benchmarks!(
...
+			<Balances as frame_support::traits::Currency<_>>::make_free_balance_be(&who, balance);
+		let ah_on_pop: AccountId = LocationToAccountId::convert_location(&AssetHub::get())?;
+		let _ = <Balances as frame_support::traits::Currency<_>>::make_free_balance_be(
+			&ah_on_pop, balance,
+		);
+
+		// We can do reserve transfers of relay native asset to AH.
+		Some((
+			Asset { fun: Fungible(ExistentialDeposit::get()), id: AssetId(Location::from(Parent)) },
+			ParentThen(Parachain(1000).into()).into(),
```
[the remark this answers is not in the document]
al3mart: All locations are now defined in terms of defined types in xcm config file
[dbf83e7](https://github.com/r0gue-io/pop-node/pull/479/commits/dbf83e7915b2a16913e9ea23cf25a324399e5905)

runtime/mainnet/src/benchmarks.rs
al3mart: All locations are now defined in terms of defined types in xcm config file
[dbf83e7](https://github.com/r0gue-io/pop-node/pull/479/commits/dbf83e7915b2a16913e9ea23cf25a324399e5905)

runtime/mainnet/src/config/xcm.rs
al3mart: Don't argue that this could be a good practice, but as the math seems to be mathing I've ensured that the empty message encodes to a size of `1`.
[30c11f2](https://github.com/r0gue-io/pop-node/pull/479/commits/30c11f28413bac553eade30b51f79d6c5783582f)

runtime/mainnet/src/benchmarks.rs
```diff
@@ -1,7 +1,34 @@
...
+use frame_support::parameter_types;
+pub use pallet_xcm::benchmarking::Pallet as PalletXcmBenchmark;
+use xcm::prelude::{
+	Asset, AssetId, Fungible, Here, InteriorLocation, Junction, Location, NetworkId, Parachain,
+	Parent, ParentThen, Response,
+};
+use xcm_executor::traits::ConvertLocation;
+
+use super::*;
+use crate::{
```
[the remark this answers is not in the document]
al3mart: Included in [664d5d9](https://github.com/r0gue-io/pop-node/pull/479/commits/664d5d90786b454a34f15623d8c2db5bf986bc78)

runtime/mainnet/src/benchmarks.rs
```diff
@@ -1,7 +1,34 @@
...
+			AssetHub, LocationToAccountId, PriceForParentDelivery, PriceForSiblingDelivery,
+			RelayLocation, XcmConfig,
+		},
+	},
+	Runtime,
+};
 
 /// Instance of `pallet_assets` used for benchmarks of trust backed assets.
 pub type TrustBackedAssets = pallet_assets::Pallet<Runtime, TrustBackedAssetsInstance>;
+/// Pallet that serves no other purpose than benchmarking raw XCMs.
```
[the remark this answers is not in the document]
al3mart: docs improved in: [dfe6212](https://github.com/r0gue-io/pop-node/pull/479/commits/dfe62120e6a1d6d6912c85b3b0c02601dd0f14b3)

runtime/mainnet/src/benchmarks.rs
Daanvdplas: I think removing the RelayAsset type would be a good midway. The relay asset is used multiple times but this type has a specific amount to it but is not understood from the name. It created confusion so probably best to remove. It is also only used once so not a big deal to remove

runtime/mainnet/src/weights/pallet_xcm.rs
```diff
@@ -0,0 +1,338 @@
...
+	/// Storage: `PolkadotXcm::ShouldRecordXcm` (r:1 w:0)
+	/// Proof: `PolkadotXcm::ShouldRecordXcm` (`max_values`: Some(1), `max_size`: None, mode: `Measured`)
+	fn execute() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `0`
+		//  Estimated: `1485`
+		// Minimum execution time: 22_480_000 picoseconds.
+		Weight::from_parts(22_480_000, 0)
+			.saturating_add(Weight::from_parts(0, 1485))
+			.saturating_add(T::DbWeight::get().reads(1))
```
Daanvdplas, line 122: One read more than asset hub

Daanvdplas: This file has more differences, do we know why?

runtime/mainnet/src/weights/xcm/pallet_xcm_benchmarks_generic.rs
Daanvdplas: Link doesn't work. But, I think we can still run those benchmarks, then the weight function will just return max weight.

Daanvdplas:
Don't forget to rebase before next review :)

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +68,184 @@ frame_benchmarking::define_benchmarks!(
...
+	pub RelayAsset: Asset = Asset::from((RelayLocation::get(), 1 * UNIT));
+	pub TrustedReserve: Option<(Location, Asset)> = Some((AssetHub::get(), RelayAsset::get()));
+	// We don't set any trusted teleporters in our XCM config, but we need this for the benchmarks.
+	pub TrustedTeleporter: Option<(Location, Asset)> = Some((
+					AssetHub::get(),
+					Asset { fun: Fungible(1 * UNIT), id: AssetId(RelayLocation::get()) },
+				));
+}
+impl pallet_xcm_benchmarks::fungible::Config for Runtime {
+	type CheckedAccount = ();
```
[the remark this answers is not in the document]
al3mart: Correct

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +68,184 @@ frame_benchmarking::define_benchmarks!(
...
+	}
+}
+
+parameter_types! {
+	pub RelayAsset: Asset = Asset::from((RelayLocation::get(), 1 * UNIT));
+	pub TrustedReserve: Option<(Location, Asset)> = Some((AssetHub::get(), RelayAsset::get()));
+	// We don't set any trusted teleporters in our XCM config, but we need this for the benchmarks.
+	pub TrustedTeleporter: Option<(Location, Asset)> = Some((
+					AssetHub::get(),
+					Asset { fun: Fungible(1 * UNIT), id: AssetId(RelayLocation::get()) },
```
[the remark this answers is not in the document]
al3mart: [4a7b8c4](https://github.com/r0gue-io/pop-node/pull/479/commits/4a7b8c44a8d59d28f2b11ef41ad373b2a34871cc)

runtime/mainnet/src/benchmarks.rs
al3mart: gone [4a7b8c4](https://github.com/r0gue-io/pop-node/pull/479/commits/4a7b8c44a8d59d28f2b11ef41ad373b2a34871cc)

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +70,184 @@ frame_benchmarking::define_benchmarks!(
...
+
+	fn worst_case_asset_exchange(
+	) -> Result<(xcm::prelude::Assets, xcm::prelude::Assets), BenchmarkError> {
+		// Pop doesn't support asset exchange for now.
+		Err(BenchmarkError::Skip)
+	}
+
+	fn universal_alias() -> Result<(Location, Junction), BenchmarkError> {
+		// Pop's `UniversalAliases` is configured to `Nothing`.
+		Err(BenchmarkError::Skip)
```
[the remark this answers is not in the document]
al3mart: I believe the key difference is that Skip won't stop the benchmark of other extrinsics even though this one errs.
Which seems handy.

runtime/mainnet/src/apis.rs
```diff
@@ -285,7 +285,6 @@ impl_runtime_apis! {
 			use frame_system_benchmarking::Pallet as SystemBench;
 			use frame_system_benchmarking::extensions::Pallet as SystemExtensionsBench;
 
-
 			impl frame_system_benchmarking::Config for Runtime {
```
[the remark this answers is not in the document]
al3mart: Will check, but let's keep this PR focused on the XCM benches

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +70,184 @@ frame_benchmarking::define_benchmarks!(
...
+		vec![Asset { id: AssetId(Location::parent()), fun: Fungible(u128::MAX) }].into()
+	}
+}
+
+impl pallet_xcm_benchmarks::generic::Config for Runtime {
+	type RuntimeCall = RuntimeCall;
+	type TransactAsset = Balances;
+
+	fn worst_case_response() -> (u64, Response) {
+		(0u64, Response::Version(Default::default()))
```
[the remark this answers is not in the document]
al3mart: I've improved this in: [a648260](https://github.com/r0gue-io/pop-node/pull/479/commits/a648260509d96fa6378e7bf10fdea1b89c37f596)

runtime/mainnet/src/benchmarks.rs
al3mart: [8a9150e](https://github.com/r0gue-io/pop-node/pull/479/commits/8a9150ee72996ba20cde7f778a6c208aa96af3e0)

runtime/mainnet/src/weights/pallet_xcm.rs
al3mart: Might be because of our MaintenanceAccount. But is just a guess.

runtime/mainnet/src/apis.rs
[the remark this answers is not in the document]
Daanvdplas: Broken link

runtime/mainnet/src/benchmarks.rs
```diff
@@ -37,3 +65,186 @@ frame_benchmarking::define_benchmarks!(
...
+		Some((
+			Asset { fun: Fungible(ExistentialDeposit::get()), id: AssetId(RelayLocation::get()) },
+			AssetHub::get(),
+		))
+	}
+
+	fn set_up_complex_asset_transfer(
+	) -> Option<(xcm::prelude::Assets, u32, Location, Box<dyn FnOnce()>)> {
+		ParachainSystem::open_outbound_hrmp_channel_for_benchmarks_or_tests(ParaId::from(
+			AssetHubParaId::get(),
```
Daanvdplas, line 128:
```suggestion
		ParachainSystem::open_outbound_hrmp_channel_for_benchmarks_or_tests(
			AssetHubParaId::get(),
```

runtime/mainnet/src/benchmarks.rs
```diff
@@ -42,3 +68,184 @@ frame_benchmarking::define_benchmarks!(
...
+	}
+
+	fn get_asset() -> Asset {
+		Asset { id: AssetId(RelayLocation::get()), fun: Fungible(ExistentialDeposit::get()) }
+	}
+}
+
+impl pallet_xcm_benchmarks::Config for Runtime {
+	type AccountIdConverter = LocationToAccountId;
+	type DeliveryHelper = DeliveryHelper;
```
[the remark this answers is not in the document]
Daanvdplas: Can we ask this specific thing?

runtime/mainnet/src/benchmarks.rs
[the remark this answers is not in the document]
Daanvdplas: Asset hub uses it so seems to be fine.

runtime/mainnet/src/benchmarks.rs
```diff
@@ -37,3 +65,186 @@ frame_benchmarking::define_benchmarks!(
...
+	}
+}
+
+impl pallet_xcm_benchmarks::generic::Config for Runtime {
+	type RuntimeCall = RuntimeCall;
+	type TransactAsset = Balances;
+
+	fn worst_case_response() -> (u64, Response) {
+		let notify = frame_system::Call::remark { remark: vec![] };
+		PolkadotXcm::new_notify_query(Location::here(), notify, 10, Location::here());
```
Daanvdplas, line 182: Nothing is done with this?

runtime/mainnet/src/config/xcm.rs
```diff
@@ -240,12 +241,18 @@ impl pallet_xcm::Config for Runtime {
...
+	type WeightInfo = weights::pallet_xcm::WeightInfo<Runtime>;
 	type XcmExecuteFilter = Everything;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
+	// Benchmark `set_up_complex_asset_transfer` should be revisited as more assets are included.
+	// Benchmark `worst_case_holding` should reflect all possible assets that can be transferred.
 	type XcmReserveTransferFilter = FilterByAssets<Equals<RelayLocation>>;
 	type XcmRouter = XcmRouter;
+	// Benchmark `set_up_complex_asset_transfer` needs to be updated when asset teleporting is
+	// introduced.
+	// Benchmark `worst_case_holding` should also be updates consequently.
```
Daanvdplas, line 255: Perhaps a comment on `set_up_complex_asset_transfer` explaining that when a new asset is added that `worst_case_handling should be updated, in stead of two places here

runtime/mainnet/src/config/xcm.rs
Daanvdplas: So if this value is not used we wouldn't be sure, that is more what I'm trying to say.

runtime/mainnet/src/weights/xcm/mod.rs
```diff
@@ -0,0 +1,316 @@
...
+					// As per: https://github.com/paritytech/polkadot-sdk/blob/4a400dc1866f11707331fb6408df1055d0f42a70/polkadot/xcm/xcm-executor/src/lib.rs#L470-L473
+					// Only applicable to `NonFungible` because `AllOf` here means all the assets of
+					// a collection. Whereas for `Fungible` `AllOf` is only one asset.
+					WildFungibility::NonFungible => weight.saturating_mul((MAX_ASSETS * 2) as u64),
```
Daanvdplas, line 37: clippy:
```suggestion
					WildFungibility::NonFungible => weight.saturating_mul(MAX_ASSETS * 2),
```

runtime/mainnet/src/weights/xcm/pallet_xcm_benchmarks_generic.rs
Daanvdplas: So should we just run all?

Daanvdplas (commented):
A few last comments

peterwht (approved):
Approving again. 
- The code to generate the benchmarks are solid
- The outputted weight files discrepancies make sense to me as well due to hardware, number of steps, and different XCM configs. So, until we run on production collators this will be sufficient.

runtime/mainnet/src/benchmarks.rs
al3mart: Populates the state with a query, such that we can hit the worst case scenario in `query_response` bench.
(Results did show higher weights after this addition)

runtime/mainnet/src/config/xcm.rs
al3mart: I kinda prefer to have the surfaced all the way there, easier to discover and more difficult to miss.

runtime/mainnet/src/weights/xcm/mod.rs
al3mart: Thanks!

runtime/mainnet/src/benchmarks.rs
al3mart: [7cb4c2a](https://github.com/r0gue-io/pop-node/pull/479/commits/7cb4c2a8ef8daa6c64ad820af7de931f905c13a5)

runtime/mainnet/src/weights/xcm/pallet_xcm_benchmarks_generic.rs
al3mart: Unfortunately running benchmarks for the above ext won't allow the bench cli to generate a result. So we do generate weights for every extrinsic we can benchmark, and then we set maximum weight for those ext we don't support - https://github.com/r0gue-io/pop-node/pull/479/files#diff-177869dac5e632df6b7f0346b03a2864ec2da591425313bd42426df8a102161fR131

runtime/mainnet/src/config/xcm.rs
al3mart: I understand what you meant :)
I also think that testing `saturating_mul` is not that much of this test scope.

Have pushed an arbitrary msg with length diff than 1 anyway: [f458b53](https://github.com/r0gue-io/pop-node/pull/479/commits/f458b5374350166181cc077433f1285e1effc52e)

Daanvdplas (approved):
Well done on this PR, very big and complex undertaking and you helped me understand everything very well!

PR #452, opened 2025-01-27
feat(mainnet): add pallet revive

peterwht:
Adds pallet revive to the mainnet runtime. 

Please review individual commits for specific components added (pallet, apis, tests).

Supersedes https://github.com/r0gue-io/pop-node/pull/427

[sc-2219]

PR #448, opened 2025-01-23
feat(mainnet): add gov: collective, membership, motion

peterwht:
Adds a Council based governance that:

- Has NO elections (selected councillors, to start)
- Han NO membership pallet.
- Uses pallet-motion to allow council to use Root origin.

---

- Relevant configuration:

- `AtLeastThreeFourthsOfCouncil`: at least 3/4 * total seats are aye votes.
- `UnanimousCouncilVote`: 100% of the votes are aye votes.

- `pallet_collective`

```rust
type DefaultVote = MoreThanMajorityThenPrimeDefaultVote;
type DisapproveOrigin = EnsureRoot<AccountId>;
type KillOrigin = EnsureRoot<AccountId>;
type SetMembersOrigin = EnsureRoot<AccountId>; // <-- This is the only way to modify the members of council.
```

- `pallet_motion`

This pallet is how for the council achieves dispatch with Root origin.

```rust
// Simple majority is disabled.
type SimpleMajorityOrigin = NeverEnsureOrigin<()>;
// At least 3/4 of the council vote is needed.
type SuperMajorityOrigin = AtLeastThreeFourthsOfCouncil;
// A unanimous council vote is needed.
type UnanimousOrigin = UnanimousCouncilVote;
```

To familiarize yourself with this governance mechanism, please use the following guide.

### Goal Checklist
- [ ] Create a Motion that requires Root
    - [ ] Vote on Motion
    - [ ] Execute Motion

### Create a Motion that requires Root
1. Launch chain `pop up parachain -f networks/mainnet.toml`
2. Navigate to [Council page](https://polkadot.js.org/apps/?rpc=ws://127.0.0.1:9944#/council) and note Members
3. Navigate to [sub-page for Motions](https://polkadot.js.org/apps/?rpc=ws://127.0.0.1:9944#/council/motions)
4. Press "Propose Motion"
5. Min threshold is currently 1/2 + 1. I recommend setting threshold to `4` (unanimous)
6. Set the proposal call to `motion.unanimous`, providing Root origin
7. Set the sub-call to something that requires Root. One recommendation is to use `balances.forceSetBalance`
8. You will see a `council.Proposed` event
9. On the [Motions page](https://polkadot.js.org/apps/?rpc=ws://127.0.0.1:9944#/council/motions) you will see a new Motion available for vote, along with proposal info
10. Press the `Vote` button and vote AYE. Repeat this with 4 dev accounts
[...]

PR #440, opened 2025-01-21
refactor(mainnet): monetary config

al3mart:
Creates **monetary** config module including:
- `pallet_balances`
- `pallet_transaction_payment`
- `fee` constants

Notable configuration items are:

- deposit

Mainnet to use its own deposit function different from the one defined in `pop_runtime_common` being used in devnet and testnet. It aligns the deposit pricing with system chains in Polkadot. 
```rust
pub const fn deposit(items: u32, bytes: u32) -> Balance {
	// src: https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/constants/src/polkadot.rs#L70
	(items as Balance * 20 * UNIT + (bytes as Balance) * 100 * fee::MILLI_CENTS) / 100
}
```

---

- Weight to fee

Maps `ExtrinsicBaseWeight::get().ref_time()` to 1/200 CENT.
Maps 20kb proof to 1 CENT.

---

- Existential deposit

`1_000_000_000` units.

---

- `pallet_balances`

```rust
type MaxLocks = ConstU32<0>;
type MaxReserves = ConstU32<0>;
type ReserveIdentifier = ();
```
Configured this way as these types are deprecated: https://github.com/paritytech/substrate/pull/12951

---

- `pallet_transaction_payment`

`OnChargeTransaction` needs to align with the result of [sc-2260]
```rust
type OnChargeTransaction =
		pallet_transaction_payment::FungibleAdapter<Balances, ResolveTo<SudoAddress, Balances>>;
```

[sc-2205]

PR #428, opened 2025-01-14
feat(mainnet): Add treasury pallet and handle fees

f-gate:
Have added the treasury pallet with reasonable parameters and have taken inspiration from polkadot and rococo for these. The only one really of concern is `MaxSpend` i have left it as u32::MAX.

The `DealWithFees` simply takes both the fees and tips and splits equally between the newly created `TreasuryAccount` and `MaintenenceAccount`. 

In the future i will split the tasks down into 2 PRs as Peter and i discussed yesterday.

[sc-2260]
[sc-2199]

PR #482, opened 2025-03-01
feat(testnet): mainnet & devnet into testnet runtime

al3mart:
This PR introduces different features from mainnet and devnet runtimes into testnet.
And refactors some code just to make my work around the code a bit friendlier.

From mainnet:

- pallet-collective: f69df95fbf2d4b7635a96d7c2c9519575e164896 - added to mainnet: https://github.com/r0gue-io/pop-node/pull/448
- pallet-motion: f69df95fbf2d4b7635a96d7c2c9519575e164896 - added to mainnet: https://github.com/r0gue-io/pop-node/pull/448
- pallet-revive: 29d135b3e4b77e9e4048836bed4f15694ffb7524 - added to mainnet: https://github.com/r0gue-io/pop-node/pull/452
- pallet-treasury: c224de33dcec8d35897a6570a0cd70993603b295 - added to mainnet: https://github.com/r0gue-io/pop-node/pull/428 & refactored in https://github.com/r0gue-io/pop-node/pull/440
- Refactors xcm delivery fees to be more representative. - following the latest fees as per: https://github.com/r0gue-io/pop-node/pull/479/files#diff-275cd9b719e24b90693a6e8fcfc17f7313359cd1432fb8ae5ea744fd1fa7fb2e

From devnet:

- Messaging: #483

[sc-2987]


runtime/testnet/src/lib.rs
```diff
@@ -191,18 +214,13 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: Cow::Borrowed("pop"),
 	authoring_version: 1,
 	#[allow(clippy::zero_prefixed_literal)]
-	spec_version: 00_04_02,
+	spec_version: 00_04_03,
```
peterwht, line 217: nit: seems like this should be a major bump as we are introducing new features: `00_05_00`. 

Non-blocking though.

runtime/testnet/src/config/governance.rs
```diff
@@ -0,0 +1,68 @@
...
+	EnsureRoot<AccountId>,
+	EnsureProportionAtLeast<AccountId, CouncilCollective, 3, 4>,
+>;
+type UnanimousCouncilVote = EitherOfDiverse<
+	EnsureRoot<AccountId>,
+	EnsureProportionAtLeast<AccountId, CouncilCollective, 1, 1>,
+>;
+
+/// SUDO account set at genesis.
+const SUDO_ACCOUNT: &str = "5FPL3ZLqUk6MyBoZrQZ1Co29WAteX6T6N68TZ6jitHvhpyuD";
```
peterwht, line 25: Just notes for other reviewers: this is the correct Sudo address as exists on Testnet currently.

Screenshot 2025-03-03 at 11 20 37 AM
Screenshot 2025-03-03 at 11 20 46 AM

peterwht (approved):
Approving! No concerns here.

- Mainnet / testnet diffs show minimal config differences. The ones that exist aren't relevant to the goal here
- All configs moved from testnet lib.rs to individual modules are also fine without any concerning changes
- APIs are good as well

peterwht (approved):

runtime/testnet/src/config/governance.rs
Daanvdplas, line 25:
```suggestion
const SUDO_ADDRESS: &str = "5FPL3ZLqUk6MyBoZrQZ1Co29WAteX6T6N68TZ6jitHvhpyuD";
```

runtime/testnet/src/config/contracts.rs
```diff
@@ -83,6 +88,49 @@ impl pallet_contracts::Config for Runtime {
...
+		match value {
+			RuntimeCall::Revive(call) => Ok(call),
+			_ => Err(()),
+		}
+	}
+}
+
 // IMPORTANT: only runtime calls through the api are allowed.
 #[test]
 fn contracts_prevents_runtime_calls() {
```
Daanvdplas, line 136: Not related to this PR necessarily but do we still want this if we want people to build xcm messages perhaps outside the API @peterwht

Daanvdplas (commented):
I'm a bit suprised to see so many differences and inconsistencies. It should be super easy to go over the differences and make the runtime folders the same where possible, making maintenance and reviewing so much easier. In addition, a lot of types can be reused and put into `runtime/common`.

I think also while we are at it we should add non fungibles.

runtime/testnet/src/config/contracts.rs
peterwht: they can still use xcm_send directly.

peterwht:
> I'm a bit suprised to see so many differences and inconsistencies between the runtimes. It should be super easy to go over the differences and make the runtime folders the same where possible, making maintenance and reviewing so much easier. In addition, a lot of types can be reused and put into `runtime/common`.
> 
> I think also while we are at it we should add non fungibles.

The inconsistencies are because we aren't doing a regenesis, so existing pallets will maintain more or less the same config. 

Did you see any inconsistency that causes a significant change to mainnet that we need to address?

runtime/testnet/src/config/contracts.rs
peterwht: yes

runtime/testnet/src/config/xcm.rs
```diff
@@ -124,26 +157,17 @@ impl<T: Get<Location>> ContainsPair<Asset, Location> for NativeAssetFrom<T> {
...
+pub struct FilterByAssets<Assets>(PhantomData<Assets>);
+impl<Assets: Contains<Location>> Contains<(Location, Vec<Asset>)> for FilterByAssets<Assets> {
+	fn contains(t: &(Location, Vec<Asset>)) -> bool {
+		t.1.iter().all(|a| Assets::contains(&a.id.0))
 	}
 }
 
 /// Combinations of (Asset, Location) pairs which we trust as reserves.
-pub type TrustedReserves = (NativeAssetFrom<AssetHub>, NativeAssetExceptRelay);
+pub type TrustedReserves = NativeAssetFrom<AssetHub>;
```
Daanvdplas, line 170: We probably want more to allow cross chain use cases

al3mart: This sounds like runtime upgrades

Daanvdplas (approved):

peterwht (approved):

PR #483, opened 2025-03-01
feat(messaging): introduce messaging into testnet runtime

al3mart:
Required for #482

Introduces messaging api for contracts into testnet runtime. These feats were previously pushed to [`sub0`](https://github.com/r0gue-io/pop-node/tree/sub0) branch, so I've mainly ported the changes from there.
Now contracts are enabled to use ISMP and XCM to dispatch cross chain state read requests + callbacks. On top of the usual XCM operations.

The feature is currently piped via `pallet-contracts` even though revive is also present in the runtime.

Messaging api was introduced to devnet in: https://github.com/r0gue-io/pop-node/pull/338
And the callbacks were added to devnet in: https://github.com/r0gue-io/pop-node/pull/376


peterwht (commented):
Looking good.

- We need to update the CI (release) to build the testnet binary with --ismp enabled.
- For pop-api, can we include the examples as well please from this branch https://github.com/r0gue-io/pop-node/tree/daan/sub0/pop-api

ISMP adding parachain works:
Screenshot 2025-03-03 at 12 06 58 PM

al3mart:
> Looking good.
>     * We need to update the CI (release) to build the testnet binary with --ismp enabled.

Thanks Peter!

This is taken care here: https://github.com/r0gue-io/pop-node/pull/484.
Sorry this wasn't added to the description

peterwht (approved):
LGTM!

PR #487, opened 2025-03-04
refactor(devnet): remove dev accounts from live config

al3mart:
Removes dev accounts from `pop-runtime-devnet` genesis config using a similar configuration than testnet for now.

We might want to update collator keys.


al3mart:
Closing in favor of: https://github.com/r0gue-io/pop-node/pull/490

PR #488, opened 2025-03-04
fix(ci): different fixes for release workflow

al3mart:
- Moves away from using `upload-artifact@v3` as it is [deprecated](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). Uses v4 instead.

- Adds `NODE_BUILD_OPTS` to build-node task: https://github.com/r0gue-io/pop-node/pull/488/commits/c1e0ee03680cf88591f4afea7145db6157ab5461

- `Install packages` runs `apt-get update` https://github.com/r0gue-io/pop-node/pull/488/commits/b7b30f49fd6aa3277aa909f59a2ddbc3696ae6d9

- Downgrades srtool to 0.8.0 https://github.com/r0gue-io/pop-node/pull/488/commits/0a1c92eda04c487f0d6992e057800ea871648985

- Provides unique names for the artifacts being used on upload-artifact@v4 by appending the sha256 of the binary. [073885c](https://github.com/r0gue-io/pop-node/pull/488/commits/073885c7fa2df68c7a69ab75a0cf44a3ed9c2ff8). This change was a result of the version update for this action, which brought in some special requirements when it comes to working with multiple elements. [Issue for reference](https://github.com/actions/upload-artifact/issues/478#issuecomment-1872945805). It seemed that the matrix we define to use as compilation targets has, as understood by the action behavior and results, on `x86_64-unknown-linux-gnu`, although both our cases for this arch are clearly different. As a result and to protect the time dedicated to this I added the digests.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -72,7 +72,7 @@ jobs:
 
       - name: Build ${{ env.RUNTIME }} runtime
         id: srtool_build
-        uses: chevdor/srtool-actions@v0.9.2
+        uses: chevdor/srtool-actions@v0.8.0
         env:
           # runtimes do not have ismp feature
           BUILD_OPTS: '--features on-chain-release-build'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -185,6 +185,7 @@ jobs:
       - name: Install packages (Linux)
         if: contains(matrix.platform.target, 'linux')
         run: |
+          sudo apt-get update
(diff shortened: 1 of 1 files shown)


peterwht (approved):

al3mart:
I've triggered this build for ubuntu x86_64 with this changes. Haven't tested it yet:
-https://github.com/r0gue-io/pop-node/actions/runs/13669217847

peterwht (approved):

AlexD10S (approved):
Great! Thanks for addressing this.

PR #491, opened 2025-03-05
fix(testnet): bump testnet tx version

al3mart:
On the last testing done prior to testnet runtime v0.5.0 I've noticed that we forgot to bump the transaction_version.
This PR solves that.

This was noticed by comparing the running wasm vs the one we had built with `subwasm`.
Instructions have been included to our release tracking template to account for the above check - commit [0f541e6](https://github.com/r0gue-io/pop-node/pull/491/commits/0f541e6a2a7989d047658f25f6117ee998144d78)

### DIFF
--- a/.github/ISSUE_TEMPLATE/release_tracking.md
+++ b/.github/ISSUE_TEMPLATE/release_tracking.md
@@ -19,6 +19,8 @@ assignees: ''
 - [ ] All changes have passed peer-review. < link PR >
 - [ ] Update runtime crate version. Note that `pop-runtime-devnet` is usually not updated.
 - [ ] Runtime spec version is updated.
+- [ ] If needed, the `transaction_version` has been updated. This can be checked with
+  `subwasm`(https://github.com/chevdor/subwasm).
 - [ ] If needed, new benchmarks have been run. A diff between the new weights and the current ones has been reviewed.
     - [`substrate-weight-compare`](https://github.com/ggwpez/substrate-weight-compare) can be used for this purpose.
 - [ ] Execution of [`try-runtime`](https://github.com/paritytech/try-runtime-cli) doesn't point out any missing migrations or other items requiring action.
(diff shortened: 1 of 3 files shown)


peterwht (approved):

.github/ISSUE_TEMPLATE/release_tracking.md
```diff
@@ -76,6 +78,27 @@ _More instructions around using chopsticks for this can be found in [.chopsticks
     {"jsonrpc":"2.0","id":2,"method":"dev_newBlock","params":[{"count":20}]}
 ```
 
+- [ ] Verify if `transaction_version` needs to be updated:
+
+1. Build the new runtime. No concrete features need to be active, it's not a problem if there `runtime-benchmarks` or
+   `try-runtime` are included.
```
Daanvdplas, line 84:
```suggestion
1. Build the new runtime. No concrete features need to be active, and it's not a problem if `runtime-benchmarks` or `try-runtime` is enabled.
```

.github/ISSUE_TEMPLATE/release_tracking.md
```diff
@@ -76,6 +78,27 @@ _More instructions around using chopsticks for this can be found in [.chopsticks
...
+
+```shell
+    cargo build --release -p <runtime-crate>
+```
+
+2. Fetch the runtime that is on chain at the moment. This can be fetched from the corresponding github release, or a
```
Daanvdplas, line 90:
```suggestion
2. Fetch the on chain runtime. This can be fetched from the corresponding github release, or a
```

.github/ISSUE_TEMPLATE/release_tracking.md
```diff
@@ -76,6 +78,27 @@ _More instructions around using chopsticks for this can be found in [.chopsticks
...
+   node connected to the network via `subwasm`.
+
+```shell
+    subwasm get <node-endpoint> -o ./output/path
+```
+
+3. Use `subwasm` to create a diff between the runtimes and verify if `transaction_version` is compatible between them.
```
Daanvdplas, line 97:
```suggestion
3. Use `subwasm` to create a diff between the runtimes and verify if the `transaction_version` is compatible.
```

.github/ISSUE_TEMPLATE/release_tracking.md
```diff
@@ -76,6 +78,27 @@ _More instructions around using chopsticks for this can be found in [.chopsticks
...
+
+```shell
+    subwasm diff <old-runtime> <new-runtime>
```
Daanvdplas, line 100: old runtime is the path to the binary?

Daanvdplas (approved):
Nice I didn't know this! Exploring how we can make all of this easier or automated with the CLI could be interesting.

.github/ISSUE_TEMPLATE/release_tracking.md
al3mart: no, both are the corresponding runtime wasm blobs

al3mart: improved in: [9c3f7ec](https://github.com/r0gue-io/pop-node/pull/491/commits/9c3f7ec6a5a763b5fd5a8b6bc35f81cf466e7080)

al3mart:
> Nice I didn't know this! Exploring how we can make all of this easier or automated with the CLI could be interesting.

Agree! While `subwasm` provides means to solve this. There should be a better experience available for devs out there.

Daanvdplas (approved):

PR #494, opened 2025-03-06
Messaging general

f-gate:
This goes over most of the General section of the messaging pallet's tasks. 

## Losing ones deposit
Previously, in the extrinsics that take a message deposit, there was a possibility that the extrinsic would error before the message is stored and therefore when one tried to removed the message, it did not exist and therefore, the deposit could not be claimed. This has been fixed by moving much of the validation code before taking the deposit and then handling the fallible code after taking deposit more appropriately.

## Remove extrinsic
Notably the `remove` extrinsic has had a makeover so that it can be more comfortably unit tested.
I have moved much of the logic into the impl of `Message`, encapsulating the logic like this allowed me to cut down the complexity of the tests and the code.
Many tests have been written and cover all bases as far as im concerned.

The extrinsic has also been made transactional this is because we deposit an event at the end of the extrinsic that emits the messages removed, The previous implementation would remove messages without emitting an event which will be a pain for app devs . We could build it so it ignores erroneous messages and emits only the ones that pass but i thought this was a quick fix and i have lots to do thats of higher priority.

## MessageStatus
The `MessageStatus` has been introduced and ive moved away from combination of message status and message type.

Previously a message would be of type `IsmpTimedOut` now it is an `IsmpMessage` with the status of `TimedOut`.
This was done to differentiate between what a message is and what status it is in. I think this will allow more flexibility for messages in the future.  The downside is that one must handle all message statuses for a message and the previous version uses Rust's type system to enforce a messages status. This is the change im least confident about.

## Storage Deposits
Previously the deposit implementations were scattered around the codebase and there was a trait called `CalculateDeposits`. Following this code was more difficult than it had to be. I have made `deposits.rs` to contain all the deposit calculation code, and used a blanket implementation instead of the specific impls of the `CalculateDeposits`. This has improved readability nicely.
The `IsmpByteFee` is intended for Offchain storage costs hence it has been replaced with OffchainByteFee.
ByteFee has subsequently been changed to OnChainByteFee.

## Other
`CallbackT` has been renamed to CallbackExecutor for better semantics.
MessageId has been changed to a [u8; 32] from an unsigned integer. 

[sc-3066]

### DIFF
--- a/pallets/api/src/messaging/benchmarking.rs
+++ b/pallets/api/src/messaging/benchmarking.rs
@@ -1,0 +1,72 @@
+// //! Benchmarking setup for pallet_api::nonfungibles
+// #![cfg(feature = "runtime-benchmarks")]
+
+// use frame_benchmarking::{account, v2::*};
+// use frame_support::{BoundedVec, dispatch::RawOrigin, traits::Currency};
+// use sp_runtime::traits::Zero;
+
+
+// use super::*;
+// use crate::Read as _;
+// const SEED: u32 = 1;
+
+// // See if `generic_event` has been emitted.
+// fn assert_has_event<T: Config>
+// (generic_event: <T as crate::messaging::Config>::RuntimeEvent) 
+// where
+
+// {
+// 	frame_system::Pallet::<T>::assert_has_event(generic_event.into());
+// }
+
+
+// #[benchmarks(
+// 	where
+// 	T: pallet_balances::Config
+// )]
+// mod messaging_benchmarks {
+// 	use super::*;
+
+// 	#[benchmark]
+// 	// x: The number of removals required.
+// 	fn remove(x: Linear<1, 255>) {
+// 		let deposit: BalanceOf<T> = sp_runtime::traits::One::one();
+// 		let owner: AccountIdOf<T> = account("Alice", 0, SEED);
+// 		let mut message_ids: BoundedVec<MessageId, T::MaxRemovals> = BoundedVec::new();
+// 		pallet_balances::Pallet::<T>::make_free_balance_be(&owner, u32::MAX.into());
+
+// 		for i in 0..x {
+// 			<T as crate::messaging::Config>::Deposit::hold(
+// 				&HoldReason::Messaging.into(),
+// 				&owner,
+// 				deposit,
+// 			)
+// 			.unwrap();
+			
+// 			let message_id = H256::repeat_byte(i as u8);
+// 			let commitment = H256::repeat_byte(i as u8);
+
+// 			let good_message = Message::IsmpResponse {
+// 				commitment: commitment.clone(),
+// 				deposit,
+// 				response: Default::default(),
+// 				status: MessageStatus::Ok,
+// 			};
+
+// 			Messages::<T>::insert(&owner, &message_id.0, &good_message);
+// 			IsmpRequests::<T>::insert(&commitment, (&owner, &message_id.0));
+// 			message_ids.try_push(message_id.0).unwrap()
+// 		}
+// 		#[extrinsic_call]
+// 		Pallet::<T>::remove(RawOrigin::Signed(owner.clone()), message_ids.clone());
+		
+// 		assert_has_event::<T>(
+// 			crate::messaging::Event::Removed {
+// 				origin: owner,
+// 				messages: message_ids.to_vec(),
+// 			}.into()
+// 		)
+// 	}
+
+// 	impl_benchmark_test_suite!(Pallet, crate::mock::new_test_ext(), crate::mock::Test);
+// }
--- a/pallets/api/src/messaging/benchmarking.rs
+++ b/pallets/api/src/messaging/benchmarking.rs
@@ -2,7 +2,7 @@
 #![cfg(feature = "runtime-benchmarks")]
 
 use frame_benchmarking::{account, v2::*};
-use frame_support::{BoundedVec, dispatch::RawOrigin, traits::Currency};
+use frame_support::{dispatch::RawOrigin, traits::Currency, BoundedVec};
 use sp_runtime::traits::Zero;
 
 use super::*;
@@ -10,11 +10,7 @@ use crate::Read as _;
 const SEED: u32 = 1;
 
 // See if `generic_event` has been emitted.
-fn assert_has_event<T: Config>
-(generic_event: <T as crate::messaging::Config>::RuntimeEvent)
-where
-
-{
+fn assert_has_event<T: Config>(generic_event: <T as crate::messaging::Config>::RuntimeEvent) {
 	frame_system::Pallet::<T>::assert_has_event(generic_event.into());
 }
 
@@ -59,10 +55,8 @@ mod messaging_benchmarks {
 		Pallet::<T>::remove(RawOrigin::Signed(owner.clone()), message_ids.clone());
 
 		assert_has_event::<T>(
-			crate::messaging::Event::Removed {
-				origin: owner,
-				messages: message_ids.to_vec(),
-			}.into()
+			crate::messaging::Event::Removed { origin: owner, messages: message_ids.to_vec() }
+				.into(),
 		)
 	}
 
--- a/pallets/api/src/messaging/benchmarking.rs
+++ b/pallets/api/src/messaging/benchmarking.rs
@@ -1,72 +1,70 @@
-// //! Benchmarking setup for pallet_api::nonfungibles
-// #![cfg(feature = "runtime-benchmarks")]
+//! Benchmarking setup for pallet_api::nonfungibles
+#![cfg(feature = "runtime-benchmarks")]
 
-// use frame_benchmarking::{account, v2::*};
-// use frame_support::{BoundedVec, dispatch::RawOrigin, traits::Currency};
-// use sp_runtime::traits::Zero;
+use frame_benchmarking::{account, v2::*};
+use frame_support::{BoundedVec, dispatch::RawOrigin, traits::Currency};
+use sp_runtime::traits::Zero;
 
+use super::*;
+use crate::Read as _;
+const SEED: u32 = 1;
 
-// use super::*;
-// use crate::Read as _;
-// const SEED: u32 = 1;
+// See if `generic_event` has been emitted.
+fn assert_has_event<T: Config>
+(generic_event: <T as crate::messaging::Config>::RuntimeEvent)
+where
 
-// // See if `generic_event` has been emitted.
-// fn assert_has_event<T: Config>
-// (generic_event: <T as crate::messaging::Config>::RuntimeEvent) 
-// where
+{
+	frame_system::Pallet::<T>::assert_has_event(generic_event.into());
+}
 
-// {
-// 	frame_system::Pallet::<T>::assert_has_event(generic_event.into());
-// }
+#[benchmarks(
+	where
+	T: pallet_balances::Config
+)]
+mod messaging_benchmarks {
+	use super::*;
 
+	#[benchmark]
+	// x: The number of removals required.
+	fn remove(x: Linear<1, 255>) {
+		let deposit: BalanceOf<T> = sp_runtime::traits::One::one();
+		let owner: AccountIdOf<T> = account("Alice", 0, SEED);
+		let mut message_ids: BoundedVec<MessageId, T::MaxRemovals> = BoundedVec::new();
+		pallet_balances::Pallet::<T>::make_free_balance_be(&owner, u32::MAX.into());
 
-// #[benchmarks(
-// 	where
-// 	T: pallet_balances::Config
-// )]
-// mod messaging_benchmarks {
-// 	use super::*;
+		for i in 0..x {
+			<T as crate::messaging::Config>::Deposit::hold(
+				&HoldReason::Messaging.into(),
+				&owner,
+				deposit,
+			)
+			.unwrap();
 
-// 	#[benchmark]
-// 	// x: The number of removals required.
-// 	fn remove(x: Linear<1, 255>) {
-// 		let deposit: BalanceOf<T> = sp_runtime::traits::One::one();
-// 		let owner: AccountIdOf<T> = account("Alice", 0, SEED);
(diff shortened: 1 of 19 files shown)


pallets/api/src/messaging/mod.rs
```diff
@@ -234,151 +249,174 @@ pub mod pallet {
...
-				origin,
-				id,
-				commitment,
-				callback,
-			});
-			Ok(())
+			let maybe_commitment = T::IsmpDispatcher::default()
+				.dispatch_request(message.into(), FeeMetadata { payer: origin.clone(), fee });
+
+			match maybe_commitment {
```
peterwht, line 276: hmm, this is an interesting addition.

If the `dispatch_request` fails, why aren't we just returning a dispatch error and rolling back the extrinsic? The contract will be able to decode the error directly and then decide what to do in the same block (instead of having to do another call). All dispatchables are transactional.

pallets/api/src/messaging/mod.rs
```diff
@@ -234,151 +249,174 @@ pub mod pallet {
...
 
 		// TODO: does ismp allow querying to ensure that specified para id is supported?
 		#[pallet::call_index(2)]
-		#[pallet::weight(Weight::zero() // todo: benchmarking after consolidating storage
-			// Add any additional gas limit specified for callback execution
-			.saturating_add(callback.map(|cb| {
-				T::CallbackExecutor::execution_weight().saturating_add(cb.weight)
-			}).unwrap_or_default())
-		)]
+		#[pallet::weight(Weight::zero())]
```
peterwht, line 318: any reason for why the callback executor weight was removed?

pallets/api/src/messaging/mod.rs
```diff
@@ -234,151 +249,174 @@ pub mod pallet {
...
-			callback: Option<Callback>,
+			callback: Option<Callback<T::AccountId>>,
 		) -> DispatchResult {
 			let origin = ensure_signed(origin)?;
+			ensure!(!Messages::<T>::contains_key(&origin, &id), Error::<T>::MessageExists);
+
+			let deposit = calculate_protocol_deposit::<T, T::OnChainByteFee>(
+				ProtocolStorageDeposit::IsmpRequests,
+			) + calculate_message_deposit::<T, T::OnChainByteFee>() +
+				calculate_deposit_of::<T, T::OffChainByteFee, ismp::Post<T>>();
```
peterwht, line 332: Should we use safe arithmetic here (saturating add or checked add)?

pallets/api/src/messaging/mod.rs
```diff
@@ -234,151 +249,174 @@ pub mod pallet {
...
 
-			// Calculate deposit and place on hold.
-			let deposit = Self::calculate_deposit(
-				message.calculate_deposit() +
```
peterwht, line 312: This `message.calculate_deposit()` is missing in the newest calculation. Is it no longer needed?

pallets/api/src/messaging/mod.rs
```diff
@@ -234,151 +249,174 @@ pub mod pallet {
...
-				origin,
-				id,
-				commitment,
-				callback,
-			});
-			Ok(())
+			let maybe_commitment = T::IsmpDispatcher::default()
+				.dispatch_request(message.into(), FeeMetadata { payer: origin.clone(), fee });
+
+			match maybe_commitment {
```
peterwht, line 340: similar to the last one, I'm curious why we don't just error out, rollback the extrinsic, and then allow the contract to handle the error in the same function call & block?

pallets/api/src/messaging/mod.rs
```diff
@@ -234,151 +249,174 @@ pub mod pallet {
...
 
 		#[pallet::call_index(3)]
-		#[pallet::weight(Weight::zero() // todo: benchmarking after consolidating storage
-			// Add any additional gas limit specified for callback execution
-			.saturating_add(callback.map(|cb| {
-				T::CallbackExecutor::execution_weight().saturating_add(cb.weight)
-			}).unwrap_or_default())
-			// TODO: add weight of xcm_response dispatchable once benchmarked
-		)]
+		#[pallet::weight(Weight::zero())]
```
peterwht, line 381: any reason this is just zero now? (fine if it's just a placeholder until benchmarks are done)

pallets/api/src/messaging/mod.rs
```diff
@@ -234,151 +249,174 @@ pub mod pallet {
...
 		) -> DispatchResult {
 			let origin = ensure_signed(origin)?;
+			let querier = T::OriginConverter::try_convert(T::RuntimeOrigin::signed(origin.clone()))
+				.map_err(|_| Error::<T>::OriginConversionFailed)?;
+
+			ensure!(!Messages::<T>::contains_key(&origin, &id), Error::<T>::MessageExists);
+
+			let deposit = calculate_protocol_deposit::<T, T::OnChainByteFee>(
+				ProtocolStorageDeposit::IsmpRequests,
+			) + calculate_message_deposit::<T, T::OnChainByteFee>();
```
peterwht, line 397: should this use safe arithmetic?

pallets/api/src/messaging/mod.rs
```diff
@@ -399,8 +437,8 @@ pub mod pallet {
 			T::XcmResponseOrigin::ensure_origin(origin)?;
 
 			// Lookup message from query id.
-			let (origin, id) = XcmQueries::<T>::take(query_id).ok_or(Error::<T>::InvalidQuery)?;
-			let Some(Message::XcmQuery { query_id, callback, deposit }) =
+			let (origin, id) = XcmQueries::<T>::get(query_id).ok_or(Error::<T>::InvalidQuery)?;
+			let Some(Message::XcmQuery { query_id, callback, deposit, status }) =
 				Messages::<T>::get(&origin, &id)
 			else {
 				return Err(Error::<T>::InvalidMessage.into())
```
peterwht, line 444: could be refactored by using a `ok_or()`

pallets/api/src/messaging/mod.rs
```diff
@@ -435,67 +478,40 @@ pub mod pallet {
 			Ok(())
 		}
 
-		// Remove a request/response, returning any deposit previously taken.
+		/// Try and remove a collection of messages.
+		/// Will revert if any one of the messages is erroneous.
+		#[frame_support::transactional]
```
peterwht, line 483: this is generally not needed as all dispatchables by default are transactional.

pallets/api/src/messaging/mod.rs
```diff
@@ -399,8 +437,8 @@ pub mod pallet {
 			T::XcmResponseOrigin::ensure_origin(origin)?;
 
 			// Lookup message from query id.
-			let (origin, id) = XcmQueries::<T>::take(query_id).ok_or(Error::<T>::InvalidQuery)?;
-			let Some(Message::XcmQuery { query_id, callback, deposit }) =
+			let (origin, id) = XcmQueries::<T>::get(query_id).ok_or(Error::<T>::InvalidQuery)?;
```
peterwht, line 440: any reason you are wanting to leave the xcm query in the storage? (no longer using `take`).

pallets/api/src/messaging/mod.rs
```diff
@@ -435,67 +478,40 @@ pub mod pallet {
...
+		#[frame_support::transactional]
 		#[pallet::call_index(5)]
-		#[pallet::weight(Weight::zero())] // todo: benchmarking after consolidating storage
+		#[pallet::weight(Weight::zero())]
 		pub fn remove(
 			origin: OriginFor<T>,
 			messages: BoundedVec<MessageId, T::MaxRemovals>,
 		) -> DispatchResult {
 			let origin = ensure_signed(origin)?;
+
```
peterwht, line 491: likely out of scope for this PR, but right now anyone can remove a message. Should there be limited origins for this?

pallets/api/src/messaging/mod.rs
```diff
@@ -608,44 +626,129 @@ impl<T: Config> crate::Read for Pallet<T> {
...
+			// XCM responses can always be removed.
+			Message::XcmResponse { .. } => {
+				self.remove(origin, id);
+				Ok(())
+			},
+		}
+	}
+
+	/// Remove a message from storage.
+	/// Does no check on wether a message should be removed.
```
peterwht, line 692:
```suggestion
	/// Does no check on whether a message should be removed.
```

small typo :)

pallets/api/src/messaging/mod.rs
```diff
@@ -608,44 +626,129 @@ impl<T: Config> crate::Read for Pallet<T> {
...
+		T::Deposit::release(&HoldReason::Messaging.into(), origin, *deposit, Exact)?;
+		Ok(())
+	}
+}
+
+#[derive(Clone, Debug, Encode, Eq, Decode, MaxEncodedLen, PartialEq, TypeInfo)]
+pub enum MessageStatus {
+	/// No errors have been recorded.
+	Ok,
+	/// An error has occurred with this message>
```
peterwht, line 727:
```suggestion
	/// An error has occurred with this message.
```

pop-api/Cargo.toml
```diff
@@ -40,3 +40,4 @@ std = [
 	"pop-primitives/std",
 	"sp-io/std",
 ]
+messaging = [	]
```
peterwht, line 43: unfortunately, zepter or taplo will yell you at for an empty feature. Populating with some default feature works to resolve it.

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -0,0 +1,64 @@
...
+#[benchmarks(
+	where
+	T: pallet_balances::Config
+)]
+mod messaging_benchmarks {
+	use super::*;
+
+	#[benchmark]
+	// x: The number of removals required.
+	fn remove(x: Linear<1, 255>) {
```
peterwht, line 26: is it possible to use `T::MaxRemovals` here? (instead of `255`)

pallets/api/src/messaging/mod.rs
```diff
@@ -608,44 +626,129 @@ impl<T: Config> crate::Read for Pallet<T> {
...
+			Message::XcmQuery { deposit, .. } => deposit,
+			Message::XcmResponse { deposit, .. } => deposit,
+		};
+		T::Deposit::release(&HoldReason::Messaging.into(), origin, *deposit, Exact)?;
+		Ok(())
+	}
+}
+
+#[derive(Clone, Debug, Encode, Eq, Decode, MaxEncodedLen, PartialEq, TypeInfo)]
+pub enum MessageStatus {
```
peterwht, line 724: Nice! This feels better and keeps the `Message` struct a little cleaner. Also makes sense as it's easier to track what the original request was.

peterwht (changes requested):
Nice refactor! And thanks for the great PR description.

I haven't reviewed the tests yet. 

I've left some comments. I would love some clarity on why a few dispatchables don't rollback and return the error to the contract in case of failure.

pallets/api/src/messaging/mod.rs
```diff
@@ -26,22 +24,29 @@ use xcm::Response;
...
+use deposits::*;
+
+#[cfg(test)]
+mod tests;
+
 type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
 type BlockNumberOf<T> = BlockNumberFor<T>;
 type BalanceOf<T> = <<T as Config>::Deposit as Inspect<AccountIdOf<T>>>::Balance;
-pub type MessageId = u64;
+pub type MessageId = [u8; 32];
```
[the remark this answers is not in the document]
f-gate: So that one can pass a 32 byte hash as an id (Blake2/3, sha256)

pallets/api/src/messaging/mod.rs
f-gate: Huh! transactional was opt in when i was working at Imbue, looks like this has changed to opt out!
Awesome.

pallets/api/src/messaging/mod.rs
f-gate: Yes absolutely and test, second iteration will cover this

pallets/api/src/messaging/mod.rs
f-gate: Yea place holders, they were distracting

pallets/api/src/messaging/mod.rs
f-gate: i cant think of what other origin would be useful for now.

pallets/api/src/messaging/mod.rs
f-gate: turns out a wether is a castrated ram!

pallets/api/src/messaging/benchmarking.rs
f-gate: No but im tempted to increase the limit, what do you think?

f-gate:
(rebased commits with fixed git config)

pallets/api/src/messaging/mod.rs
peterwht: Please correct me if I'm wrong, but it seems this is a little different though. `message` is of type `ismp::Get()`, so `message.calculate_deposit()` seems more accurate for that particular message.

`calculate_message_deposit` seems to be more specific to our `Message` storage in the pallet.

pallets/api/src/messaging/mod.rs
peterwht: Shouldn't only the creator of the message be able to remove it?

pallets/api/src/messaging/mod.rs
peterwht: word of the day!

pallets/api/src/messaging/benchmarking.rs
peterwht: Hmm, 255 seems like a lot of messages. Maybe we can keep it there for now.

Instead of 255 directly, would it be possible to add a config (const) item to the mock and then use that instead?

peterwht (approved):
Nice, going to approve! This seems good for this initial refactor and setup.

PR #496, opened 2025-03-11
chore(runtime): include assetid migration

al3mart:
- Includes a new migration to populate `NextAssetId` for pallet_assets. Sets the next id to `10_000` as the biggest id existing on testnet live right now is > `7_000` so I've rounded up to `10_000`.
- Configures `CallbackHandle` for `pallet_assets` such that asset id is automatically incremented after new assets creation.
- Adds similar state population for testnet genesis state.
- Bumps the spec version to `00_05_02`.
- Bump the runtime crate version to `0.5.2`.

--- 

- [x] Changes for mainnet runtime can be found in: https://github.com/r0gue-io/pop-node/pull/497
- [x] Changes for devnet runtime can be found in: https://github.com/r0gue-io/pop-node/pull/498

---

[sc-3216]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13846,7 +13846,7 @@ dependencies = [
 
 [[package]]
 name = "pop-runtime-testnet"
-version = "0.5.0"
+version = "0.5.2"
 dependencies = [
  "anyhow",
  "cumulus-pallet-aura-ext 0.18.0",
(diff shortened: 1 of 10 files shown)


Daanvdplas (commented):
Does it make sense to apply this to the current testnet if we will replace it in the near future?

If we keep it for a while and we want to use foreign assets (which we will have to finalise and review first to deploy on testnet) then it makes sense. But if we will redeploy testnet before we don't need this.

al3mart:
> Does it make sense to apply this to the current testnet if we will replace it in the near future?

Seeing how we are with our testnet I'd say let's apply it. If it is never enacted no harm is done, changes are already there so no extra work.

But yeah, hopefully we regenesis testnet before doing more upgrades.

Daanvdplas (approved):

Daanvdplas (approved):

peterwht (approved):

PR #497, opened 2025-03-11
chore(mainnet): include assetid migration

al3mart:
- Includes a new migration to populate `NextAssetId` for pallet_assets. Sets the next id to `1`.
- Adds similar state population for testnet genesis state.
- Configures `CallbackHandle` for `pallet_assets` such that the asset id is increased after the creation of a new asset.
- Includes tests for the runtime config as well as for the genesis config.

### DIFF
--- a/runtime/mainnet/src/config/assets.rs
+++ b/runtime/mainnet/src/config/assets.rs
@@ -116,7 +116,18 @@ mod tests {
 	use crate::{AccountId, Balance};
 
 	mod assets {
+		use frame_support::traits::Incrementable;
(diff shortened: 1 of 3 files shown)


Daanvdplas (approved):
Nice work! Would be great to see a test showing that the asset id gets incremented.

al3mart:
> Nice work! Would be great to see a test showing that the asset id gets incremented.

[c2f6e16](https://github.com/r0gue-io/pop-node/pull/497/commits/c2f6e16fdab04237e943e258c1e32c75597ab518)

PR #499, opened 2025-03-12
chore(governance): initiate council migration for mainnet runtime

al3mart:
Introduces `OnRuntimeUpgrade` impl for `SetCouncilors`, initiating the council with some given members.
The migration is included in the runtime as unreleased, and can be removed once it has been executed on chain.

Tests are provide to demonstrate:
- the the migration does nothing if Members is already populated.
- the weight of `on_runtime_upgrade` is smaller than the maximum allowed in a block. (Can be fine tuned)

`pre_upgrade` and `post_upgrade` checks are provided for `try-runtime` feature.
- `post_upgrade` ensures that the members in storage after the upgrade contain the expected ones.
- `pre_upgrade` is implemented, but wasn't sure what to do with it as the council doesn't exist previous to the upgrade executing this migration. [Seems pre and post checks can only be executed together](https://github.com/paritytech/polkadot-sdk/blob/bcc272a1058f5b7cf9688a23b27d45349b66cb6a/substrate/frame/support/src/traits/try_runtime/mod.rs#L104-L105).

---

[sc-3216]

### DIFF
--- a/runtime/mainnet/src/config/governance.rs
+++ b/runtime/mainnet/src/config/governance.rs
@@ -85,7 +85,6 @@ pub(crate) mod initiate_council_migration {
 	use alloc::{vec, vec::Vec};
 
 	use frame_support::traits::OnRuntimeUpgrade;
-	use pallet_collective::WeightInfo;
 	#[cfg(feature = "try-runtime")]
 	use sp_runtime::TryRuntimeError;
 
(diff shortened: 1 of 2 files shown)


runtime/mainnet/src/config/governance.rs
```diff
@@ -80,6 +80,121 @@ impl pallet_sudo::Config for Runtime {
...
+			let members_in_storage: Vec<AccountId> =
+				pallet_collective::Members::<Runtime, CouncilCollective>::get();
+
+			if !members_in_storage.is_empty() {
+				// Members already set. Nothing to do.
+				return <Runtime as frame_system::Config>::DbWeight::get().reads(1);
+			}
+
+			let mut members: Vec<AccountId> = Members::get();
+			members.sort();
```
al3mart, line 117: Given the members are known we can sort out of this code and provide a sorted list.
Less operations to do during the migration.

al3mart: Members are sorted as of [b587c13](https://github.com/r0gue-io/pop-node/pull/499/commits/b587c13d0a5af189b1a540eb3697aa95c7f1f734)

runtime/mainnet/src/config/governance.rs
```diff
@@ -80,6 +80,124 @@ impl pallet_sudo::Config for Runtime {
...
+	use alloc::{vec, vec::Vec};
+
+	use frame_support::traits::OnRuntimeUpgrade;
+	#[cfg(feature = "try-runtime")]
+	use sp_runtime::TryRuntimeError;
+
+	use super::*;
+
+	parameter_types! {
+		C1: AccountId = AccountId::from_ss58check("13BL7T6bTgeEdfEdZqLCKJZPN8ncyFNxxHRKFb2YMATvyfH4").expect("address is valid SS58");
```
f-gate, line 94: Perhaps more descriptive names would be nice for me.

f-gate (approved):
lgtm

runtime/mainnet/src/config/governance.rs
f-gate: disclaimer: Havent checked the accounts

al3mart: Fair :) Just aimed to provide a list of councilors: C1, C2... name of each of them wasn't that important to me as the individual accounts.

peterwht (approved):
LGTM!

PR #501, opened 2025-03-18
Messaging: XCM

f-gate:
## XCM

Other than the predicted items in the task: 

### A Reversion
We ended up reverting the abstraction of a `Message` to how it was origininally conceived by Frank <3. 
I think the one to one mapping of status and message, but also the succinct and simple structs with no wasted data makes a nice decision. 

### XCM Query Timeouts
As of Polkadot 2412 although a timeout can be specified for a message the timeout does not actually enforce anything in the `OnResponse` implementation of `pallet-xcm` hence we are now recording timeouts in the messaging pallet. 
A message which receives a response which has been updated to an `XcmTimeout` will throw an Error and the callback will be ignored. Timeout messages must be removed using the `remove` extrinsic to release the deposit.

[sc-3063]


pallets/api/src/messaging/mod.rs
```diff
@@ -250,6 +274,29 @@ pub mod pallet {
 		Messaging,
 	}
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberOf<T>> for Pallet<T> {
+		fn on_initialize(n: BlockNumberOf<T>) -> Weight {
+			// As of polkadot-2412 XCM timeouts are not handled by the implementation of OnResponse
```
Daanvdplas, line 280: Have you made an issue for this?

pallets/api/src/messaging/mod.rs
```diff
@@ -250,6 +274,29 @@ pub mod pallet {
 		Messaging,
 	}
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberOf<T>> for Pallet<T> {
+		fn on_initialize(n: BlockNumberOf<T>) -> Weight {
```
peterwht, line 279: do we need to use a hook for this?

Wouldn't it be a little simpler to handle this lazily in `fn xcm_response`.

Basically, something along the lines of:
- store the timeout in `XcmQueries`, 
- in `xcm_response` compare the timeout from `XcmQueries` with the current block and then handle the message.

pallets/api/src/messaging/mod.rs
```diff
@@ -403,49 +443,68 @@ pub mod pallet {
...
+						T::Deposit::release(
+							&HoldReason::Messaging.into(),
+							&origin,
+							*deposit,
+							Exact,
+						)?;
+					},
+					Err(_) => {
+						// Due to dependance on runtime implementation this is hard to unit test.
+						// TODO: Discuss with peter options
```
peterwht, line 483: I may not fully understand the comment, but couldn't you create a mock implementation `CallbackExecutor` that you can force a failure with?

pallets/api/src/messaging/mod.rs
```diff
@@ -403,49 +443,68 @@ pub mod pallet {
...
+						XcmQueries::<T>::remove(query_id);
+						T::Deposit::release(
+							&HoldReason::Messaging.into(),
+							&origin,
+							*deposit,
+							Exact,
+						)?;
+					},
+					Err(_) => {
+						// Due to dependance on runtime implementation this is hard to unit test.
```
peterwht, line 482:
```suggestion
						// Due to dependence on runtime implementation this is hard to unit test.
```

small typo :)

pallets/api/src/messaging/mod.rs
```diff
@@ -403,49 +443,68 @@ pub mod pallet {
...
+						Messages::<T>::remove(&origin, &id);
+						XcmQueries::<T>::remove(query_id);
+						T::Deposit::release(
+							&HoldReason::Messaging.into(),
+							&origin,
+							*deposit,
+							Exact,
+						)?;
+					},
+					Err(_) => {
```
peterwht, line 481: the original code was a little less verbose. Do you have a reason for expanding this out?

See https://github.com/r0gue-io/pop-node/blob/main/pallets/api/src/messaging/mod.rs#L408

pallets/api/src/messaging/mod.rs
```diff
@@ -459,8 +518,36 @@ pub mod pallet {
 			let origin = ensure_signed(origin)?;
 
 			for id in &messages {
-				let message = Messages::<T>::get(&origin, id).ok_or(Error::<T>::MessageNotFound)?;
-				message.try_remove(&origin, id).and_then(|_| message.release_deposit(&origin))?;
+				let Some(message) = Messages::<T>::get(&origin, id) else {
```
peterwht, line 521: I see a small issue with this implementation, particularly around the origin of removing these messages.

Assume that a callback is NOT registered. A response is received that a contract relies on. Its offchain application is listening for the response, and then is going to read the state to get the response.

However, before that transaction can be sent to the contract, a malicious actor calls this remove function before the contract can access the data.

Seems this data should only be removable if (a couple ideas):
- The contract is the calling origin -- seems more limiting
- The data has been "claimed" or "pulled" from storage by the contract already.

pallets/api/src/messaging/mod.rs
```diff
@@ -471,41 +558,35 @@ pub mod pallet {
...
+		origin: &AccountIdOf<T>,
 		callback: Callback<T::AccountId>,
-		id: MessageId,
+		id: &MessageId,
 		data: &impl Encode,
-		deposit: BalanceOf<T>,
 	) -> DispatchResult {
 		let result = T::CallbackExecutor::execute(
 			origin.clone(),
 			[callback.selector.to_vec(), (id, data).encode()].concat(),
```
peterwht, line 569: just a comment, but in this spot the `(id,data).encode()` would change depending on what the ABI is.

pallets/api/src/messaging/transports/xcm.rs
```diff
@@ -1,10 +1,12 @@
 pub(crate) use xcm::latest::{Location, QueryId, Response};
+use xcm_builder::QueryControllerWeightInfo;
 
 use crate::messaging::{pallet::Call, BlockNumberOf, Config};
 
 pub trait NotifyQueryHandler<T: Config> {
+	type WeightInfo: QueryControllerWeightInfo;
 	/// Attempt to create a new query ID and register it as a query that is yet to respond, and
-	/// which will call a dispatchable when a response happens.
+	///  which will call a dispatchable when a response happens.
```
peterwht, line 9:
```suggestion
	/// which will call a dispatchable when a response happens.
```

Cargo.toml
```diff
@@ -24,7 +24,7 @@ exclude = [
 members = [
 	"node",
 	"pallets/*",
-	"pop-api/integration-tests",
+	#"pop-api/integration-tests",
```
peterwht, line 27: reason for this?

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -60,5 +60,80 @@ mod messaging_benchmarks {
...
+	/// x: Wether a successfully executing callback is provided.
+	#[benchmark]
+	fn xcm_response(x: Linear<0, 1>) {
+		let owner: AccountIdOf<T> = account("Alice", 0, SEED);
+		let message_id: [u8; 32] = [0; 32];
+		let responder = Location { parents: 1, interior: Junctions::Here };
+		let timeout = <BlockNumberOf<T> as One>::one() + frame_system::Pallet::<T>::block_number();
+		let callback = None;
+		let response = Response::ExecutionResult(None);
+		if x == 1 {
```
peterwht, line 103: A comment here would be great saying that the mock config has a always successful callback.

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -60,5 +60,80 @@ mod messaging_benchmarks {
...
+			RawOrigin::Signed(owner.clone()).into(),
+			message_id.clone(),
+			responder.clone(),
+			timeout,
+			callback.clone(),
+		)
+		.unwrap();
+
+		#[extrinsic_call]
+		Pallet::<T>::xcm_response(RawOrigin::Root, 0, response.clone());
```
peterwht, line 123: any reason this can't use a signed origin as well (instead of root)?

pallets/api/src/messaging/mod.rs
```diff
@@ -104,13 +108,18 @@ pub mod pallet {
 		type Xcm: NotifyQueryHandler<Self>;
 
 		type XcmResponseOrigin: EnsureOrigin<Self::RuntimeOrigin, Success = Location>;
+
+		/// SAFETY: Recommended this is small as is used to updated a message status in the hooks.
+		/// The maximum number of xcm timout updates that can be processed per block.
```
peterwht, line 113:
```suggestion
		/// The maximum number of xcm timeout updates that can be processed per block.
```

peterwht (changes requested):
Looking good!! Left some improvements and clarifications throughout

pallets/api/src/messaging/mod.rs
peterwht: in particular, the error case is handled twice, when it could be refactored to use a `let Some = ` with an early return.

pallets/api/src/messaging/mod.rs
f-gate: https://github.com/paritytech/polkadot-sdk/issues/7881

pallets/api/src/messaging/mod.rs
f-gate: This has been handled in the callback executor PR and is unrelated to this PR, i will remove the comment for now.

pallets/api/src/messaging/mod.rs
f-gate: Since origin is a key in Messages Alice cannot delete messages created by Bob (or a contract for that matter). 

Assuming the origin here is the contract, i see how a malicious contract can remove a response before polling has occured...
The claiming idea may have issues of its own but at least will allow responses to have a bit more resistance to this.

pallets/api/src/messaging/benchmarking.rs
f-gate: I wanted to insist on it being a different origin from signed since generally it would be a `Response` origin.
I thought root would be clearer and then returning a static response location.

pallets/api/src/messaging/mod.rs
f-gate: One must also update and check the timeout if the message state is ever read otherwise it will be inconsistent.

This is impossible if a dapp is just reading storage directly and not going through the read API. 
Hence i think until https://github.com/paritytech/polkadot-sdk/pull/7896 is merged this will do as a temp replacement.

pallets/api/src/messaging/mod.rs
f-gate: Agreed, left over from the change to using message status will simplify

Cargo.toml
f-gate: Currently breaking compilation and is out of scope for this PR

pallets/api/src/messaging/mod.rs
f-gate: I am going to make this into its own issue, since its not related to xcm ill push it into the next sprint.

f-gate: https://app.shortcut.com/r0gue/story/3341/allow-removal-of-messages-with-responses-to-be-removed-only-if-they-have-been-polled

peterwht (approved):

PR #504, opened 2025-03-21
feat(xcm): include `HasedDescription` to location conversation

al3mart:
Includes `HashedDescription` to location conversion configuration. This allows any user account to control their "foreign account" on another chain from Pop.

Adds some handy tests for mainnet sourced from: https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/assets/asset-hub-westend/tests/tests.rs

---

[sc-3389]

### DIFF
--- a/runtime/devnet/src/config/xcm.rs
+++ b/runtime/devnet/src/config/xcm.rs
@@ -12,9 +12,10 @@ use xcm::latest::prelude::*;
 use xcm_builder::{
 	AccountId32Aliases, AllowExplicitUnpaidExecutionFrom, AllowKnownQueryResponses,
 	AllowTopLevelPaidExecutionFrom, DescribeAllTerminal, DescribeFamily, EnsureXcmOrigin,
-	FixedWeightBounds, FrameTransactionalProcessor, FungibleAdapter, HashedDescription, IsConcrete, RelayChainAsNative, SiblingParachainAsNative,
-	SignedAccountId32AsNative, SignedToAccountId32, SovereignSignedViaLocation, TakeWeightCredit,
-	TrailingSetTopicAsId, UsingComponents, WithComputedOrigin, WithUniqueTopic,
+	FixedWeightBounds, FrameTransactionalProcessor, FungibleAdapter, HashedDescription, IsConcrete,
+	RelayChainAsNative, SiblingParachainAsNative, SignedAccountId32AsNative, SignedToAccountId32,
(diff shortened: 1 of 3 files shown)


peterwht (approved):

peterwht:
Thanks for getting this done!!

al3mart:
Based on this issue: https://github.com/r0gue-io/base-parachain/issues/56#issue-2940571840 We might be able to remove `ParentIsPreset` and `SiblingParachainConvertsVia`. I'm double checking if we would break anything if we do.

franciscoaguirre:
If the chain is live then you should keep the converters as removing them would break sovereign accounts. If it's not live then it's the best moment to only use `HashedDescription` and `AccountId32Aliases`. Those two are enough. That's why I proposed adding it to the templates

al3mart:
Based on @franciscoaguirre comment I've removed the two deprecated converters from devnet only as mainnet wasm is already registered on chain.

We might be able to remove these from testnet if we eventually do a regenesis of the testing environment.

Also, I can confirm that changing the configuration in the mainnet runtime resulted in different accounts than expected.

evilrobot-01:
> Based on @franciscoaguirre comment I've removed the two deprecated converters from devnet only as mainnet wasm is already registered on chain.
> 
> We might be able to remove these from testnet if we eventually do a regenesis of the testing environment.
> 
> Also, I can confirm that changing the configuration in the mainnet runtime resulted in different accounts than expected.

Technically mainnet isnt 'live' on Polkadot. I wonder if removing them via the first runtime upgrade would be fine, even if we have to manually move any SA funds, say via a migration, in the same upgrade?

Daanvdplas:
> Technically mainnet isnt 'live' on Polkadot. I wonder if removing them via the first runtime upgrade would be fine, even if we have to manually move any SA funds, say via a migration, in the same upgrade?

I think so, no xcm message has been send yet so no funds have been transferred to the sovereign accounts.

al3mart:
I can confirm that we should be OK by removing these from mainnet as well. I've pushed the changes in [ab2efb7](https://github.com/r0gue-io/pop-node/pull/504/commits/ab2efb7c3f2dc9579b1b2dc9376c6c37f13307cd) :+1:

Daanvdplas (approved):

peterwht (approved):

peterwht (approved):

PR #506, opened 2025-04-01
Messaging ismp core

f-gate:
ref: https://app.shortcut.com/r0gue/story/3061/feat-messaging-ismp

todo: 
- [ ] Research signed requests and how we should handle unsigned

[sc-3061]


pallets/api/src/messaging/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+	#[benchmark]
+	fn xcm_response(x: Linear<0, 1>) {
+		let owner: AccountIdOf<T> = account("Alice", 0, SEED);
+		let message_id: [u8; 32] = [0; 32];
+		let responder = Location { parents: 1, interior: Junctions::Here };
+		let timeout = <BlockNumberOf<T> as One>::one() + frame_system::Pallet::<T>::block_number();
+		let response = Response::ExecutionResult(None);
+
+		let callback = if x == 1 {
+			// The mock will always assume successfull callback.
```
peterwht, line 117: nit: typo
```suggestion
			// The mock will always assume successful callback.
```

pallets/api/src/messaging/test_utils.rs
```diff
@@ -0,0 +1,42 @@
...
+		nonce: 100u64,
+		from: vec![],
+		to: vec![],
+		timeout_timestamp: 10000,
+		body: vec![],
+	}
+}
+
+pub fn ismp_get_response() -> GetResponse {
+	GetResponse { get: ismp_get_request(), values: vec![] }
```
peterwht, line 37: does the values vec not affect the weight?

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+			.into(),
+		);
+		assert!(Messages::<T>::get(&owner, &message_id).is_none());
+		assert!(XcmQueries::<T>::get(0).is_none());
+	}
+
+	/// x: Is it a get. (example: 1 = get, 0 = post)
+	/// y: the response has a callback.
+	#[benchmark]
+	fn ismp_on_response(x: Linear<0, 1>, y: Linear<0, 1>) {
```
peterwht, line 158: this is not used anywhere, as far as I can tell. Shouldn't it be placed here: https://github.com/r0gue-io/pop-node/blob/74a8a782dd2a434d02beda7a7d368b3f07fcb8e2/pallets/api/src/messaging/transports/ismp.rs#L176

see here as another reference as well: 
https://github.com/RegionX-Labs/RegionX-Node/blob/main/pallets/regions/src/lib.rs#L488

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+		}
+
+		assert_has_event::<T>(event.into())
+	}
+
+	// Assuming a Timeout::Request(Request::Get) is handled the same as a
+	// Timeout::Request(Request::Post) x: Is a response timeout (example: 1 = response timeout, 0 =
+	// request timeout)
+	#[benchmark]
+	fn ismp_on_timeout(x: Linear<0, 1>) {
```
peterwht, line 217: similarly, needs to be used.

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+		let commitment = H256::repeat_byte(2u8);
+		let origin: T::AccountId = account("alice", 0, SEED);
+		let message_id = [1; 32];
+		let message = Message::Ismp { commitment, callback: None, deposit: One::one() };
```
peterwht, line 221: although a callback won't be executed, would it be more accurate to have a callback registered (worst-case).

peterwht, line 221: although a callback won't be executed, would it be more accurate to have a callback registered (worst-case)?

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+	fn ismp_get() {
+		let commitment = H256::repeat_byte(2u8);
+		let origin: T::AccountId = account("alice", 0, SEED);
+		let message_id = [1; 32];
+		let get = crate::messaging::ismp::Get::<T> {
+			dest: 0,
+			height: 0,
+			timeout: 0,
+			context: bounded_vec!(),
+			keys: bounded_vec!(),
```
peterwht, line 249: shouldn't these be variably affect the weight? So, we should do iterations up-to `T::MaxContextLen`, `T::MaxKeyLen` and `T::MaxKeys`.

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+		};
+
+		#[extrinsic_call]
+		Pallet::<T>::ismp_get(RawOrigin::Signed(origin.clone()), message_id, get, One::one(), None);
```
peterwht, line 253: seems like a callback should be an option here as well.

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+	}
+
+	#[benchmark]
+	fn ismp_post() {
+		let commitment = H256::repeat_byte(2u8);
+		let origin: T::AccountId = account("alice", 0, SEED);
+		let message_id = [1; 32];
+		let get = crate::messaging::ismp::Post::<T> { dest: 0, timeout: 0, data: bounded_vec![] };
```
peterwht, line 261: similarly, the data should be variably included as well. Up to `T::MaxDataLen`

peterwht: Also, variable name is `get` which is incorrect.

pallets/api/src/messaging/benchmarking.rs
peterwht: I haven't built this, but I believe compilation is going to fail. This PR removes the fee parameter, but you have `One::one()` in that place still.

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -0,0 +1,274 @@
...
+
+		#[extrinsic_call]
+		Pallet::<T>::ismp_post(
+			RawOrigin::Signed(origin.clone()),
+			message_id,
+			get,
+			One::one(),
```
peterwht, line 268: This will need to be removed (fee param no longer exists)

peterwht (changes requested):
It's getting there! But this PR does need a little work I'd say. 

- It seems a bit incomplete with the benchmarking: benchmarked functions are unused, needs to account for vectors a little better, and I don't believe it will compile right now
- Additionally, the PR is a little noisy due to needing a merge / rebase, which makes it a difficult to review.

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -236,37 +270,57 @@ mod messaging_benchmarks {
...
-			keys: bounded_vec!(),
+			context: vec![1u8].repeat(x as usize).try_into().unwrap(),
+			keys: outer_keys.try_into().unwrap(),
 		};
 
+		pallet_balances::Pallet::<T>::make_free_balance_be(&origin, u32::MAX.into());
+
 		#[extrinsic_call]
-		Pallet::<T>::ismp_get(RawOrigin::Signed(origin.clone()), message_id, get, One::one(), None);
+		Pallet::<T>::ismp_get(RawOrigin::Signed(origin.clone()), message_id, get, None);
```
peterwht, line 290: should we have a callback added as well?

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -236,37 +270,57 @@ mod messaging_benchmarks {
...
 	}
 
+	/// x: Maximun byte len of outgoing data.
 	#[benchmark]
-	fn ismp_post() {
-		let commitment = H256::repeat_byte(2u8);
+	fn ismp_post(x: Linear<0, {u16::MAX.into()}>) {
```
peterwht: similarly, I feel this might be better if you use the bound defined in the config (it also means the generated weights update depending on the specific runtime config)

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -236,37 +270,57 @@ mod messaging_benchmarks {
...
+		pallet_balances::Pallet::<T>::make_free_balance_be(&origin, u32::MAX.into());
+		
 		#[extrinsic_call]
 		Pallet::<T>::ismp_post(
 			RawOrigin::Signed(origin.clone()),
 			message_id,
 			get,
-			One::one(),
-			None,
+			Some(callback),
```
peterwht: What do you think of making the callback a variable as well (`Linear<0,1>`)? I believe you did this in another test -- that way the weight is more accurate if a callback or without.

This would apply to ismp_get as well.

pallets/api/src/messaging/mod.rs
```diff
@@ -93,33 +94,48 @@ pub mod pallet {
...
 		#[pallet::constant]
 		type MaxKeys: Get<u32>;
+		
+		/// SAFETY: should be less than or equal to u16.
+		/// The maximum byte length for a single key of an ismp request.
 		#[pallet::constant]
 		type MaxKeyLen: Get<u32>;
-
+		
+		/// The maximum lenght for a response.
```
peterwht, line 113: nit: typo in length.

peterwht: might be good to add a spell-check extension to VSCode. Rust Rover highlights these in red for me so I often point them out :)

pallets/api/src/messaging/test_utils.rs
```diff
@@ -26,10 +26,10 @@ pub fn ismp_post_request() -> PostRequest {
 		source: StateMachine::Polkadot(2000),
 		dest: StateMachine::Polkadot(2001),
 		nonce: 100u64,
-		from: vec![],
-		to: vec![],
+		from: [1u8].repeat(1000),
+		to: [1u8].repeat(1000),
```
peterwht, line 30: hmm, it seems like these are "module identifiers". Without looking into ISMP too closely, it seems likely the max size of this would be around 32 bytes (account id). 

These module IDs are controlled by the runtime, right? So, that would mean it's safe to set a max to 32 bytes.

pallets/api/src/messaging/test_utils.rs
```diff
@@ -14,9 +14,9 @@ pub fn ismp_get_request() -> GetRequest {
 		dest: StateMachine::Polkadot(2001),
 		nonce: 100u64,
 		from: vec![],
-		keys: vec![vec![]],
+		keys: vec![[1u8].repeat(1000)],
 		height: 1,
-		context: vec![],
+		context: [1u8].repeat(1000),
```
peterwht: seems like this might be better to make these into a parameter that is passed by the benchmark function (i.e., what is x currently equal to).

pallets/api/src/messaging/test_utils.rs
```diff
@@ -26,10 +26,10 @@ pub fn ismp_post_request() -> PostRequest {
 		source: StateMachine::Polkadot(2000),
 		dest: StateMachine::Polkadot(2001),
 		nonce: 100u64,
-		from: vec![],
-		to: vec![],
+		from: [1u8].repeat(1000),
+		to: [1u8].repeat(1000),
 		timeout_timestamp: 10000,
-		body: vec![],
+		body: [1u8].repeat(1000),
```
peterwht, line 32: Similarly, does this one make sense to bound by the benchmarks variables as well? Do we have any config item for the "body" of the Post?

peterwht (changes requested):

peterwht:
One last comment, the following still need to be populated by the benchmarks. If you will do in a follow up PR (the adding weights to the macros), that's fine!

```rust
impl<T: Config> IsmpModuleWeight for Pallet<T> {
	// Static as not in use.
	fn on_accept(&self, _request: &PostRequest) -> Weight {
		DbWeightOf::<T>::get().reads_writes(2, 1)
	}

	fn on_timeout(&self, _timeout: &Timeout) -> Weight {
		DbWeightOf::<T>::get().reads_writes(2, 1)
	}

	fn on_response(&self, _response: &Response) -> Weight {
		DbWeightOf::<T>::get().reads_writes(2, 2)
	}
}
```

pallets/api/src/messaging/benchmarking.rs
f-gate: Will be part of another task

pallets/api/src/messaging/benchmarking.rs
```diff
@@ -236,37 +270,57 @@ mod messaging_benchmarks {
...
 	}
 
+
+	/// x: Maximum context len: bound to u16::MAX T::MaxContextLen
+	/// y: Maximum length of keys (inner) len: bound to u16::MAX T::MaxKeyLen
+	/// z: Maximum amount of keys (outer) len: bound to u16::MAX T::MaxKeys
 	#[benchmark]
-	fn ismp_get() {
-		let commitment = H256::repeat_byte(2u8);
+	fn ismp_get(x: Linear<0, {u16::MAX.into()}>, y: Linear<0, {u16::MAX.into()}>, z: Linear<0, {u16::MAX.into()}>) {
```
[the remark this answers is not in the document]
f-gate: Yes mistaken understanding on my behalf

pallets/api/src/messaging/test_utils.rs
f-gate: great idea.

PR #509, opened 2025-04-15
chore(messaging): take fee for response

f-gate:
Awaiting completion runtime weights. : https://github.com/r0gue-io/pop-node/pull/510

For ISMP: 
We dont need to take a fee for the response as its included in the extrinsic that handles the messages. 

For XCM: 
Take a fee for the response.  I havent found anywhere that states how the execution of a QueryResponse is paid for and hence play safe and transfer a fee in advance. Assuming the weight is charged from the blockspace using the standard RuntimeCall dispatcher, we can extrapolate to say the blockspace is managed and all response fees are paid (and callback gas reserved). 

[sc-3158]


peterwht (approved):
LGTM!

PR #511, opened 2025-04-15
fix(testnet): metadata api

Daanvdplas:
The testnet runtime live on paseo is not returning the available api provided by the runtime: https://github.com/paritytech/frame-metadata/blob/a060c2d488771e5976c26d1bf64fdf8ad734cf1b/frame-metadata/src/v15.rs#L58

This is caused by an [issue in 4012 release](https://github.com/paritytech/polkadot-sdk/pull/6665) and will be fixed by 2503 yet we are not sure if that will be deployed (in stead of relaunching our testnet according to plans) or when this can be done. As a result a fix will be pushed to testnet.

Tests have been added that show the problem and will have to be changed accordingly with the 2503 release.


Cargo.toml
```diff
@@ -81,6 +81,7 @@ pop-runtime-testnet = { path = "runtime/testnet", default-features = true } # de
 frame-benchmarking = { git = "https://github.com/paritytech/polkadot-sdk", branch = "stable2412", default-features = false }
 frame-benchmarking-cli = { git = "https://github.com/paritytech/polkadot-sdk", branch = "stable2412" }
 frame-executive = { git = "https://github.com/paritytech/polkadot-sdk", branch = "stable2412", default-features = false }
+frame-metadata = { version = "18.0.0", default-features = false }
```
Daanvdplas, line 84: Not in the polkadot sdk: https://github.com/paritytech/frame-metadata/blob/main/frame-metadata/src/v15.rs#L58

runtime/testnet/src/lib.rs
```diff
@@ -413,3 +948,41 @@ fn test_lookup_config() {
...
+		let version = 15;
+		let opague_meta = T::metadata_at_version(version).expect("V15 should exist");
+		let prefixed_meta_bytes = opague_meta.deref();
+		assert_eq!(prefixed_meta_bytes, Runtime::metadata_at_version(version).unwrap().deref());
+		let prefixed_meta = RuntimeMetadataPrefixed::decode(&mut &prefixed_meta_bytes[..]).unwrap();
+		// Ensure that we have the V15 variant.
+		let RuntimeMetadata::V15(metadata) = prefixed_meta.1 else {
+			panic!("Expected metadata variant V15, got a different variant");
+		};
+		assert!(!metadata.apis.is_empty());
```
Daanvdplas, line 981: This indicates that the problem is solved and the api list is not empty

runtime/mainnet/src/apis.rs
```diff
@@ -505,3 +505,43 @@ impl_runtime_apis! {
...
+		let opague_meta = T::metadata_at_version(version).expect("V15 should exist");
+		let prefixed_meta_bytes = opague_meta.deref();
+		assert_eq!(prefixed_meta_bytes, Runtime::metadata_at_version(version).unwrap().deref());
+		let prefixed_meta = RuntimeMetadataPrefixed::decode(&mut &prefixed_meta_bytes[..])
+			.expect("Decoding prefixed metadata should succeed");
+		// Ensure that we have the V15 variant.
+		let RuntimeMetadata::V15(metadata) = prefixed_meta.1 else {
+			panic!("Expected metadata variant V15, got a different variant");
+		};
+		// TODO: This shouldn't be empty and should be resolved with 2503.
```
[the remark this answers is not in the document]
al3mart: Yeah, this makes sense! ❤️

al3mart (approved):
Test pass nicely and was able to see how after this change runtime apis can be read again using dev.papi.how

On the running version of the testnet no runtime apis are accessible, after the fix, they are available again.

:+1:

peterwht (approved):
Thanks for handling!

PR #515, opened 2025-04-22
feat(messaging): callback gas topup

f-gate:
As mentioned in the telegram allow people to top up callback gas for requesting messages.
Was low hanging fruit so thought id get it in before the ismp_weight issue call.

[sc-3601]

### DIFF
--- a/pallets/api/src/messaging/mod.rs
+++ b/pallets/api/src/messaging/mod.rs
@@ -280,6 +280,8 @@ pub mod pallet {
 		IsmpTimedOut { commitment: H256 },
 		/// A collection of xcm queries have timed out.
 		XcmQueriesTimedOut { query_ids: Vec<QueryId> },
+		/// Callback gas has been topped up.
+		CallbackGasIncreased { message_id: MessageId, total_weight: Weight },
 	}
 
 	#[pallet::error]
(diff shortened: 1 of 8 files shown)


pallets/api/src/messaging/mod.rs
```diff
@@ -688,6 +693,74 @@ pub mod pallet {
...
 			Ok(())
 		}
+
+		/// Top up the callback weight for a pending message.
+		///
+		/// This extrinsic allows users to increase the gas (weight) budget allocated for a callback
+		/// associated with an in-flight message. This is useful when the initially allocated weight
+		/// is insufficient to complete the callback.
+		///
+		/// The additional fee for the new weight is held from the user's balance using the
```
peterwht, line 703: "user's balance" seems incorrect -- most likely a contract, right?

pallets/api/src/messaging/mod.rs
```diff
@@ -688,6 +693,74 @@ pub mod pallet {
...
+		/// Only pending requests can have their weight increased.
+		///
+		/// # Parameters
+		///
+		/// - `origin`: Must be a signed account.
+		/// - `message_id`: The identifier of the message to be topped up.
+		/// - `additional_weight`: The additional weight to be appended to the message's existing
+		///   callback weight.
+		#[pallet::call_index(6)]
+		#[pallet::weight(Weight::zero())]
```
peterwht, line 715: needs updating -- still zero!

peterwht (changes requested):
looks good, just needs a small change in the weight macro

peterwht (approved):

PR #517, opened 2025-04-23
chore: `BlockNumberProvider` for multisig, nfts & proxy

al3mart:
Changes for [polkadot-sdk#5723](https://github.com/paritytech/polkadot-sdk/pull/5723).

This PR adds the ability for these pallets to specify their source of the block number.
They have been all configured to depend on the local block number.

Note that `pallet_nfts` in devnet is using a local forked version of the pallet, not depending on the newer version upstream.
Tests to ensure these configurations have been added to mainnet runtime.

---

[sc-3547]


f-gate (approved):
lgtm

PR #386, opened 2024-11-19
chore: psp34 example contract

chungquantin:

PR #526, opened 2025-04-24
fix: nonfungible api mismatched balance type

chungquantin:
There is an issue in the nonfungible API that when user calls `api::nonfungibles::create`, it throws `Contract Decode Failed`. Reason to that is the mismatched type in the `price` field of `MintSettings` in the runtime and in the API crate. 
- Remove the balance `u32` type in the nonfungibles api crate. Use the `primitives::Balance` type instead.
- Update all tests in the nonfungible integration tests to test with the min / max value of the data types instead of `None`.

### DIFF
--- a/pop-api/Cargo.toml
+++ b/pop-api/Cargo.toml
@@ -6,8 +6,8 @@ name = "pop-api"
 version = "0.0.0"
 
 [dependencies]
-bitflags = { version = "1.3.2" }
-enumflags2 = "0.7.9"
+bitflags = { version = "1.3.2", optional = true }
+enumflags2 = { version = "0.7.9", optional = true }
 
 # Pop.
 pop-primitives = { path = "../primitives", default-features = false }
@@ -35,7 +35,7 @@ path = "src/lib.rs"
 default = [ "std" ]
 fungibles = [  ]
 messaging = [  ]
-nonfungibles = [  ]
+nonfungibles = [ "dep:bitflags", "dep:enumflags2" ]
 std = [
 	"ink/std",
 	"pop-primitives/std",
--- a/pop-api/Cargo.toml
+++ b/pop-api/Cargo.toml
@@ -6,7 +6,7 @@ name = "pop-api"
 version = "0.0.0"
 
 [dependencies]
-bitflags = { version = "1.3.2", optional = true }
+bitflags = { version = "2.8.0", optional = true }
 enumflags2 = { version = "0.7.9", optional = true }
 
 # Pop.
--- a/pop-api/integration-tests/src/nonfungibles/utils.rs
+++ b/pop-api/integration-tests/src/nonfungibles/utils.rs
@@ -352,9 +352,9 @@ pub(super) mod nfts {
 	pub(crate) fn default_mint_settings() -> MintSettings {
 		MintSettings {
 			mint_type: MintType::Issuer,
-			price: None,
-			start_block: None,
-			end_block: None,
+			price: Some(Balance::MAX),
+			start_block: Some(BlockNumber::MIN),
+			end_block: Some(BlockNumber::MAX),
 			default_item_settings: ItemSettings::all_enabled(),
 		}
 	}
--- a/pop-api/integration-tests/src/nonfungibles/utils.rs
+++ b/pop-api/integration-tests/src/nonfungibles/utils.rs
@@ -283,7 +283,7 @@ pub(super) mod nfts {
 	) -> pallet_nfts::CollectionConfig<u128, BlockNumber, CollectionId> {
 		pallet_nfts::CollectionConfig {
 			settings: pallet_nfts::CollectionSettings::all_enabled(),
-			max_supply: None,
+			max_supply: Some(u32::MAX),
 			mint_settings: pallet_nfts::MintSettings::default(),
 		}
 	}
@@ -312,7 +312,7 @@ pub(super) mod nfts {
 			collection,
 			item,
 			operator.clone().into(),
-			None
+			Some(BlockNumber::MAX)
 		));
 		(collection, item)
 	}
--- a/pop-api/integration-tests/src/nonfungibles/utils.rs
+++ b/pop-api/integration-tests/src/nonfungibles/utils.rs
@@ -312,7 +312,7 @@ pub(super) mod nfts {
 			collection,
 			item,
 			operator.clone().into(),
-			Some(BlockNumber::MAX)
+			Some(u32::MAX)
 		));
 		(collection, item)
 	}
@@ -352,9 +352,9 @@ pub(super) mod nfts {
 	pub(crate) fn default_mint_settings() -> MintSettings {
 		MintSettings {
 			mint_type: MintType::Issuer,
-			price: Some(Balance::MAX),
-			start_block: Some(BlockNumber::MIN),
-			end_block: Some(BlockNumber::MAX),
+			price: Some(u128::MAX),
+			start_block: Some(u32::MIN),
+			end_block: Some(u32::MAX),
 			default_item_settings: ItemSettings::all_enabled(),
 		}
 	}
--- a/pop-api/src/v0/nonfungibles/types.rs
+++ b/pop-api/src/v0/nonfungibles/types.rs
@@ -3,9 +3,11 @@
 use enumflags2::{bitflags, BitFlags};
 
 use super::*;
-use crate::{macros::impl_codec_bitflags, primitives::AccountId};
+use crate::{
+	macros::impl_codec_bitflags,
+	primitives::{AccountId, Balance},
+};
 
-type Balance = u32;
 /// The identifier of a collection.
 pub type CollectionId = u32;
 /// The identifier of an item.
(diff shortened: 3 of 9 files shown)


pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -3,9 +3,11 @@
 use enumflags2::{bitflags, BitFlags};
 
 use super::*;
-use crate::{macros::impl_codec_bitflags, primitives::AccountId};
+use crate::{
+	macros::impl_codec_bitflags,
+	primitives::{AccountId, Balance},
```
evilrobot-01, line 8: Some of the psp34 interfaces use a u32 for Balances - does this change affect those?

chungquantin: The only place where `Balance` is used is in the mint price of `CollectionConfig` in the `create` method. And `create` method is not in the PSP34 interface.

pop-api/integration-tests/src/nonfungibles/utils.rs
```diff
@@ -352,9 +352,9 @@ pub(super) mod nfts {
 	pub(crate) fn default_mint_settings() -> MintSettings {
 		MintSettings {
 			mint_type: MintType::Issuer,
-			price: None,
-			start_block: None,
-			end_block: None,
+			price: Some(Balance::MAX),
```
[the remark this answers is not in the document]
chungquantin: Pushed a commit to update where I think it is needed. Other `None` is passed to the dispatchable methods of the pallet directly so it does not matter.

Daanvdplas (approved):
LGTM, lets get this on testnet next!

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -178,18 +180,30 @@ mod tests {
...
 	#[test]
 	fn ensure_mint_witness() {
 		assert_eq!(
-			MintWitness { owned_item: None, mint_price: None }.encode(),
-			pallet_nfts::MintWitness::<ItemId, Balance> { owned_item: None, mint_price: None }
-				.encode()
+			MintWitness { owned_item: Some(ItemId::MAX), mint_price: Some(Balance::MAX) }.encode(),
+			pallet_nfts::MintWitness::<ItemId, Balance> {
+				owned_item: Some(ItemId::MAX),
+				mint_price: Some(Balance::MAX)
```
evilrobot-01, line 204: Probably wise to always use a static type as the check value, that way any type aliases changing will be caught. A parachain bricked once due to a change in the nonce type which went undetected.

In other words, we use `u128::MAX| instead so it's explicit.

pop-api/integration-tests/src/nonfungibles/utils.rs
```diff
@@ -352,9 +352,9 @@ pub(super) mod nfts {
...
+			start_block: Some(BlockNumber::MIN),
+			end_block: Some(BlockNumber::MAX),
```
evilrobot-01, line 357:
```suggestion
			end_block: Some(u32::MAX),
```

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -178,18 +180,30 @@ mod tests {
...
 
 	#[test]
 	fn ensure_mint_witness() {
 		assert_eq!(
-			MintWitness { owned_item: None, mint_price: None }.encode(),
-			pallet_nfts::MintWitness::<ItemId, Balance> { owned_item: None, mint_price: None }
-				.encode()
+			MintWitness { owned_item: Some(ItemId::MAX), mint_price: Some(Balance::MAX) }.encode(),
+			pallet_nfts::MintWitness::<ItemId, Balance> {
+				owned_item: Some(ItemId::MAX),
```
evilrobot-01, line 203:
```suggestion
				owned_item: Some(u32::MAX),
```

pop-api/src/v0/nonfungibles/types.rs
evilrobot-01, line 204:
```suggestion
				mint_price: Some(u128::MAX)
```

Probably wise to always use an explicit type as the check value, that way any type aliases changing will be caught. A parachain bricked once due to a change in the nonce type which went undetected.

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -244,11 +258,12 @@ mod tests {
...
 	fn ensure_mint_type() {
 		assert_eq!(
-			vec![MintType::Issuer, MintType::Public, MintType::HolderOf(0),].encode(),
+			vec![MintType::Issuer, MintType::Public, MintType::HolderOf(CollectionId::MAX)]
+				.encode(),
 			vec![
 				pallet_nfts::MintType::Issuer,
 				pallet_nfts::MintType::Public,
-				pallet_nfts::MintType::HolderOf(0)
+				pallet_nfts::MintType::HolderOf(CollectionId::MAX)
```
evilrobot-01, line 266:
```suggestion
				pallet_nfts::MintType::HolderOf(u32::MAX)
```

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -315,9 +330,9 @@ mod tests {
 	fn default_mint_settings() -> MintSettings {
 		MintSettings {
 			mint_type: MintType::Public,
-			price: None,
-			start_block: None,
-			end_block: None,
+			price: Some(Balance::MAX),
```
evilrobot-01, line 333:
```suggestion
			price: Some(u128::MAX),
```

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -315,9 +330,9 @@ mod tests {
...
+			start_block: Some(BlockNumber::MIN),
```
evilrobot-01, line 334:
```suggestion
			start_block: Some(u32::MIN),
```

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -315,9 +330,9 @@ mod tests {
...
+			end_block: Some(BlockNumber::MAX),
```
evilrobot-01, line 335:
```suggestion
			end_block: Some(u32::MAX),
```

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -326,9 +341,9 @@ mod tests {
 	) -> pallet_nfts::MintSettings<Balance, BlockNumber, CollectionId> {
 		pallet_nfts::MintSettings::<Balance, BlockNumber, CollectionId> {
 			mint_type: pallet_nfts::MintType::Public,
-			price: None,
-			start_block: None,
-			end_block: None,
+			price: Some(Balance::MAX),
```
evilrobot-01, line 344:
```suggestion
			price: Some(u128::MAX),
```

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -326,9 +341,9 @@ mod tests {
...
+			start_block: Some(BlockNumber::MIN),
```
evilrobot-01, line 345:
```suggestion
			start_block: Some(u32::MIN),
```

pop-api/src/v0/nonfungibles/types.rs
```diff
@@ -326,9 +341,9 @@ mod tests {
...
+			end_block: Some(BlockNumber::MAX),
```
evilrobot-01, line 346:
```suggestion
			end_block: Some(u32::MAX),
```

evilrobot-01:
The reason for the failure when testing manually was due to the contract not having sufficient balance for the collection deposit. Whilst the goal of these tests is ensuring the non-fungibles pallet works, they perhaps set a bad example, especially as there is no explicit non-fungibles example. I found https://github.com/r0gue-io/pop-api-examples so perhaps that could be migrated across and improved?

Ideally such functions should be `payable` and should consider a proper flow rather than https://github.com/r0gue-io/pop-node/blob/27f09da4db9f0b579a22369b1c4a2017c93c3ec5/pop-api/integration-tests/src/nonfungibles/mod.rs#L365. 

I'm not suggesting a change here now, just that some proper example is made available which uses a more realistic flow.

evilrobot-01:
I see we have:
- https://github.com/r0gue-io/pop-api-examples/tree/main/nfts: should be removed imo
- https://github.com/r0gue-io/pop-node/tree/sub0/pop-api/examples/nfts: deprecated, but no reason provided why
- https://github.com/r0gue-io/pop-node/pull/386 - unfinished, assuming blocked by drink
- https://github.com/r0gue-io/project-interpop/tree/main/dao_with_nft_verification: slightly different use case

We really just need a simple starter example at https://github.com/r0gue-io/pop-node/tree/main/pop-api/examples.

pop-api/Cargo.toml
```diff
@@ -6,8 +6,8 @@ name = "pop-api"
 version = "0.0.0"
 
 [dependencies]
-bitflags = { version = "1.3.2" }
-enumflags2 = "0.7.9"
+bitflags = { version = "1.3.2", optional = true }
```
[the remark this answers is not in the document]
chungquantin: There is a conflict with [`crossterm` dependency in `contracts-build`](https://github.com/use-ink/cargo-contract/blob/master/crates/build/Cargo.toml#L45) (uses bitflag 2.8.0) when updating this crate to `2.9.0` so I updated it to `2.8.0`.

chungquantin:
I'll add the example contract for nonfungible in #386 after this PR. Would love to have this PR merged first to resolve the issue on testnet!

evilrobot-01:
> Would love to have this PR merged first to resolve the issue on testnet!

The problem isnt specific to testnet, its `v0` of the pop-api crate which would affect any runtime.

evilrobot-01 (approved):
Looks great, thanks for addressing the feedback so promptly. 

Will merge once CI is green.

PR #538, opened 2025-04-27
feat(session): implement `DisablingStrategy`

al3mart:
As per [polkadot-sdk#7581](https://github.com/paritytech/polkadot-sdk/pull/7581).

Implement `DisablingStrategy` for `pallet_session` in every runtime.
Note that the migration included upstream is not a requirement for `pop-node`.


Daanvdplas:
> Note that the migration included upstream is not a requirement for pop-node.

Why?

PR #548, opened 2025-04-28
chore(messaging): use an unbalanced handler instead of naive transfer

f-gate:
[sc-3641]


runtime/testnet/src/config/api/mod.rs
```diff
@@ -99,24 +99,27 @@ parameter_types! {
...
+	type FeeHandler = ();
 	type Fungibles = Balances;
 	type IsmpDispatcher = Ismp;
 	type IsmpRelayerFee = IsmpRelayerFee;
+	type Keccak256 = Ismp;
 	type MaxContextLen = ConstU32<64>;
-	type MaxDataLen = ConstU32<1024>;
-	type MaxKeyLen = ConstU32<32>;
+	type MaxDataLen = ConstU32<512>;
+	type MaxKeyLen = ConstU32<8>;
```
peterwht, line 110: I believe this will be a problem. Look at this storage key for asset 1 in asset hub: `0x682a59d51ab9e48a8c8cc418ff9708d2d34371a193a751eea5883e9553457b2ed82c12285b5d4551f88e8f6e7eb52b8101000000`

The length is well above 8. Something like 64 or 128 might be more appropriate..

peterwht (approved):
nice addition! I left one comment, but I will approve this PR as it isn't directly related!

runtime/devnet/src/config/api/mod.rs
```diff
@@ -114,16 +114,16 @@ impl nonfungibles::Config for Runtime {
...
+	// TODO: What is reasonable.
 	pub const IsmpRelayerFee: crate::Balance = crate::UNIT / 2;
 	pub DummyFeeAccount: AccountId = crate::PalletId(*b"dummyacc").into_account_truncating();
-
 }
 
 impl messaging::Config for Runtime {
 	type CallbackExecutor = CallbackExecutor;
-	type FeeAccount = DummyFeeAccount;
+	// Burn fees.
```
evilrobot-01, line 125: Assume this is simply a todo, but as we dont have our own token and DOT fees are required for continued operations, we cannnot be burning fees.

f-gate: Devnet seems to be burning transaction payments anyway?

```rust	
type OnChargeTransaction = pallet_transaction_payment::FungibleAdapter<Balances, ()>;
```

f-gate: I will update the devnet to use a proper deal with fees impl to split between treasury and maintenance

runtime/testnet/src/config/api/mod.rs
f-gate: i will create a task to reason about this issue. Cases like this are really useful.

f-gate: ref: https://app.shortcut.com/r0gue/story/3704/proper-parameters-for-ismp-keys-config

PR #549, opened 2025-04-28
bug(messaging): Handle additional static weight outside callback function

f-gate:
[sc-3701]

### DIFF
--- a/pallets/api/src/messaging/mod.rs
+++ b/pallets/api/src/messaging/mod.rs
@@ -649,14 +649,7 @@ pub mod pallet {
 				log::debug!(target: "pop-api::extension", "xcm callback={:?}, response={:?}", callback, xcm_response);
 				// Never roll back state if call fails.
 				// Ensure that the response can be polled.
-				if Self::call(
-					&initiating_origin,
-					callback.to_owned(),
-					&id,
-					&xcm_response,
-				)
-				.is_ok()
-				{
+				if Self::call(&initiating_origin, callback.to_owned(), &id, &xcm_response).is_ok() {
 					Messages::<T>::remove(&initiating_origin, id);
 					XcmQueries::<T>::remove(query_id);
 					T::Fungibles::release(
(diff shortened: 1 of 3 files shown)


pallets/api/src/messaging/mod.rs
```diff
@@ -615,6 +615,15 @@ pub mod pallet {
 			xcm_response: Response,
 		) -> DispatchResult {
 			T::XcmResponseOrigin::ensure_origin(origin)?;
+
+			// Manually adjust weight ahead of fallible execution.
+			// The fees of which should have been paid.
+			frame_system::Pallet::<T>::register_extra_weight_unchecked(
```
peterwht, line 621: Similar to how you did it somewhere else, shouldn't we check if the weight is already exceeded before doing this unchecked increase?

Also, it would be good to add a comment why this is charged in addition to the xcm_response in `#[pallet::weight({`

pallets/api/src/messaging/mod.rs
```diff
@@ -854,16 +845,10 @@ pub mod pallet {
...
-			// This should include T::CallbackExecutor::execution_weight() where applicable unless
-			// already charged.
-			static_weight_adjustment: Option<Weight>,
 		) -> DispatchResult {
 			// This is the total weight that should be deducted from the blockspace for callback
 			// execution.
-			let max_weight = callback
-				.weight
-				.checked_add(&static_weight_adjustment.unwrap_or(Zero::zero()))
-				.ok_or(Error::<T>::BlockspaceAllowanceReached)?;
```
peterwht, line 866: this is the piece I am talking about. Should it be saved?

pallets/api/src/messaging/mod.rs
f-gate: Yes great catch!

pallets/api/src/messaging/mod.rs
f-gate: This line is depricated now we have seperated the `static_weight_adjustment` and the callback weight

peterwht (approved):

PR #551, opened 2025-04-29
fix(nonfungible): optional next collection ID in the interface library

chungquantin:
Notice a mistake in the `next_collection_id` interface method. In the pallet api nonfungibles, we return `Option<CollectionId>` not `CollectionId` ([Reference](https://github.com/r0gue-io/pop-node/blob/main/pallets/api/src/nonfungibles/mod.rs#L614)).

EDIT Daan: In stead of pursuing the option value returned by the next collection id query, we directly return the value because we can be certain that it always is a value. A test in testnet and devnet ensures that.


al3mart (commented):
I think sticking with #552 might be a better trade off.

My personal appeal is that on 552 we are not changing the pallet interface, which helps us staying away from pushing breaking changes as long as we can.

Daanvdplas:
Closed in favour of #552

PR #552, opened 2025-05-01
fix(api/nonfungibles): align next_collection_id contract impl with the pallet return type

evilrobot-01:
Resolves the issue with the discrepancy between the pallet return type for a next_collection_id query and the contract library implementation, backed with more robust tests.

### DIFF
--- a/pop-api/integration-tests/src/nonfungibles/mod.rs
+++ b/pop-api/integration-tests/src/nonfungibles/mod.rs
@@ -27,6 +27,7 @@ type Collection = pallet_nfts::Collection<Runtime, NftsInstance>;
 type ItemAttributesApprovals = pallet_nfts::ItemAttributesApprovalsOf<Runtime, NftsInstance>;
 type Metadata = BoundedVec<u8, <Runtime as pallet_nfts::Config<NftsInstance>>::StringLimit>;
 type NextCollectionId = pallet_nfts::NextCollectionId<Runtime, NftsInstance>;
+type NonFungibles = pallet_api::nonfungibles::Pallet<Runtime>;
 
 const COLLECTION: CollectionId = 0;
 const CONTRACT: &str = "contracts/nonfungibles/target/ink/nonfungibles.wasm";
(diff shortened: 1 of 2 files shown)


pop-api/src/v0/nonfungibles/mod.rs
```diff
@@ -149,9 +149,10 @@ pub fn get_attribute(
 #[inline]
 pub fn next_collection_id() -> Result<CollectionId> {
 	build_read_state(NEXT_COLLECTION_ID)
-		.output::<Result<CollectionId>, true>()
+		.output::<Result<Option<CollectionId>>, true>()
 		.handle_error_code::<StatusCode>()
 		.call(&())
+		.map(|r| r.unwrap_or_default())
```
evilrobot-01, line 155: CollectionId is hard coded as a u32 based on the api type definition. Whilst the pallet may return an option, a None value and a Some(0) effectively resolve to the same outcome - the next collection id is zero.

This fix therefore handles this accordingly, ensuring the correct value is returned when None or Some(0): zero.

pop-api/integration-tests/src/nonfungibles/mod.rs
```diff
@@ -318,10 +319,26 @@ fn get_attribute_works() {
...
 	new_test_ext().execute_with(|| {
 		let addr = instantiate(CONTRACT, INIT_VALUE, vec![]);
-		assert_eq!(next_collection_id(&addr), Ok(COLLECTION + 1));
-		assert_eq!(next_collection_id(&addr), Ok(NextCollectionId::get().unwrap_or_default() + 1))
+		assert_eq!(next_collection_id(&addr), Ok(CollectionId::default()));
+		assert_eq!(next_collection_id(&addr), Ok(NextCollectionId::get().unwrap_or_default()));
+		assert_eq!(NextCollectionId::get(), None);
+		assert_eq!(
+			NonFungibles::read(Read::NextCollectionId),
+			ReadResult::NextCollectionId(Some(0))
```
evilrobot-01, line 330: This was the root cause and why the previous tests asserting that the next collection id was one were incorrect. The value returned is Some(0) due to the initial_value() implementation , which was being encoded to one as the return type wasnt handled properly.

To be explicit, a Some(..) value on an option encodes to 1.

pop-api/src/v0/nonfungibles/mod.rs
Daanvdplas: But if this now returns `None`, will it then not fail because this unwrap_or_default() will only happen for the `Some()` case?

evilrobot-01: .call(..) returns Result<Option<CollectionId>>, so the .map(..) closure only runs on success (Result::Ok), meaning r is of type Option<CollectionId>. If r is None, it uses a default value the type, which is CollectionId::default. If its Some(..), it returns the unwrapped value.

image

al3mart (approved):
This makes sense to me.

I think not modifying the pallet's interface as long as we can is a better approach.
Seems that we will always be returned a value from reading next collection id, so defaulting doesn't seem to be a problem and we are handling the error in case we cannot decode the result.

Seems sensible.

Daanvdplas (approved):
Looks good, thank you!

pop-api/src/v0/nonfungibles/mod.rs
Daanvdplas: You are right, thanks

PR #553, opened 2025-05-01
chore: improve network config files

evilrobot-01:
Resolves two issues spotted whilst using the network config files for testing:
- asset-hub was using rococo runtime instead of paseo within devnet.toml
- pop args for mainnet were invalid based on functionality available within mainnet runtime

### DIFF
--- a/networks/devnet.toml
+++ b/networks/devnet.toml
@@ -46,7 +46,7 @@ args = [
 
 [[parachains]]
 id = 1000
-chain = "asset-hub-rococo-local"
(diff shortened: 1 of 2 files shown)


al3mart (approved):
Fine with me. I did push this change in the uplift to 2503 but was asked to be reverted.

While Rococo AH might be always a bit ahead than Paseo AHs as long as it's runtime is maintained I think it it makes sense to do the change given that Rococo is no longer deployed.

Daanvdplas (approved):

PR #555, opened 2025-05-02
ci: enable concurrency controls

evilrobot-01:
Ensures there is only ever one active workflow running per PR. 

Already implemented at https://github.com/r0gue-io/pop-cli/blob/main/.github/workflows/ci.yml#L13.


Daanvdplas (approved):

al3mart (approved):
Nice one!

PR #556, opened 2025-05-03
feat: precompiles

evilrobot-01:
Adds a fork of revive with precompiles as at https://github.com/paritytech/polkadot-sdk/commit/398bbfccf752200d7aaeb75fddb099c5606c0bbd until it is available in a stable release. Manually updated dependencies not currently available from workspace. Also reset the error variant hex values as that feature is not yet available from a published frame-support crate.

Intended to be a base branch for further precompiles work only.


evilrobot-01:
Rebased on main

PR #563, opened 2025-05-08
fix(contract-example): enable devnet feature for fungibles drink tests

al3mart:
Example fungibles contract tests expect the devnet configuration from drink. As per https://github.com/r0gue-io/pop-drink/pull/30 now we enable testnet by default.

This change enables `devnet` such that we keep the same behavior after the linked PR is merged.

NOTE: only merge this after https://github.com/r0gue-io/pop-drink/pull/30 has been merged.


chungquantin:
This duplicates with https://github.com/r0gue-io/pop-node/pull/562

PR #564, opened 2025-05-09
refactor(revive): enable unstable host fn

al3mart:
Enables access to unstable host functions for contracts deployed on revive in devnet and testnet.

As an example, this is required for contracts to use fns like `xcm_send`, [which ink! is a user of](https://github.com/use-ink/ink/blob/d71802f482dc129b199c71715f9910e2d7d4e318/crates/env/src/backend.rs#L471).
A complete list of all host functions, stable and unstable, can be found at: https://github.com/paritytech/polkadot-sdk/blob/f73c228b7a1a4a6b7d80ae7917a61df62f2359b8/substrate/frame/revive/uapi/src/host.rs#L24

I was able to deploy an ink! contract with the unstable hostfn feature enabled without issues.

### Notice to users:

While we are making these functions directly accessible to contracts, the general approach within the ecosystem is to expose them through specific pre-compiles. Therefore, this change may be reverted in the future in favor of using the pop-api and built-in pre-compiles. By enabling access to these functions now, ink! users can fully explore the language's features.

As a result, every ink! contract that intends to use this set of functions must be built with the `unstable-hostfn` feature. Like so:

```toml
ink = { version = "6.0.0-alpha", default-features = false, features = ["unstable-hostfn"] }
```

[Example contract using XCM](https://github.com/use-ink/ink-examples/tree/v6.x/contract-xcm)

### DIFF
--- a/runtime/devnet/src/config/contracts.rs
+++ b/runtime/devnet/src/config/contracts.rs
@@ -111,8 +111,8 @@ impl pallet_revive::Config for Runtime {
 	// 128 MB. Used in an integrity that verifies the runtime has enough memory.
 	type RuntimeMemory = ConstU32<{ 128 * 1024 * 1024 }>;
 	type Time = Timestamp;
-	// Disables access to unsafe host fns such as xcm_send.
-	type UnsafeUnstableInterface = ConstBool<false>;
+	// Enables access to unsafe host fns such as xcm_send.
+	type UnsafeUnstableInterface = ConstBool<true>;
 	type UploadOrigin = EnsureSigned<Self::AccountId>;
 	type WeightInfo = pallet_revive::weights::SubstrateWeight<Self>;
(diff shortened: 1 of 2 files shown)


AlexD10S (approved):
Thanks for the links provided in the description, very helpful. I've tested the deployment, and it looks good to me.

evilrobot-01 (approved):
Whilst the code changes are easy, I'd question why this wasn't picked up by failing tests previously. I assume this is because the integration tests do not yet target revive, and that tests for messaging specifically being limited.

al3mart:
> Whilst the code changes are easy, I'd question why this wasn't picked up by failing tests previously. I assume this is because the integration tests do not yet target revive, and that tests for messaging specifically being limited.

Correct! We are not targeting our integration tests for contracts to revive yet. Now with the alpha release of ink! v6 out it would be a good moment this push these changes.

PR #565, opened 2025-05-12
chore: nonfungible example contract tests with `pop-drink`

chungquantin:
The PR add drink tests to nonfungible example contract.

Requires: https://github.com/r0gue-io/pop-drink/pull/31


evilrobot-01:
Please target this to the fungibles PR branch, so that we dont have to review the same changes which appear in that PR again here.

pop-api/examples/nonfungibles/lib.rs
```diff
@@ -260,7 +265,17 @@ pub mod nonfungibles {
...
 			self.env().emit_event(Destroyed { id: self.id });
+			Ok(())
+		}
 
+		/// Terminate the contract and destroy the collection.
+		///
+		/// This method is similar to `destroy` but terminates the contract after the collection
+		/// destruction.
+		#[ink(message)]
+		pub fn destroy_and_terminate(&mut self, destroy_witness: DestroyWitness) -> Result<()> {
```
chungquantin, line 276: Separate the method for testing purpose. Drink fails with `terminate_contract`.

pop-api/examples/nonfungibles/lib.rs
```diff
@@ -160,20 +165,20 @@ pub mod nonfungibles {
...
-			api::balance_of(self.id, owner).map_err(Psp34Error::from)
+		pub fn balance_of(&self, owner: AccountId) -> u32 {
+			api::balance_of(self.id, owner).unwrap_or_default()
 		}
 
 		/// Returns the owner of an item, if any.
 		#[ink(message)]
-		pub fn owner_of(&self, item: ItemId) -> Result<Option<AccountId>> {
-			api::owner_of(self.id, item).map_err(Psp34Error::from)
+		pub fn owner_of(&self, item: ItemId) -> Option<AccountId> {
```
chungquantin, line 174: To keep it consistent with the fungibles example

evilrobot-01:
@chungquantin please can you kindly resolve the merge conflicts?

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,445 @@
...
+const BOB: AccountId = AccountId::new([2_u8; 32]);
+const COLLECTION: CollectionId = 0;
+const ITEM: ItemId = 0;
+const INIT_AMOUNT: Balance = 100_000_000 * UNIT;
+const INIT_VALUE: Balance = 100 * UNIT;
+
+// The contract bundle provider.
+//
+// See https://github.com/r0gue-io/pop-drink/blob/main/crates/drink/drink/test-macro/src/lib.rs for more information.
+#[drink::contract_bundle_provider]
```
evilrobot-01, line 27: Can this definition be moved within the `deploy` helper function below? It eliminates the complexity/noise of having yet another thing to define, where someone can just copy the deploy function in its entirey and get this definition included.

The separate usage on 77 should then be using the deploy helper.

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,445 @@
...
+enum BundleProvider {}
+
+/// Sandbox environment for Pop Devnet Runtime.
+pub struct Pop {
```
evilrobot-01, line 31:
```suggestion
pub struct Pop(TestExternalities);
```
Would that work?

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,445 @@
...
+		Self { ext }
+	}
+}
+
+// Implement core functionalities for the `Pop` sandbox.
+drink::impl_sandbox!(Pop, Runtime, ALICE);
+
+// Deployment and constructor method tests.
+
+fn deploy_with_default(session: &mut Session<Pop>) -> Result<AccountId> {
```
evilrobot-01, line 49: Would move all the helpers to the bottom after tests. I see value in firstly showing how the sandbox is defined/initialised, but then it should be showing the tests and not more boiler plate.

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,445 @@
...
+		method,
+		input,
+		NO_SALT,
+		Some(INIT_VALUE),
+	)
+}
+
+// A set of helper methods to test the contract calls.
+
+fn collection_id(session: &mut Session<Pop>) -> CollectionId {
```
evilrobot-01, line 409: Would suggest a pattern like https://github.com/r0gue-io/pop-node/blob/9cd371fb2db7b38ae94703dcdcbe2626add89f1d/pop-api/integration-tests/src/messaging.rs#L402-L485 instead, so it then becomes obvious in the tests that one is calling a function on a contract via `contract.collection_id`

pop-api/examples/nonfungibles/tests.rs
chungquantin: No Drink requires that pattern

```bash
error[E0609]: no field `ext` on type `&mut Pop`
  --> tests.rs:43:1
   |
43 | drink::impl_sandbox!(Pop, Runtime, ALICE);
```

We can update pop-drink but would be out of scope for this PR

pop-api/examples/nonfungibles/lib.rs
```diff
@@ -160,20 +165,20 @@ pub mod nonfungibles {
...
+			api::owner_of(self.id, item).unwrap_or_default()
 		}
 
 		/// Returns the total supply of the collection.
 		#[ink(message)]
-		pub fn total_supply(&self) -> Result<u128> {
-			api::total_supply(self.id).map_err(Psp34Error::from)
+		pub fn total_supply(&self) -> u128 {
```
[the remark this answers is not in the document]
chungquantin: The contract returns `u128` because we tried to follow the PSP34 standard.

pop-api/examples/nonfungibles/lib.rs
chungquantin: Fixed in https://github.com/r0gue-io/pop-node/pull/565/commits/7607a1a492748ec62791836391921e00bfd48075.

pop-api/examples/nonfungibles/tests.rs
chungquantin: Good idea! I have moved it to the `deploy` method in https://github.com/r0gue-io/pop-node/pull/565/commits/7607a1a492748ec62791836391921e00bfd48075

pop-api/examples/nonfungibles/tests.rs
chungquantin: Thanks for the suggestion, I have refactored to bring all helpers method under the Contract struct in https://github.com/r0gue-io/pop-node/pull/565/commits/3dd2ad43030249e21bd4eaa80e70f8bfd21046ca

pop-api/examples/nonfungibles/tests.rs
chungquantin: Move this method into `Contract` struct and rename it to `default`. Hence, `Contract::default()` means deploying a contract by default to keep it short.

evilrobot-01:
@chungquantin Please could you kindly rebase again to address the CI issues and then once all green I can review again.

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,485 @@
...
+		error::{v0::Error, Nfts, NftsError::*},
+		AccountId, Balance, Runtime,
+	},
+	sandbox_api::nfts_api::NftsAPI,
+	session::Session,
+	TestExternalities, Weight, NO_SALT,
+};
+use pop_api::v0::nonfungibles::{events::Transfer, CollectionId, ItemId};
+
+use super::*;
```
evilrobot-01, line 14: As these tests just dont run via `cargo test`, would suggest adding something like the following as a hint to anyone:

```
#[cfg(debug_assertions)]
compile_error!("tests must be run using the release profile (--release)");
```

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,485 @@
...
+		}
+	);
+}
+
+#[drink::test(sandbox = Pop)]
+fn collection_id_works(mut session: Session) {
+	let _ = env_logger::try_init();
+	// Deploys a first contract.
+	let contract = Contract::default(&mut session).unwrap();
+	assert_eq!(contract.collection_id(&mut session), 0);
```
evilrobot-01, line 64: Passing session to each call is kinda awkward and makes me question whether the Contract abstraction provides much value.

I have spiked an improvement separately which eliminates passing the session to each call, but when looking at the drink examples I question whether the approach is valid based on the fluent api that it provides. My view is that calling methods on a contract struct more closely aligns with what is happening in the test, making it clear.

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,485 @@
...
+		Self { ext }
+	}
+}
+
+// Implement core functionalities for the `Pop` sandbox.
+drink::impl_sandbox!(Pop, Runtime, ALICE);
+
+#[drink::test(sandbox = Pop)]
+fn new_constructor_works(mut session: Session) {
+	let _ = env_logger::try_init();
```
evilrobot-01, line 43: Why not move this statement in each test into `Pop::default()` above?

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,485 @@
...
+	// Deploys a new contract.
+	let contract = Contract::default(&mut session).unwrap();
+	// Successfully destroys a collection.
+	session.set_gas_limit(Weight::MAX);
+	let witness_string =
+		format!("{:?}", DestroyWitness { item_metadatas: 0, item_configs: 0, attributes: 0 });
+
+	// Error returned "Not enough data to fill buffer" due to contract termination.
+	assert!(session.call::<String, ()>("destroy", &[witness_string], None).is_err());
```
evilrobot-01, line 358: This could be improved imo. it seems to stem from the fact that the call impl in pop-drink panics, eliminating the possibility of handling this cleanly. This means that every contract will have this messy approach to termination.

The "Not enough data to fill buffer" should not be leaked out beyond pop-drink - perhaps it can be converted to a known error type?

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,485 @@
...
+			DestroyWitness { item_metadatas: 1, item_configs: 1, attributes: 1 }
+		),
+		Error::Module(Nfts(BadWitness))
+	);
+}
+
+// A set of helper methods to test the contract deployment and calls.
+
+// Deploys the contract with `NO_SALT and `INIT_VALUE`.
+fn deploy(
```
evilrobot-01, line 400: This seems to only be used within `Contract::new()` now, so why not eliminate this function and just define it there instead?

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,485 @@
...
+}
+
+impl Contract {
+	// Deploy a new contract.
+	fn new(session: &mut Session<Pop>, max_supply: Option<u32>) -> Result<Self> {
+		let contract = deploy(session, "new", vec![format!("{:?}", max_supply)], NO_SALT)?;
+		Ok(Self { address: contract })
+	}
+
+	fn default(session: &mut Session<Pop>) -> Result<Self> {
```
evilrobot-01, line 434: Would eliminate and just use `new(session, None)` personally. Its a few extra characters but conveys intent better.

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,485 @@
...
+
+struct Contract {
+	pub address: AccountId,
+}
+
+impl Contract {
+	// Deploy a new contract.
+	fn new(session: &mut Session<Pop>, max_supply: Option<u32>) -> Result<Self> {
+		let contract = deploy(session, "new", vec![format!("{:?}", max_supply)], NO_SALT)?;
+		Ok(Self { address: contract })
```
evilrobot-01, line 431: Address is never used, which indicates a flaw.

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,485 @@
...
+	}
+
+	fn default(session: &mut Session<Pop>) -> Result<Self> {
+		Contract::new(session, None)
+	}
+
+	fn collection_id(&self, session: &mut Session<Pop>) -> CollectionId {
+		call::<Pop, CollectionId, Psp34Error>(session, "collection_id", vec![], None).unwrap()
```
evilrobot-01, line 439: None of these functions use the contract address, indicating that the pop-drink::call function is flawed in a multi-contract setting. See https://github.com/use-ink/drink/blob/94835ac3124cc02d556be1d9ff9fdf419fc24465/examples/quick-start-with-drink/lib.rs#L169

My assumption is that currently if I deploy two instances via `Contract::new()`, creating two collections, I will only ever be able to operate on the first collection despite calling methods on the second contract instance.

evilrobot-01 (commented):
A few improvement suggestions, along with identification of a few possible flaws.

pop-api/examples/nonfungibles/tests.rs
evilrobot-01: In reference to https://github.com/use-ink/drink/issues/124, should probably be defined in `pop-drink`

pop-api/examples/nonfungibles/tests.rs
chungquantin: Hmm I did try to store the session in the contract struct but seems like it makes things more complicated due to borrowing issues. Let me find any better way to get rid of it.

pop-api/examples/nonfungibles/tests.rs
chungquantin: Resolved in https://github.com/r0gue-io/pop-drink/pull/35 with new `call_with_address` added.

pop-api/examples/nonfungibles/tests.rs
chungquantin: With https://github.com/r0gue-io/pop-node/pull/565/commits/76080bae245d938d8cdfebfa8339e7d0a861a4c5 that refactored to use `call_with_address`, this is now valid.

pop-api/examples/nonfungibles/tests.rs
chungquantin: Agree, I have removed the method.

pop-api/examples/nonfungibles/tests.rs
chungquantin: Agree, I have moved its logic to the `new` method: https://github.com/r0gue-io/pop-node/pull/565/commits/3e1b812b51cb8c7de645d5c907d6d667900c0534

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,485 @@
...
+}
+
+// Implement core functionalities for the `Pop` sandbox.
+drink::impl_sandbox!(Pop, Runtime, ALICE);
+
+#[drink::test(sandbox = Pop)]
+fn new_constructor_works(mut session: Session) {
+	let _ = env_logger::try_init();
+	// Deploys a new contract.
+	let contract = Contract::default(&mut session).unwrap();
```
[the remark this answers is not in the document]
chungquantin: Thanks. I have removed the method `default`, agree that it makes the code readers confusing if we deploy inside the method. I changed it to `new` as you suggested.

pop-api/examples/nonfungibles/tests.rs
```diff
@@ -0,0 +1,485 @@
...
+
+#[drink::test(sandbox = Pop)]
+fn collection_id_works(mut session: Session) {
+	let _ = env_logger::try_init();
+	// Deploys a first contract.
+	let contract = Contract::default(&mut session).unwrap();
+	assert_eq!(contract.collection_id(&mut session), 0);
+
+	// Deploys a second contract increments the collection ID.
+	deploy(&mut session, "new", vec!["None".to_string()], vec![1, 2, 3, 4]).unwrap();
```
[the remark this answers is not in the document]
chungquantin: Resolved in https://github.com/r0gue-io/pop-node/pull/565/commits/3e1b812b51cb8c7de645d5c907d6d667900c0534

pop-api/examples/nonfungibles/tests.rs
chungquantin: Adding this in `pop-drink` would not make sense to me. Because it prevents us to run tests in the `pop-drink` crate. I have added your suggestion on the top of the test file here. Let me know if you disagree with it.

PR #570, opened 2025-05-16
fix(ci): pin stable-1.86 version

al3mart:
The release of rust 1.87.0 introduced issues on CI making tests involving instantiation of wasm smart contracts fail.
Pinning version to stable 1.86.0 instead of latest stable until resolved.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,4 +1,6 @@
 [toolchain]
-channel = "stable"
+# Release of Rust 1.87.0 introduced issues on CI making tests involving instantiation of wasm smart contracts fail.
+# Pinning version to stable 1.86 until resolved.
+channel = "1.86.0"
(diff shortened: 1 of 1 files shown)


evilrobot-01 (approved):

rust-toolchain.toml
```diff
@@ -1,4 +1,6 @@
 [toolchain]
-channel = "stable"
+# Release of Rust 1.87.0 introduced issues on CI making tests involving instantiation of wasm smart contracts fail.
+# Pinning version to stable 1.86 until resolved.
+channel = "1.86.0"
```
evilrobot-01, line 4: Might be worth using "1.86" so any future patch releases are automatically used.

al3mart: Ah, great call!

AlexD10S (approved):

PR #571, opened 2025-05-19
chore: sync with `polkadot-stable2503-4`

al3mart:
This PR not only syncs pop-node with stable 2503-4 version of the sdk.

It also changes the dependencies to use the github tag `polkadot-stable2503-4` instead of their respective crate release versions. The reason for this is that we needed a patch for `pallet-session` that hasn't been yet published on crates.io but is present on the tag we are now using.


Cargo.toml
```diff
@@ -78,131 +78,139 @@ pop-runtime-mainnet = { path = "runtime/mainnet", default-features = true } # de
...
 
 # Substrate
-frame-benchmarking = { version = "40.0.0", default-features = false }
-frame-benchmarking-cli = { version = "47.0.0" }
-frame-executive = { version = "40.0.0", default-features = false }
+# Dependencies changed to point to the tag for stable 2503-4 release as we needed a patch in pallet-session
+# that wasn't released in crates.io for testnet runtime v0.5.3
+frame-benchmarking = { git = "https://github.com/paritytech/polkadot-sdk", tag = "polkadot-stable2503-4", default-features = false }
+frame-benchmarking-cli = { git = "https://github.com/paritytech/polkadot-sdk", tag = "polkadot-stable2503-4" }
+frame-executive = { git = "https://github.com/paritytech/polkadot-sdk", tag = "polkadot-stable2503-4", default-features = false }
```
evilrobot-01, line 85: Why not `frame-metadata`?

al3mart: It's not part of polkadot-sdk repo. So this just follows the version being used by polkadot-sdk at that tag:
https://github.com/paritytech/polkadot-sdk/blob/0548e837980715584bfdea1fb907c9a6b3186bc2/Cargo.toml#L793

There are however more recent versions than the one being used https://crates.io/crates/frame-metadata/versions. But I'm just following upstream here.

evilrobot-01 (approved):
Looks good, thanks.

Daanvdplas (approved):

PR #581, opened 2025-06-13
chore(testnet): v0.5.4

al3mart:
These changes prepare a new release for pop's testnet runtime.

Mainly lowering `MinEligibleCollators` to `1` and removing the already released migrations.

The reason behind the value of `1` as the minimum amount of eligible collators is that to reduce the number of collators in the set we need the following condition to compute true:

```
ensure!(
	Self::eligible_collators() > T::MinEligibleCollators::get(),
	Error::<T>::TooFewEligibleCollators
);
```

Given no candidates at the moment and only `3` invulnerables. `elegible_collators` (candidates + invulnerables) won't be greater than the current value set for `MinEligibleCollators`.

By setting this value to `1` we will be able to modify the number of collators in the future without blockers if needed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14157,7 +14157,7 @@ dependencies = [
 
 [[package]]
 name = "pop-runtime-testnet"
-version = "0.5.3"
+version = "0.5.4"
 dependencies = [
  "anyhow",
  "cumulus-pallet-aura-ext 0.20.0",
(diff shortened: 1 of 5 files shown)


peterwht (approved):

evilrobot-01 (approved):
LGTM

PR #586, opened 2025-08-14
refactor: readme

Daanvdplas:


al3mart (approved):
Nice!

I think it would be awesome adding links to the XCM and ISMP docs!
- https://docs.hyperbridge.network/protocol/ismp
- https://github.com/polkadot-fellows/xcm-format

